PR #331, opened 2021-11-19
Fix KanbanPanel

SasLord:


plugins/view-resources/src/components/KanbanPanel.svelte
```diff
@@ -29,7 +29,7 @@
     <div class="flex-between label">
       <div>
         <Label {label} />
-        <span>({counter})</span>
+        <!-- <span>({counter})</span> -->
```
haiodo, line 32: Why we need this commented but not removed?

aplatoff (approved):

PR #364, opened 2021-11-25
Comments presenter

### LINKED ISSUE #359: Comments Presenter
We need `CommentsPresenter`, similar to `AttachmentsPresenter` and `ApplicationsPresenter`, which will show number of comments in the table (and other views) and show last 3 comments in popup on hover

BykhovDenis:
close #359 Signed-off-by: Denis Bykhov [bykhov.denis@gmail.com](mailto:bykhov.denis@gmail.com)


plugins/chunter-resources/src/components/CommentPresenter.svelte
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  function isToday (time: number): boolean {
+    const current = new Date()
+    const target = new Date(time)
+    return current.getDate() === target.getDate() && current.getMonth() === target.getMonth() && current.getFullYear() === target.getFullYear()
+  }
+</script>
+
+<div class="container">
```
haiodo, line 53: Why not reuse existing Comment component?

haiodo (commented):
I've don't see an server code to update comments counter, or it was done before?

models/recruit/src/index.ts
```diff
@@ -46,6 +46,9 @@ export class TCandidate extends TPerson implements Candidate {
   @Prop(TypeString(), 'Attachments' as IntlString)
   attachments?: number
 
+  @Prop(TypeString(), 'Comments' as IntlString)
+  comments?: number
```
haiodo, line 50: I've could not find there is it updated?

plugins/chunter-resources/src/components/CommentPresenter.svelte
BykhovDenis: A strange component, I did not find it to be used, I think it needs to be removed

models/recruit/src/index.ts
BykhovDenis: server/core/storage:109 here

aplatoff (approved):

PR #372, opened 2021-11-25
Fix query transactions handling

haiodo:
1. Allow to query for transactions properly, and fix some limit issues.
2. Enable tests


packages/core/src/index.ts
```diff
@@ -22,5 +22,6 @@ export * from './client'
 export * from './operator'
 export * from './query'
 export * from './server'
+export * from './__tests__/minmodel'
```
aplatoff: Why? :)

haiodo: Alternative is to have a copy of minmodel inside query package.

BykhovDenis (approved):

PR #378, opened 2021-11-26
Telegram adjustments

sumboid:


plugins/setting-resources/src/components/PluginCard.svelte
```diff
@@ -38,7 +38,9 @@
 
   async function disconnect(): Promise<void> {
```
BykhovDenis, line 39: May be we should remove `close` and `disconnect` functions here, if we create and remove setting using external service

plugins/telegram-resources/src/components/helper.ts
```diff
@@ -0,0 +1,19 @@
...
+  const response = await fetch(url + path, {
+    method: 'POST',
+    headers: {
+      Authorization: 'Bearer ' + (getMetadata(login.metadata.LoginToken) ?? ''),
+      'Content-Type': 'application/json'
+    },
+    body: serialize(data)
+  })
+
+  const res = await response.json()
```
BykhovDenis: What about tg_code error?

aplatoff (approved):

PR #383, opened 2021-11-26
Send message on button click

SasLord:

### DIFF
--- a/packages/text-editor/src/components/ReferenceInput.svelte
+++ b/packages/text-editor/src/components/ReferenceInput.svelte
@@ -23,15 +23,18 @@
   import TextEditor from './TextEditor.svelte'
 
   const dispatch = createEventDispatcher()
+  let content: string
 
 </script>
 
 <div class="ref-container">
   <div class="textInput">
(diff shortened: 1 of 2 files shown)


packages/text-editor/src/components/ReferenceInput.svelte
```diff
@@ -23,15 +23,18 @@
   import TextEditor from './TextEditor.svelte'
 
   const dispatch = createEventDispatcher()
+  let content: string
```
haiodo, line 26: I need this one to be `export let content: string` to be able to use ReferenceInput to edit comments.

packages/text-editor/src/components/TextEditor.svelte
```diff
@@ -31,6 +31,8 @@ import { getClient } from '@anticrm/presentation'
 
 import contact from '@anticrm/contact'
 
+export let content: string
```
haiodo, line 34: Text editor initial value should be set to content as well.

haiodo:
@SasLord I've did similar inside #386, so this one could be closed.

PR #436, opened 2021-11-30
Add tg contact if it is required

sumboid:
This PR is related to https://github.com/hcengineering/telegram/pull/9 so ideally it should be applied after

### DIFF
--- a/plugins/telegram-resources/src/components/Chat.svelte
+++ b/plugins/telegram-resources/src/components/Chat.svelte
@@ -57,9 +57,8 @@
     enabled = res.length > 0
   })
 
-  async function onMessage(event: CustomEvent, isRetry = false) {
-    const to = contactString?.value ?? ''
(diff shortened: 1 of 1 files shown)


plugins/telegram-resources/src/components/Chat.svelte
```diff
@@ -57,18 +57,48 @@
...
         'Content-Type': 'application/json'
       },
       body: JSON.stringify({
-        to: contactString?.value ?? '',
+        to,
         msg: event.detail
       })
     })
+
+    if (res.status === 400 && tryCount === 0) {
```
BykhovDenis: Why we need try count? May be we can just check err.code?

sumboid: Not sure how code may help, it is just a way to exit recursion. Let's call it boolean `isRetry` until we add more complex retrying logic.

BykhovDenis (approved):

PR #473, opened 2021-12-01
Interfaces Support

haiodo:
Introduce Interfaces into model.

No search for interfaces, but it is possible to use Hierarchy descendants to find all classes.

### DIFF
--- a/models/recruit/src/index.ts
+++ b/models/recruit/src/index.ts
@@ -13,20 +13,18 @@
 // limitations under the License.
 //
 
-import type { IntlString } from '@anticrm/platform'
-import { Builder, Model, UX, Prop, TypeString, TypeBoolean } from '@anticrm/model'
-import type { Ref, FindOptions, Doc, Domain, Class } from '@anticrm/core'
-import core, { TSpace, TSpaceWithStates, TDocWithState } from '@anticrm/model-core'
-import type { Vacancy, Candidates, Candidate, Applicant } from '@anticrm/recruit'
 import activity from '@anticrm/activity'
 import type { Employee } from '@anticrm/contact'
-
-import workbench from '@anticrm/model-workbench'
-
-import view from '@anticrm/model-view'
-import contact, { TPerson } from '@anticrm/model-contact'
-import recruit from './plugin'
+import type { Doc, Domain, FindOptions, Ref } from '@anticrm/core'
+import { Builder, Model, Prop, TypeBoolean, TypeString, UX } from '@anticrm/model'
 import chunter from '@anticrm/model-chunter'
+import contact, { TPerson } from '@anticrm/model-contact'
+import core, { TAttachedDoc, TDocWithState, TSpace, TSpaceWithStates } from '@anticrm/model-core'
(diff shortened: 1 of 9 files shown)


models/recruit/src/index.ts
```diff
@@ -80,6 +76,10 @@ export class TApplicant extends TDocWithState implements Applicant {
 
   @Prop(TypeString(), 'Assigned recruiter' as IntlString)
   employee!: Ref<Employee>
+
+  // We need this two to make typescript happy.
+  declare state: TDocWithState['state']
```
haiodo, line 81: Probable we could somehow exclude them for typescript with types, but I've not found an easy way.

packages/core/src/classes.ts
```diff
@@ -118,12 +119,22 @@ export interface Classifier extends Doc, UXObject {
...
 
+/**
+ * @public
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export interface Interface<T extends Obj> extends Classifier {
+  extends?: Ref<Interface<Obj>>[]
+  shortLabel?: IntlString
+}
+
```
aplatoff: `Interface<T extends Doc>` is more correct. What is `shortLabel`? :) If we really need it in `Class` also why not move it to `Classifier`?

packages/core/src/hierarchy.ts
```diff
@@ -96,10 +113,9 @@ export class Hierarchy {
   }
 
   private txCreateDoc (tx: TxCreateDoc<Doc>): void {
-    if (tx.objectClass === core.class.Class) {
-      const createTx = tx as TxCreateDoc<Class<Obj>>
-      const _id = createTx.objectId
-      this.classes.set(_id, TxProcessor.createDoc2Doc(createTx))
+    if ([core.class.Class, core.class.Interface].includes(tx.objectClass)) {
```
aplatoff: I believe this is much slower than ```tx.objectClass === core.class.Class || tx.objectClass ===core.class.Interface```

packages/model/src/dsl.ts
```diff
@@ -105,17 +106,34 @@ export function Index (kind: IndexKind) {
...
+    txes.implements = _implements
     txes.domain = domain
     txes.kind = ClassifierKind.CLASS
   }
 }
 
+/**
+ * @public
+ */
+export function ModelInterface<T extends Obj> (
```
aplatoff: Let's change `ModelInterface` to `Implements`? for better model look

aplatoff (changes requested):
Look nice!

packages/core/src/classes.ts
haiodo: Not really needed for now, removed.

packages/model/src/dsl.ts
haiodo: Done.

packages/core/src/hierarchy.ts
haiodo: Fixed.

aplatoff (approved):

PR #477, opened 2021-12-02
Show/hide navigator

SasLord:


plugins/workbench-resources/src/components/AppItem.svelte
```diff
@@ -27,7 +28,11 @@
 <Tooltip {label}>
   <button class="app" class:selected={selected} on:click={action}>
     <div class="icon-container" class:noty={notify}>
-      <Icon icon={icon} size={'large'}/>
+      {#if typeof (icon) === 'string'}
```
haiodo: @SasLord This is redundant, Icon component already support both asset and svelte component. So Please use just <Icon .../>

aplatoff (approved):

PR #484, opened 2021-12-02
Adjust telegram service expected error shape

sumboid:
Also covers multiple issues


sumboid:
Related to https://github.com/hcengineering/telegram/pull/15

aplatoff (approved):

PR #514, opened 2021-12-03
Fix workspace create

haiodo:
Fix workspace create


plugins/task/src/index.ts
```diff
@@ -52,3 +54,44 @@ export default plugin(taskId, {
     Task: '' as Asset
   }
 })
+
+/**
+ * @public
+ */
+export async function createProjectKanban (
```
haiodo, line 61: Ideally it should not be here, but no better place for now.

BykhovDenis (approved):

PR #567, opened 2021-12-08
Sorting

### LINKED ISSUE #509: `Modified On` sorting does not work for Applications

### LINKED ISSUE #538: state sorting do not work in tasks and applications
Screenshot 2021-12-06 at 14 24 09

BykhovDenis:
close #509 close #538 Signed-off-by: Denis Bykhov [bykhov.denis@gmail.com](mailto:bykhov.denis@gmail.com)

### DIFF
--- a/packages/core/src/query.ts
+++ b/packages/core/src/query.ts
@@ -75,7 +75,15 @@ function arrayOrValue (vv: any): any[] {
 export function resultSort<T extends Doc> (result: T[], sortOptions: SortingQuery<T>): void {
   const sortFunc = (a: any, b: any): number => {
     for (const key in sortOptions) {
-      const result = typeof a[key] === 'string' ? a[key].localeCompare(b[key]) : a[key] - b[key]
+      let aValue = getNestedValue(key, a)
+      if (typeof aValue === 'object') {
+        aValue = JSON.stringify(aValue)
+      }
+      let bValue = getNestedValue(key, b)
+      if (typeof bValue === 'object') {
+        bValue = JSON.stringify(bValue)
+      }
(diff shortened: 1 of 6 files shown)


packages/core/src/query.ts
```diff
@@ -75,7 +75,15 @@ function arrayOrValue (vv: any): any[] {
 export function resultSort<T extends Doc> (result: T[], sortOptions: SortingQuery<T>): void {
   const sortFunc = (a: any, b: any): number => {
     for (const key in sortOptions) {
-      const result = typeof a[key] === 'string' ? a[key].localeCompare(b[key]) : a[key] - b[key]
+      let aValue = getNestedValue(key, a)
+      if (typeof aValue === 'object') {
+        aValue = JSON.stringify(aValue)
```
haiodo, line 80: Why we need to convert to JSON?

BykhovDenis: Simple way for compare objects. Let me know if you have any ideas

haiodo: Lets's define sorters based on type, or use some contribution logic. It is not good if object for example has number, it not work well.

BykhovDenis: How can I use type-based soreters with MongoDB...?

BykhovDenis: I think we should not allow sorting by objects.

haiodo: Why we need this part of code in this case?

BykhovDenis: We can extend the `BuildModeKey` from #576 with the `sortKey` property, for example ...

aplatoff (approved):

PR #680, opened 2021-12-20
Move with state and validate

### LINKED ISSUE #616: Support Move for `Tasks`
Now moving Applicaton to another vacancy do not work -- not showing in Kanban. Also move semantic may be different for some classes: e.g. for Applications there should be no other application for same candidate... We need to address this issue somehow

BykhovDenis:
close #616 Signed-off-by: Denis Bykhov [80476319+BykhovDenis@users.noreply.github.com](mailto:80476319+BykhovDenis@users.noreply.github.com)

### DIFF
--- a/plugins/recruit-resources/src/components/CreateApplication.svelte
+++ b/plugins/recruit-resources/src/components/CreateApplication.svelte
@@ -15,16 +15,17 @@
 <script lang="ts">
   import type { Employee } from '@anticrm/contact'
   import contact from '@anticrm/contact'
-  import { Ref, SortingOrder, Space } from '@anticrm/core'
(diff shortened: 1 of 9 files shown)


plugins/recruit-resources/src/components/CreateApplication.svelte
```diff
@@ -33,17 +34,33 @@
   export let preserveCandidate = false
 
   let status: Status = OK
-  let _space = space
+
+  const doc: Applicant = {
```
haiodo, line 38: Why not use AttachedData\<Applicant>?

aplatoff (approved):

PR #702, opened 2021-12-22
Fix workspace creation

### LINKED ISSUE #696: Fix workspace creation

sumboid:
Resolves #696
Also task, lead and recruit migrations have been removed

### DIFF
--- a/dev/tool/src/workspace.ts
+++ b/dev/tool/src/workspace.ts
@@ -16,7 +16,7 @@
 
 import contact from '@anticrm/contact'
 import core, { DOMAIN_TX, Tx } from '@anticrm/core'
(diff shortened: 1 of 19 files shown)


models/all/src/migration.ts
```diff
@@ -17,17 +17,11 @@ import { MigrateOperation } from '@anticrm/model'
 
 // Import migrate operations.
 import { coreOperation } from '@anticrm/model-core'
-import { taskOperation } from '@anticrm/model-task'
```
haiodo: I recommend to not remove them, since I assume a lot of times we will need them back.

sumboid: Done

aplatoff (approved):

PR #752, opened 2021-12-30
Gmail plugin

### LINKED ISSUE #717: Gmail integration

BykhovDenis:
close #717


models/contact/src/index.ts
```diff
@@ -189,17 +189,6 @@ export function createModel (builder: Builder): void {
     presenter: contact.component.ChannelsPresenter
   })
 
-  builder.createDoc(
```
haiodo, line 192: Not sure it is a good idea, since it could be many email providers.

plugins/gmail-resources/src/components/FullMessageContent.svelte
```diff
@@ -0,0 +1,53 @@
...
+<style lang="scss">
+  .input {
+    overflow: auto;
+    padding: 1rem;
+    background-color: #fff;
+    color: #1f212b;
+    border-radius: .5rem;
+
+    :global(a) {
+      // PLEASE FIX IT
```
haiodo: Should it be fixed later?

plugins/gmail-resources/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "compilerOptions": {
+    "moduleResolution": "node",
+    "target": "esnext",
```
haiodo, line 4: Ed ext is not working on safari 13

BykhovDenis: изображение
But....

plugins/gmail-resources/src/components/FullMessageContent.svelte
BykhovDenis: Removed)

models/contact/src/index.ts
BykhovDenis: I think when we will have several providers then we will be puzzled by this, because there will be additional questions on the logic of choosing a provider that will have to be resolved

aplatoff (approved):

PR #819, opened 2022-01-13
Use ui colors in kanban

### LINKED ISSUE #806: Replace RGB colors with color numbers in states

sumboid:
Resolves #806


dev/tool/src/importer.ts
```diff
@@ -388,7 +388,7 @@ async function createUpdateSpaceKanban (spaceId: Ref<Vacancy>, client: TxOperati
       sid,
```
haiodo, line 386: You need to take length using `getPlatformColorCount`

haiodo (changes requested):

haiodo (approved):

aplatoff (approved):

PR #821, opened 2022-01-13
Add svelte-check in setting. Fix errors.

SasLord:


models/gmail/src/index.ts
```diff
@@ -86,7 +86,7 @@ export function createModel (builder: Builder): void {
     setting.class.IntegrationType,
     core.space.Model,
     {
-      label: 'Gmail',
+      label: 'Gmail' as IntlString,
```
haiodo, line 89: Not a good idea, better to move 'Gmail' to a label and en.json

models/telegram/src/index.ts
```diff
@@ -71,7 +71,7 @@ export function createModel (builder: Builder): void {
     setting.class.IntegrationType,
     core.space.Model,
     {
-      label: 'Telegram',
+      label: 'Telegram' as IntlString,
```
haiodo, line 74: Please create telegram.string.Telegram and use it instead

haiodo (approved):

PR #834, opened 2022-01-18
EditDoc

### LINKED ISSUE #826: Unify EditContact

BykhovDenis:
close #826 Signed-off-by: Denis Bykhov [80476319+BykhovDenis@users.noreply.github.com](mailto:80476319+BykhovDenis@users.noreply.github.com)


models/contact/src/index.ts
```diff
@@ -124,25 +123,30 @@ export function createModel (builder: Builder): void {
...
+      component: contact.component.Contacts
+    },
+    contact.app.Contacts
+  )
 
   builder.createDoc(view.class.Viewlet, core.space.Model, {
     attachTo: contact.class.Contact,
     descriptor: view.viewlet.Table,
-    open: contact.component.EditContact,
+    open: contact.component.EditPerson,
```
haiodo: why? what about Organization?

BykhovDenis: I think we must remove this unused param....

plugins/view-resources/src/components/EditDoc.svelte
```diff
@@ -0,0 +1,328 @@
...
+    }
+    throw new Error(`Icon not found for ${_class}`)
+  }
+
+  $: icon = object && getIcon(object._class)
+
+  let mainEditor: HTMLElement
+  let prevEditor: HTMLElement
+  let maxHeight = 0
+  const observer = new ResizeObserver(() => {
```
haiodo: We discussed with @aplatoff about same size headers for all mixins per/class. So it is ok to have, but it should not be necessary.

models/contact/src/index.ts
haiodo: If it is unused, let's remove it. Because it is confusing.

aplatoff (approved):

PR #838, opened 2022-01-18
Fix full text search control behaviour

### LINKED ISSUE #833: Click on comment in table reset search
open object from table also resets search

haiodo:
Fix full text search control behaviour

1. Search after Enter only.
2. Update search on empty value.

Closes #833


packages/ui/src/components/SearchEdit.svelte
```diff
@@ -0,0 +1,42 @@
...
+  on:change={() => {
+    if (_search === '') {
+      value = ''
+      dispatch('change', '')
+    }
+  }}
+  on:input={() => {
+    if (_search === '') {
+      value = ''
+      dispatch('change', '')
```
BykhovDenis, line 33: Why we need clear search when input?

haiodo: input is handled for any change, like copy/paste. And for empty case I've specially added to inform about changes with no waiting for pressing Enter

BykhovDenis (approved):

PR #852, opened 2022-01-21
Attached channel

BykhovDenis:


models/lead/src/index.ts
```diff
@@ -115,15 +115,33 @@ export function createModel (builder: Builder): void {
...
+    options: {
+      lookup: { _id: contact.class.Channel, as: 'channels' }
+    } as FindOptions<Doc>, // TODO: fix
+    config: [
+      '',
+      { key: 'leads', presenter: lead.component.LeadsPresenter, label: lead.string.Leads },
+      'modifiedOn',
+      '$lookup.channels'
+    ]
+  })
```
haiodo, line 131: We need stable Id here.

models/lead/src/index.ts
```diff
@@ -115,15 +115,33 @@ export function createModel (builder: Builder): void {
     lead.space.DefaultFunnel
   )
 
+  builder.createDoc(view.class.Viewlet, core.space.Model, {
```
haiodo, line 118: Why we need it?

plugins/recruit-resources/src/components/CreateCandidate.svelte
```diff
@@ -153,15 +154,7 @@
...
-    {#if !object.channels || object.channels.length === 0}
-      <CircleButton icon={IconAdd} size={'small'} transparent on:click={(ev) => showPopup(contact.component.SocialEditor, { values: object.channels ?? [] }, ev.target, (result) => { object.channels = result })} />
-      <span><Label label={presentation.string.AddSocialLinks} /></span>
-    {:else}
-      <Channels value={object.channels} size={'small'} />
-      <div class="ml-1">
-        <CircleButton icon={contact.icon.Edit} size={'small'} transparent on:click={(ev) => showPopup(contact.component.SocialEditor, { values: object.channels ?? [] }, ev.target, (result) => { object.channels = result })} />
-      </div>
-    {/if}
+    <ChannelsEditor attachedTo={candidateId} attachedClass={contact.class.Person} />
```
haiodo, line 157: What it we press cancel, channel will be left behind and not removed.

BykhovDenis: Yes. Like a resume for example ;) ... I think we need call `deleteObject` before close, but I'm not sure about it.

haiodo: I think we should not create anythink before candidate is created. Because transactions will left not removed if you will call deleteObject.

PR #970, opened 2022-02-08
Clear space on app change

haiodo:
Clear space on app change

### DIFF
--- a/plugins/workbench-resources/src/components/Workbench.svelte
+++ b/plugins/workbench-resources/src/components/Workbench.svelte
@@ -170,6 +170,10 @@
   $: isNavigate = !!navigatorModel
 
   function navigateApp (app: Application) {
+    if (currentApp === app._id) {
+      // Nothing to do.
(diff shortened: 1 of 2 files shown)


plugins/workbench-resources/src/components/Workbench.svelte
```diff
@@ -170,6 +168,22 @@
 
   let isNavigate: boolean = false
   $: isNavigate = !!navigatorModel
+
+  function navigateApp (app: Application) {
+    currentApp = app._id
+    currentApplication = app
+    navigatorModel = currentApplication?.navigatorModel
+
+    currentSpace = undefined
```
BykhovDenis, line 181: may be we should compare currentApp with new and clear only if change?

haiodo: Thanks, resolved.

BykhovDenis (approved):

PR #1001, opened 2022-02-11
Update StatesBar (Safari fix). CSS optimize.

SasLord:


plugins/task-resources/src/components/state/StatesBar.svelte
```diff
@@ -33,7 +33,7 @@
   let maskLeft: boolean = false
   let maskRight: boolean = false
   let mask: 'left' | 'right' | 'both' | 'none' = 'none'
-  let stepStyle = (gap === 'small') ? 'step-lr25' : 'step-lr75'
+  let stepStyle = (gap === 'small') ? 'safari-gap-1' : 'safari-gap-2'
```
haiodo: may be we rename it from safari-gap to just gap-1, gap-2?

haiodo (approved):

PR #1126, opened 2022-03-10
Minor fix

BykhovDenis:
Signed-off-by: Denis Bykhov [80476319+BykhovDenis@users.noreply.github.com](mailto:80476319+BykhovDenis@users.noreply.github.com)

### DIFF
--- a/server/core/src/storage.ts
+++ b/server/core/src/storage.ts
@@ -160,15 +160,19 @@ class TServerStorage implements ServerStorage {
           const baseClass = this.hierarchy.getBaseClass(_class)
           if (baseClass !== _class) {
             // Mixin opeeration is required.
-            return [txFactory.createTxMixin(_id, attachedTo._class, attachedTo.space, _class, {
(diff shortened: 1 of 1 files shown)


server/core/src/storage.ts
```diff
@@ -198,7 +202,7 @@ class TServerStorage implements ServerStorage {
     const _class = txClass(tx)
     const objClass = txObjectClass(tx)
     return await ctx.with('tx', { _class, objClass }, async (ctx) => {
-      if (tx.objectSpace !== core.space.DerivedTx) {
```
haiodo, line 201: It should be tx.space, please check why it not.

haiodo (changes requested):

server/core/src/storage.ts
BykhovDenis: It was my mistake, only here, everywhere DerivedTx is tx space

haiodo (approved):

PR #1218, opened 2022-03-28
Edit and remove avatar

### LINKED ISSUE #1066: Click On Avatar should show Avatar edit dialog
Right now click on avatar open file attach dialog, so there is no way to remove avatar.

BykhovDenis:
close #1066 Signed-off-by: Denis Bykhov [80476319+BykhovDenis@users.noreply.github.com](mailto:80476319+BykhovDenis@users.noreply.github.com)


packages/presentation/src/components/Avatar.svelte
```diff
@@ -27,7 +27,7 @@
     getBlobURL(direct).then((blobURL) => {
       url = blobURL
     })
-  } else if (avatar !== undefined) {
+  } else if (avatar !== undefined && avatar !== null) {
```
haiodo, line 30: It is same as 
```suggestion
  } else if (avatar != null) {
```
?

haiodo (approved):

PR #1273, opened 2022-04-04
Chunter last views

BykhovDenis:
Signed-off-by: Denis Bykhov [80476319+BykhovDenis@users.noreply.github.com](mailto:80476319+BykhovDenis@users.noreply.github.com)


ksilobait:
Hovering over the first item leads to partial overlapping of reply count
image

ksilobait:
Switching between sections of the app (e.g. Chat -> Contacts -> Chat) closes the channel but doesn't close the thread, is this expected?

plugins/chunter-resources/src/components/Message.svelte
```diff
@@ -14,10 +14,13 @@
 -->
 
 <script lang="ts">
+import { Attachment } from '@anticrm/attachment';
```
haiodo:
```suggestion
  import { Attachment } from '@anticrm/attachment'
```

plugins/chunter-resources/src/components/Message.svelte
```diff
@@ -14,10 +14,13 @@
...
+
   import { AttachmentDocList } from '@anticrm/attachment-resources'
+import AttachmentList from '@anticrm/attachment-resources/src/components/AttachmentList.svelte';
```
haiodo:
```suggestion
  import AttachmentList from '@anticrm/attachment-resources/src/components/AttachmentList.svelte'
```

haiodo (approved):

BykhovDenis:
> Switching between sections of the app (e.g. Chat -> Contacts -> Chat) closes the channel but doesn't close the thread, is this expected?

Currently is expected, it can be changed in future

haiodo (approved):

ksilobait:
I see these `x replies Last reply` only in dev build but not in docker build, is this expected?
image

BykhovDenis:
> I see these `x replies Last reply` only in dev build but not in docker build, is this expected? image

You should rebuild docker and update workspaces

ksilobait:
> > I see these `x replies Last reply` only in dev build but not in docker build, is this expected? image
> 
> You should rebuild docker and update workspaces

Ok, now it works for all new messages

PR #1291, opened 2022-04-06
Support new Svelte & Allow to configure space component

### LINKED ISSUE #1287: Upgrade Svelte to use
Upgrade Svelte to use
{@const a=1+4} and <div style:width={20}/>

### LINKED ISSUE #1288: Add component to workbench.mixin.SpaceView
Allow space to configure component to display content.
workbench.mixin.SpaceView

haiodo:
Support new Svelte & Allow to configure space component

Closes #1288
Closes #1287

### DIFF
--- a/plugins/view-resources/src/components/Move.svelte
+++ b/plugins/view-resources/src/components/Move.svelte
@@ -105,10 +105,8 @@
   </div>
   <div class="spaceSelect">
     {#await getSpace() then}
-      {#if currentSpace}
-        {#if _class}
-          <SpaceSelect _class={currentSpace._class} label={_class} placeholder={_class} bind:value={object.space} />
-        {/if}
+      {#if currentSpace && _class}
+        <SpaceSelect _class={currentSpace._class} label={_class} placeholder={_class} bind:value={object.space} />
       {/if}
     {/await}
   </div>
--- a/plugins/view-resources/src/components/Move.svelte
+++ b/plugins/view-resources/src/components/Move.svelte
@@ -106,7 +106,9 @@
   <div class="spaceSelect">
     {#await getSpace() then}
       {#if currentSpace}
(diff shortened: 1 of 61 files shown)


plugins/view-resources/src/components/StringPresenter.svelte
```diff
@@ -15,12 +15,10 @@
 -->
 
 <script lang="ts">
-  import type { Type } from '@anticrm/core'
-
   export let value: string
   
   // eslint-disable-next-line
-  export let attributeType: Type<any>
+  // export let attributeType: Type<any>
```
BykhovDenis, line 21: Why we need it?

BykhovDenis (approved):

plugins/view-resources/src/components/StringPresenter.svelte
haiodo: Svelte show warning if presenter is called without property specified, but if it is not used it show error during compile time. So I've not removed since plan to find a way how to solve it for both compile and dev plugin :-)

plugins/view-resources/src/components/Move.svelte
```diff
@@ -106,7 +106,9 @@
   <div class="spaceSelect">
     {#await getSpace() then}
       {#if currentSpace}
-        <SpaceSelect _class={currentSpace._class} label={_class ?? ''} bind:value={object.space} />
+        {#if _class}
```
lvfxx: Can we write `{#if currentSpace && _class}` to get rid of nested `if`s?

lvfxx (approved):

plugins/view-resources/src/components/Move.svelte
haiodo: Thanks, fixed.

PR #1299, opened 2022-04-06
Board: Add "Create new Kanban list" panel

### LINKED ISSUE #1263: Board: Kanban show allow to create new list
1. We should be able to create new list/state in Kanban board itself.
2. Option should be optional

Dvinyanin:


plugins/task-assets/lang/en.json
```diff
@@ -28,6 +28,9 @@
     "MarkAsDone": "Mark as done",
     "MarkAsUndone": "Mark as undone",
     "Kanban": "Kanban",
+    "NewList": "Add another list",
```
haiodo, line 31: Please add ru.json changes.

plugins/task-resources/src/components/kanban/KanbanPanelEmpty.svelte
```diff
@@ -0,0 +1,71 @@
...
+    .panel-container {
+        display: flex;
+        flex-direction: column;
+        align-items: stretch;
+        width: 20rem;
+        height: fit-content;
+        background-color: transparent;
+        border: 1px solid transparent;
+        border-radius: .25rem;
+        padding: 4px;
```
haiodo: Please do not mix px and rem sizes. Also please use flex if possible without explicit with. global.scss could be source for some existing styles.

plugins/task-resources/src/components/kanban/KanbanView.svelte
```diff
@@ -207,7 +221,7 @@
             {/each}
           </KanbanPanel>
         {/each}
-        <!-- <KanbanPanelEmpty label={'Add new column'} /> -->
+        <KanbanPanelEmpty on:add={(e) => addItem(e.detail)}/>
```
haiodo: Better to use slot and allow to pass KanbanPanelEmpty if required for Kanban users, since this functionality will be optional.

haiodo (changes requested):

plugins/task-resources/src/components/kanban/KanbanPanelEmpty.svelte
```diff
@@ -0,0 +1,71 @@
...
+    const dispatch = createEventDispatcher()
+    const newList = {
+        title: '',
+        color: ''
+    }
+
+    async function onAdd () {
+        if (!newList.title) return;
+        dispatch('add', newList)
+        isAdding = false
```
haiodo, line 15: Should we set title to '' after value is added?

plugins/task-assets/lang/en.json
Dvinyanin: Added

plugins/task-resources/src/components/kanban/KanbanPanelEmpty.svelte
Dvinyanin: Fixed

plugins/task-resources/src/components/kanban/KanbanView.svelte
Dvinyanin: Done

plugins/task-resources/src/components/kanban/KanbanPanelEmpty.svelte
Dvinyanin: Set

haiodo (approved):

PR #1319, opened 2022-04-07
Board: 1265: Make Card Actions extensible

annano:


plugins/board-resources/src/utils/CardActionUtils.ts
```diff
@@ -1,86 +1,9 @@
...
-      label: board.string.Automation,
-      actions: [AddButtonAction]
-    },
-    {
-      label: board.string.Actions,
-      actions: [MoveAction, CopyAction, MakeTemplateAction, WatchAction, ArchiveAction]
-    }
-  ]
+export const getCardActions = async (client: Client) => {
+  return await client.findAll(board.class.CardAction, {});
```
BykhovDenis: Why you need this? You can use it directy

plugins/board-resources/src/components/editor/CardActions.svelte
```diff
@@ -14,30 +14,81 @@
...
+  const client = getClient()
+
+  let addToCardActions: CardAction[] = []
+  let automationActions: CardAction[] = []
+  let actions: CardAction[] = []
+
+  let actionGroups: { label: IntlString; actions: CardAction[] }[] = []
+
+  getCardActions(client).then((result) => {
+    result.forEach(async (action) => {
```
BykhovDenis: It will not work, because .forEach doesn't await inner func. You should use `for of` or any another loop and await all promises. Currently sort may work incorrect

BykhovDenis (approved):

BykhovDenis (changes requested):
Please fix this ```getCardActions(client).then((result) => {
    result.forEach(async (action) => {```

plugins/board-resources/src/utils/CardActionUtils.ts
annano: Fixed

plugins/board-resources/src/components/editor/CardActions.svelte
annano: Fixed

BykhovDenis (approved):

haiodo (approved):

PR #1329, opened 2022-04-08
Board: Update board card model

annano:
Signed-off-by: Anna No [anna.no@xored.com](mailto:anna.no@xored.com)

### DIFF
--- a/plugins/board/src/index.ts
+++ b/plugins/board/src/index.ts
@@ -15,10 +15,11 @@
 //
 
 import { Employee } from '@anticrm/contact'
-import type { AttachedDoc, Class, TxOperations as Client, Doc, Markup, Ref } from '@anticrm/core'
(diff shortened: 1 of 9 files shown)


plugins/board/src/index.ts
```diff
@@ -53,9 +53,18 @@ export interface CardLabel extends AttachedDoc {
 export interface Card extends Task {
   title: string
 
+  date?: {
```
BykhovDenis: You should extract it in another interface. And please do same changes in model

BykhovDenis (changes requested):

plugins/board/src/index.ts
annano: Extracted `CardDate` into separate interface, but this part `And please do same changes in model` I do not get.

annano: Updated

BykhovDenis (approved):

haiodo (approved):

PR #1336, opened 2022-04-08
Chunter: Delete message

denis-bunakalya:
https://github.com/hcengineering/anticrm/issues/1300


plugins/chunter-resources/src/components/Message.svelte
```diff
@@ -59,6 +60,23 @@
         action: chunter.actionImpl.SubscribeMessage
       } as Action)
 
+  async function deleteMessage () {
+    if (space) {
+      await client.removeDoc(chunter.class.Message, space, message._id)
+      const loc = getCurrentLocation()
+
+      if (loc.path.length > 3) {
+        loc.path.length = 3
```
BykhovDenis, line 67: Why we need change current url?

plugins/chunter-resources/src/components/Message.svelte
```diff
@@ -59,6 +60,23 @@
         action: chunter.actionImpl.SubscribeMessage
       } as Action)
 
+  async function deleteMessage () {
+    if (space) {
+      await client.removeDoc(chunter.class.Message, space, message._id)
```
BykhovDenis: You have space in message, you don't need it. Also you should use `await client.remove(message)`

plugins/chunter-resources/src/components/Message.svelte
```diff
@@ -37,6 +37,7 @@
   export let message: WithLookup<Message>
   export let employees: Map<Ref<Employee>, Employee>
   export let thread: boolean = false
+  export let space: Ref<Space> | undefined
```
BykhovDenis: You don't need it

plugins/chunter-resources/src/components/Message.svelte
BykhovDenis: You should check it before close thread `if (loc.path[3] === message._id)`

BykhovDenis (changes requested):

plugins/chunter-resources/src/components/Message.svelte
denis-bunakalya: Done

plugins/chunter-resources/src/components/Message.svelte
denis-bunakalya: Done

plugins/chunter-resources/src/components/Message.svelte
denis-bunakalya: Removed

BykhovDenis (approved):

PR #1376, opened 2022-04-12
Board: Add card details (members, labels, date)

annano:


packages/ui/src/components/Button.svelte
```diff
@@ -47,6 +48,8 @@
   bind:this={input}
   class="button {kind} {size} jf-{justify}"
   class:only-icon={iconOnly}
+  class:border-radius-1={!isCircle}
```
haiodo, line 51: Maybe better to add additional kind style?

haiodo (approved):

PR #1386, opened 2022-04-13
Remove paddings from card-container

annano:


plugins/recruit-resources/src/components/KanbanCard.svelte
```diff
@@ -28,51 +28,56 @@
   export let object: WithLookup<Applicant>
   export let dragged: boolean
 
-  function showCandidate () {
+  function showCandidate() {
```
haiodo: Please use eslint

packages/kanban/src/components/Kanban.svelte
```diff
@@ -261,9 +257,7 @@
       cursor: grab;
     }
     &.dragged {
-      padding: 1rem;
```
haiodo, line 264: Why? this was done by purpose, to make all of them look similar during dragging.

annano: I think this padding is to compensate the border, which I have also removed from class, the border was the same color as a card-container or unset, so I think we can just remove it at all, and without the border this padding is also not needed.

haiodo (approved):

plugins/recruit-resources/src/components/KanbanCard.svelte
annano: Done

haiodo (approved):

PR #1395, opened 2022-04-13
Chunter: Edit message

denis-bunakalya:
https://github.com/hcengineering/anticrm/issues/1304
Cancel button and edited label will be added in a separate pr


BykhovDenis (changes requested):
We need show `(edited)` label when message was edited. For example
изображение
изображение

BykhovDenis (approved):

PR #1396, opened 2022-04-13
Chunter: pin messages

budaeva:


plugins/chunter-resources/src/components/PinnedMessages.svelte
```diff
@@ -0,0 +1,31 @@
...
+        UnpinMessage(props.message)
+      }
+    })
+  }
+</script>
+
+{#if pinnedIds.length > 0}
+  <div class="bottom-divider container">
+    <div on:click={(ev) => showMessages(ev)}>
+      {pinnedIds.length}
```
BykhovDenis, line 20: Let's add  `Pinned: ` label before

plugins/chunter-resources/src/components/PinnedMessages.svelte
```diff
@@ -0,0 +1,31 @@
...
+    showPopup(PinnedMessagesPopup, { pinnedIds }, ev.target, (props) => {
+      if (props) {
+        UnpinMessage(props.message)
+      }
+    })
+  }
+</script>
+
+{#if pinnedIds.length > 0}
+  <div class="bottom-divider container">
```
BykhovDenis: Please add `over-underline`, `pt-2`, `pb-2` classes

BykhovDenis: And use `ml-10` instead `container`

plugins/chunter-resources/src/components/PinnedMessages.svelte
```diff
@@ -0,0 +1,31 @@
...
+    <div on:click={(ev) => showMessages(ev)}>
+      {pinnedIds.length}
+    </div>
+  </div>
+{/if}
+
+<style lang="scss">
+  .container {
+    padding-left: 0.5rem;
```
BykhovDenis: I think it should be 2.5rem

BykhovDenis (approved):

PR #1405, opened 2022-04-14
Board: Fix labels model

annano:


plugins/board/src/index.ts
```diff
@@ -42,9 +42,7 @@ export interface BoardView extends SpaceWithStates {
  * @public
  */
 export interface CardLabel extends AttachedDoc {
-  attachedTo: Ref<Board>
-
-  title: string
+  title?: string
```
BykhovDenis: It really can be undefined?

BykhovDenis (approved):

plugins/board/src/index.ts
annano: Yes, system labels have only color by default, user defined labels may also have texts.

haiodo: It will be better to use title: string | null, since we do not support undefined for model fields, because mongo doesn't :-)

haiodo: Let's make it not optional and empty if empty/

annano: Updated

haiodo (approved):

PR #1448, opened 2022-04-19
Board: Add EditMember popup

### LINKED ISSUE #1370: Card: Member presenter Popup (make actions configurable from props)

lvfxx:
Closes #1370

### DIFF
--- a/plugins/board-resources/src/components/popups/EditMember.svelte
+++ b/plugins/board-resources/src/components/popups/EditMember.svelte
@@ -1,0 +1,88 @@
+<script lang="ts">
+  import { Employee, formatName } from '@anticrm/contact'
+  import { getFirstName, getLastName } from '@anticrm/contact'
+  import type { IntlString } from '@anticrm/platform'
+  import { ActionIcon, IconClose, Label } from '@anticrm/ui'
+
+  export let member: Employee
+  export let menuItems: { title: IntlString; handler: () => void }[][]
+  export let onClose: () => void
+
+  const firstName = getFirstName(member.name)
+  const lastName = getLastName(member.name)
+  const nameLabel = `${firstName?.[0] ?? ''}${lastName?.[0] ?? ''}`.toUpperCase()
+  const formattedName = formatName(member.name)
+</script>
+
+<div class="antiPopup container pb-4 w-85">
+  <div class="close-icon">
+    <ActionIcon icon={IconClose} size={'small'} action={onClose} />
+  </div>
+  <div class="flex top p-4">
+    <div class="avatar flex-center">
+      <span class="name-label fs-bold">{nameLabel}</span>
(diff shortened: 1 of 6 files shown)


plugins/board-resources/src/components/popups/EditMember.svelte
```diff
@@ -0,0 +1,88 @@
+<script lang="ts">
+  import { Employee, formatName } from '@anticrm/contact'
+  import { getFirstName, getLastName } from '@anticrm/contact'
```
annano, line 3: Please combine these two imports in one, maybe better to import Employee separately as type `import type {Employee} from...`

plugins/board-resources/src/components/popups/EditMember.svelte
```diff
@@ -0,0 +1,88 @@
...
+  const lastName = getLastName(member.name)
+  const nameLabel = `${firstName?.[0] ?? ''}${lastName?.[0] ?? ''}`.toUpperCase()
+  const formattedName = formatName(member.name)
+</script>
+
+<div class="antiPopup container pb-4 w-85">
+  <div class="close-icon">
+    <ActionIcon icon={IconClose} size={'small'} action={onClose} />
+  </div>
+  <div class="flex top p-4">
```
annano, line 21: I think it would be better to reuse `ContactPresenter`
`import { ContactPresenter } from '@anticrm/contact-resources'`

Also other presenters from that plugin can be explored (maybe `EmployeeAccountPresenter`), but it would be better to reuse code here

plugins/board-resources/src/components/popups/EditMember.svelte
```diff
@@ -0,0 +1,88 @@
...
+      {/each}
+      {#if i + 1 < menuItems.length}
+        <div class="divisor" />
+      {/if}
+    {/each}
+  {/if}
+</div>
+
+<style lang="scss">
+  .avatar {
```
annano, line 50: Can we reuse global styles here as much as possible here: https://github.com/hcengineering/anticrm/blob/main/packages/theme/styles/_layouts.scss

haiodo (approved):

plugins/board-resources/src/components/popups/EditMember.svelte
lvfxx: Ok, reused `ContactPresenter`

plugins/board-resources/src/components/popups/EditMember.svelte
lvfxx: Done.

plugins/board-resources/src/components/popups/EditMember.svelte
lvfxx: Done.

PR #1449, opened 2022-04-19
Tracker: move "IssueStatus" enum into model

### LINKED ISSUE #1331: Tracker: Move Issue Priority/State enums into model.
We should move IssuePriority/IssueStatus into model and put localization/icons/etc into them.

ghost:
closes #1331


models/task/src/index.ts
```diff
@@ -463,10 +463,6 @@ export function createModel (builder: Builder): void {
     task.viewlet.Kanban
   )
 
-  builder.mixin(task.class.DoneState, core.class.Class, view.mixin.AttributePresenter, {
```
haiodo, line 471: This change is not related, it will break Recruit.

models/tracker/src/index.ts
```diff
@@ -44,6 +44,30 @@ export { default } from './plugin'
 
 export const DOMAIN_TRACKER = 'tracker' as Domain
 
+/**
+ * @public
+ */
+@Model(tracker.class.IssueStatus, core.class.AttachedDoc, DOMAIN_MODEL)
```
haiodo: It should be not in model domain, but in tracker. Since this kind of objects for every team should be accessible only on need.

models/tracker/src/migration.ts
```diff
@@ -39,10 +72,13 @@ async function createDefaultTeam (tx: TxOperations): Promise<void> {
...
         archived: false,
         identifier: 'TSK',
-        sequence: 0
+        sequence: 0,
+        issueStatuses: 0,
+        defaultIssueStatus: defaultStatusId
       },
       tracker.team.DefaultTeam
     )
+    await createTeamIssueStatuses({ tx, teamId: tracker.team.DefaultTeam, categories, defaultStatusId })
```
haiodo, line 100: You missed case, then team is already created but there is no statues, so you need to check if at least one status exist per every existing team instance, not just for default one.

packages/ui/src/components/SelectPopup.svelte
```diff
@@ -1,5 +1,5 @@
 <!--
-// Copyright © 2020 Anticrm Platform Contributors.
+// Copyright © 2022 Hardcore Engineering Inc.
```
haiodo: Not relevant for this changes.

plugins/tracker-resources/src/components/CreateIssue.svelte
```diff
@@ -40,7 +38,7 @@
     assignee: null,
     number: 0,
     rank: '',
-    status: issueStatus,
+    status: issueStatus ?? ('' as Ref<IssueStatus>),
```
haiodo: I think you should choose first one status instead.

plugins/tracker-resources/src/components/CreateIssue.svelte
```diff
@@ -50,6 +48,26 @@
   const client = getClient()
   const taskId: Ref<Issue> = generateId()
 
+  $: _space = space
+  $: _parent = parent
+  $: if (object.status === '') {
+    getDefaultIssueStatusId(_space).then((statusId) => {
```
haiodo: Better to not mix promises and async/await and use one style. For platform it is preferred to use async/await

plugins/tracker-resources/src/components/issues/Active.svelte
```diff
@@ -1,10 +1,28 @@
...
   import tracker from '../../plugin'
+  import Issues from './Issues.svelte'
 
   export let currentSpace: Ref<Team>
 </script>
 
-<Issues {currentSpace} categories={[IssueStatus.InProgress, IssueStatus.Todo]} title={tracker.string.ActiveIssues} />
+<Issues
+  {currentSpace}
+  statusCategories={new Set([tracker.issueStatusCategory.Unstarted, tracker.issueStatusCategory.Started])}
```
haiodo: Why we need a set here, it would be more nice to pass just an array and construct a set inside component if required.

plugins/tracker-resources/src/components/issues/Board.svelte
```diff
@@ -1,58 +1,52 @@
...
 
   let currentTeam: Team | undefined
-
   $: spaceQuery.query(tracker.class.Team, { _id: currentSpace }, (res) => {
     currentTeam = res.shift()
   })
+
+  let states: TypeState[] | undefined
+  $: if (states === undefined) {
+    getIssueStatuses(currentSpace).then((statuses) => {
```
haiodo: Please use async updateStates function and call it from $:

plugins/tracker-resources/src/components/issues/Issues.svelte
```diff
@@ -50,10 +45,18 @@
...
 
   let currentTeam: Team | undefined
-
   $: spaceQuery.query(tracker.class.Team, { _id: currentSpace }, (res) => {
     currentTeam = res.shift()
   })
+
+  let categories: WithLookup<IssueStatus>[] = []
+  let filteredCategories: WithLookup<IssueStatus>[] = []
+  $: getIssueStatuses(currentSpace).then((statuses) => {
```
haiodo: Please use async function if possible

plugins/tracker/src/index.ts
```diff
@@ -14,28 +14,40 @@
...
  */
-export enum IssueStatus {
-  Backlog,
-  Todo,
-  InProgress,
-  Done,
-  Canceled
+export interface IssueStatusCategory extends Doc {
+  icon: Asset
+  name: string
```
haiodo: Better to use label and IntlString for category title.

haiodo (changes requested):

models/tracker/src/index.ts
ghost: Done.

models/task/src/index.ts
haiodo: it was duplicate, so resolving

packages/ui/src/components/SelectPopup.svelte
ghost: Reverted.

plugins/tracker-resources/src/components/CreateIssue.svelte
ghost: I assign the default status in the `getDefaultIssueStatusId` function. We have no statuses when creating this object.

plugins/tracker-resources/src/components/CreateIssue.svelte
ghost: Done.

plugins/tracker-resources/src/components/issues/Board.svelte
ghost: Done.

plugins/tracker-resources/src/components/issues/Active.svelte
ghost: Done.

plugins/tracker-resources/src/components/issues/Issues.svelte
ghost: Done.

plugins/tracker/src/index.ts
ghost: Done.

models/tracker/src/migration.ts
ghost: I've added migration for this, but for some reason I'm getting an empty response on request for issue status category. Could you please take a look?

models/tracker/src/index.ts
```diff
@@ -141,7 +176,72 @@ export class TDocument extends TDoc implements Document {
...
+    tracker.issueStatusCategory.Backlog
+  )
+
+  builder.createDoc(
+    tracker.class.IssueStatusCategory,
+    core.space.Model,
+    {
+      label: tracker.string.CategoryUnstarted,
+      icon: tracker.icon.CategoryUnstarted,
+      color: 1,
```
haiodo, line 200: I think we could use order for color, and do not have both :-)

plugins/tracker/package.json
```diff
@@ -32,6 +32,7 @@
     "@anticrm/contact": "~0.6.5",
     "@anticrm/chunter": "~0.6.1",
     "@anticrm/attachment": "~0.6.1",
-    "@anticrm/task": "~0.6.0"
+    "@anticrm/task": "~0.6.0",
+    "lexorank": "~1.0.4"
```
haiodo, line 36: Why we need it here? I think we should use it from tracker-resources.

models/tracker/src/index.ts
ghost: I'm not sure about this because the color can be changed later. We cannot guarantee that the color will always match to the correct order of categories.

models/tracker/src/migration.ts
ghost: Fixed. I've used a wrong type of model domain.

plugins/tracker/package.json
ghost: We can't import utils from the `tracker-resources` because it has some dependencies from `@anticrm/tracker`. So, I got this error when trying to make `rushx upgrade`:
```
/Users/ogs/Projects/Xored/anticrm/plugins/tracker-resources/src/components/issues/Active.svelte:15
<script lang="ts">
^

SyntaxError: Unexpected token '<'
    at Object.compileFunction (node:vm:352:18)
```

haiodo (approved):

haiodo (approved):

PR #1486, opened 2022-04-21
Disable strict-boolean-expressions & promise-function-async eslint rules

annano:
Signed-off-by: Anna No [anna.no@xored.com](mailto:anna.no@xored.com)

### DIFF
--- a/packages/platform-rig/profiles/ui/config/eslint.config.json
+++ b/packages/platform-rig/profiles/ui/config/eslint.config.json
@@ -1,44 +1,43 @@
 {
-    "root": true,
-    "env": {
-      "browser": true,
-      "es6": true,
-      "node": true
-    },
-    "extends": [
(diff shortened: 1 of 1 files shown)


packages/platform-rig/profiles/ui/config/eslint.config.json
```diff
@@ -1,44 +1,43 @@
...
-    ],
-    "settings": {
-      "svelte3/typescript": true
     }
+  ],
+  "settings": {
+    "svelte3/typescript": true
+  },
+  "rules": {
+    "@typescript-eslint/strict-boolean-expressions": "off",
```
haiodo, line 40: Why we need this ones? We not turned them by desire.

haiodo (changes requested):
I'm not agree with turning this options off.

PR #1492, opened 2022-04-22
Board: Customize board header

### LINKED ISSUE #1477: Board: Customize board header

Dvinyanin:
Signed-off-by: Dvinyanin Alexandr [dvinyanin.alexandr@gmail.com](mailto:dvinyanin.alexandr@gmail.com)


plugins/board-resources/src/components/BoardHeader.svelte
```diff
@@ -0,0 +1,24 @@
+<script lang="ts">
+    import core, { Ref, Space } from "@anticrm/core";
+    import { Button, showPopup } from "@anticrm/ui";
+    import { createQuery, getClient } from "@anticrm/presentation";
+    import { Header, classIcon } from '@anticrm/workbench-resources'
```
haiodo: It is better to not import workbench* stuff from applications, since we could change it dramatically potentially. If you need some useful components from workbench, it is better to extract them into some platform package and use from it.

Dvinyanin: Done

haiodo (approved):

PR #1533, opened 2022-04-25
Move backlinks to core

### LINKED ISSUE #1517: Platform: Redesign Backlink support
Create type MarkupWithRef in core and move all backlink creation/update logic to server.

We should use MarkupWithRef type for every document wanted backlinks organization.

Create/update operations on references should be performed on server.

annano:


models/core/src/core.ts
```diff
@@ -68,6 +69,13 @@ export class TAttachedDoc extends TDoc implements AttachedDoc {
   collection!: string
 }
 
+@Model(core.class.Backlink, core.class.AttachedDoc, DOMAIN_MODEL)
```
haiodo: Backlinks should leave not in model

packages/core/src/classes.ts
```diff
@@ -267,3 +267,21 @@ export interface Version extends Doc {
...
+  // A target document class
+  // attachedToClass
+
+  message: string
+  // Source document we have reference from, it should be parent document for Comment/Message.
+  backlinkId: Ref<Doc>
+  // Source document class
+  backlinkClass: Ref<Class<Doc>>
+  // Reference to comment documentId
+  attachedDocId?: Ref<Doc>
```
haiodo, line 287: Why we need this one?

annano: It is a comment id, for example, if we are adding comment to a card, `backlinkId` would be card id, `attachedDocId` would be a comment id

models/core/src/core.ts
annano: Updated to `DOMAIN_BACKLINK`

models/chunter/src/migration.ts
```diff
@@ -158,10 +166,49 @@ export async function migrateThreadMessages (client: MigrationClient): Promise<v
   }
 }
 
+export async function migrateBacklinks (client: MigrationClient): Promise<void> {
+  const _class = 'chunter:class:Backlink' as Ref<Class<ChunterBacklink>>
+  const backlinks = await client.find<ChunterBacklink>(DOMAIN_COMMENT, { _class })
+  for (const backlink of backlinks) {
+    await client.delete(DOMAIN_COMMENT, backlink._id)
```
haiodo, line 173: Why not use client.move?

models/chunter/src/migration.ts
```diff
@@ -158,10 +166,49 @@ export async function migrateThreadMessages (client: MigrationClient): Promise<v
...
+      message: backlink.message,
+      backlinkId: backlink.backlinkId,
+      backlinkClass: backlink.backlinkClass,
+      attachedDocId: backlink.attachedDocId,
+      modifiedBy: backlink.modifiedBy,
+      modifiedOn: backlink.modifiedOn
+    })
+  }
+
+  const classTxes = await client.find<TxCreateDoc<Comment>>(DOMAIN_TX, { 'tx.objectClass': 'chunter:class:Backlink' })
```
haiodo, line 190: Please use client.update instead

models/chunter/src/migration.ts
```diff
@@ -158,10 +166,49 @@ export async function migrateThreadMessages (client: MigrationClient): Promise<v
...
+  for (const tx of classTxes) {
+    await client.update(
+      DOMAIN_TX,
+      { _id: tx._id },
+      { 'tx.objectClass': core.class.Backlink }
+    )
+  }
+
+  const spaceTxes = await client.find<TxCreateDoc<Comment>>(DOMAIN_TX, { 'tx.objectSpace': 'chunter:space:Backlinks' })
```
haiodo, line 195: Same here.

packages/core/src/classes.ts
haiodo: 1. We could add Backlink to card if card has field marked with MarkupWithRef, like description and only backlinkId, backlinkClass is required.
2. If Card has comment, comment itself will be backlinkId and backlinkClass.
3. Same for any checkitemlist/checkitem attached to card.
So I do not see this field is required.

haiodo (changes requested):

PR #1539, opened 2022-04-26
Tracker: Add keyboard support for issues list

agrigorovich-xored:


haiodo (approved):

plugins/tracker-resources/src/components/issues/IssuesList.svelte
```diff
@@ -93,94 +150,145 @@
...
+          <span class="eLabelCounter ml-2">{(groupedIssues[category] ?? []).length}</span>
+        </div>
+        <div class="flex mr-1">
+          <Tooltip label={tracker.string.AddIssueTooltip} direction={'left'}>
+            <Button icon={IconAdd} kind={'transparent'} on:click={(event) => handleNewIssueAdded(event, category)} />
+          </Tooltip>
+        </div>
+      </div>
+    {/if}
+    <Scroller>
```
haiodo: I do not think we need scroller here, since it for tables only.

agrigorovich-xored: Done.

haiodo (approved):

PR #1554, opened 2022-04-26
Fix backlinks ut

annano:


haiodo (changes requested):
We should not update transactions until migrations will be applied to production.

PR #1556, opened 2022-04-26
New panels

SasLord:


haiodo:
Does telegram still work with this PR? Look nice.

haiodo (approved):

PR #1558, opened 2022-04-27
Board: Handle attachments when copying card

### LINKED ISSUE #1294: Card: Copy

lvfxx:
Closes #1294

### DIFF
--- a/plugins/board-resources/src/components/popups/CopyCard.svelte
+++ b/plugins/board-resources/src/components/popups/CopyCard.svelte
@@ -1,7 +1,9 @@
 <script lang="ts">
+  import attachment, { Attachment } from '@anticrm/attachment'
+  import chunter, { Comment } from '@anticrm/chunter'
   import { createEventDispatcher, onMount } from 'svelte'
-  import { Label, Button, Status as StatusControl, TextArea } from '@anticrm/ui'
-  import { Class, Client, Doc, Ref } from '@anticrm/core'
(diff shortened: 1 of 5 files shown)


plugins/board-resources/src/components/popups/CopyCard.svelte
```diff
@@ -65,6 +75,83 @@
...
+    if (keepAttachments) {
+      const sourceAttachments = await client.findAll(attachment.class.Attachment, {
+        space: object.space,
+        attachedTo: object._id
+      })
+
+      for (let i = 0; i < sourceAttachments.total; i++) {
+        const a = sourceAttachments[i]
+
+        await client.addCollection(
```
haiodo, line 88: We have big trouble here, we need to have API to copy or increment reference on S3/minio for particular file, since it could be deleted if original card will be deleted, since Attachment object is just a descriptor for a real document inside file storage. Let's discuss on our general meeting.

plugins/board-resources/src/components/popups/CopyCard.svelte
```diff
@@ -65,6 +75,83 @@
...
+    if (keepComments) {
+      const sourceComments = await client.findAll(chunter.class.Comment, {
+        space: object.space,
+        attachedTo: object._id
+      })
+
+      for (let i = 0; i < sourceComments.total; i++) {
+        const comment = sourceComments[i]
+
+        const commentAttachments =
```
haiodo, line 114: Comments could have attachments by itself, so it it better to have some logic based on model and collection types. instead of individual collections.

PR #1567, opened 2022-04-27
Board: Checklists model

annano:


plugins/board/src/index.ts
```diff
@@ -56,6 +56,23 @@ export interface CardDate extends Obj {
...
+  assignee?: Ref<Employee>
+  dueDate?: Timestamp
+  isChecked: boolean
+  name: Markup
+}
+
+/**
+ * @public
+ */
+export interface CardChecklist extends Obj {
```
BykhovDenis, line 72: Why you extend Obj?

plugins/board/src/index.ts
```diff
@@ -56,6 +56,23 @@ export interface CardDate extends Obj {
   startDate?: Timestamp
 }
 
+/**
+ * @public
+ */
+export interface CardChecklistItem extends Obj {
```
BykhovDenis, line 62: Why you extend Obj?

models/board/src/index.ts
```diff
@@ -70,6 +70,12 @@ export class TCardLabel extends TAttachedDoc implements CardLabel {
   isHidden?: boolean
 }
 
+@Model(board.class.CardChecklist, core.class.Obj, DOMAIN_MODEL)
+@UX(board.string.Checklist)
+export class TCardChecklist extends TObj implements CardChecklist {
```
BykhovDenis, line 75: Why you need it in model?

BykhovDenis (changes requested):

models/board/src/index.ts
annano: We would need to add presenters for this in the future to display it in activity

models/board/src/index.ts
```diff
@@ -105,6 +111,9 @@ export class TCard extends TTask implements Card {
 
   @Prop(Collection(contact.class.Employee), board.string.Members)
   members?: Ref<Employee>[]
+
+  @Prop(Collection(board.class.CardChecklist), board.string.Checklist)
```
BykhovDenis, line 115: It should be 
```
@Prop(ArrOf(TypeCardChecklist()), board.string.Checklist)


@Model(contact.class.TypeCardChecklist, core.class.Type)
export class TTypeCardChecklist extends TType {}

export function TypeCardChecklist (): Type<CardCheckList> {
  return { _class: board.class.TypeCardChecklist, label: board.string.Checklist }
}
 ```

PR #1578, opened 2022-04-28
Tracker: Fix IssuesList selection

### LINKED ISSUE #1569: Tracker: IssuesList incorrect selection
1. `--menu-bg-select` color should be used for selected elements.
2. `selection` is not updated on mouse move.

agrigorovich-xored:
Closes #1569

### DIFF
--- a/plugins/tracker-resources/src/components/issues/IssuesList.svelte
+++ b/plugins/tracker-resources/src/components/issues/IssuesList.svelte
@@ -305,7 +305,7 @@
     border-bottom: 1px solid var(--theme-button-border-hovered);
 
     &.mListGridChecked {
-      background-color: var(--menu-bg-select);
+      background-color: var(--theme-table-bg-hover);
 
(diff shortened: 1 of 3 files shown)


plugins/tracker-resources/src/components/issues/IssuesList.svelte
```diff
@@ -305,7 +305,7 @@
     border-bottom: 1px solid var(--theme-button-border-hovered);
 
     &.mListGridChecked {
-      background-color: var(--theme-table-bg-hover);
+      background-color: var(--menu-bg-select);
```
haiodo: For table we have following styles.

&.checking { background-color: var(--theme-table-bg-hover); }
&.selected { background-color: var(--menu-bg-select); }

Let's use

agrigorovich-xored: Done.

haiodo (approved):

PR #1620, opened 2022-05-02
Board: Fix card members update

### LINKED ISSUE #1561: [Bug] Use $push/$pull for UserBoxList component (instead of update)

Dvinyanin:
Signed-off-by: Dvinyanin Alexandr [dvinyanin.alexandr@gmail.com](mailto:dvinyanin.alexandr@gmail.com)

### DIFF
--- a/plugins/board-resources/src/utils/CardUtils.ts
+++ b/plugins/board-resources/src/utils/CardUtils.ts
@@ -68,19 +68,14 @@ export function unarchiveCard (card: Card, client: Client): Promise<TxResult> | 
   return updateCard(client, card, 'isArchived', false)
 }
 
-export function updateCardMembers (
-  card: Card,
-  client: Client,
-  users: Array<Ref<Employee>>
-): Array<Promise<TxResult>> | undefined {
+export function updateCardMembers (card: Card, client: Client, users: Array<Ref<Employee>>): void {
   if (card?.members == null) return
   const { members } = card
-  return [
-    ...members
-      .filter((member) => !users.includes(member))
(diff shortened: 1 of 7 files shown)


plugins/board-resources/src/utils/CardUtils.ts
```diff
@@ -67,3 +67,20 @@ export function archiveCard (card: Card, client: Client): Promise<TxResult> | un
 export function unarchiveCard (card: Card, client: Client): Promise<TxResult> | undefined {
   return updateCard(client, card, 'isArchived', false)
 }
+
+export function updateCardMembers (
```
haiodo: Better to use one transaction with both $push: { members: [newMembers] } and $pull: { members: [oldMemebrs]}. Please check if it working, or at least two transactions instead of multiple ones. 
Also pleas make function async and return Promise<void>, it will be faster and less transactions.

Dvinyanin: `$pull` doesn't support array
https://github.com/hcengineering/anticrm/blob/eb9526f0153f10550c1eb537dd8c176d31921cf8/packages/core/src/operator.ts#L42
Also using both operators cause
`ERROR: platform:status:UnknownError {"message":"Updating the path 'members' would create a conflict at 'members'"}`

haiodo: I suppose $pull: {values: {$in: array}} should work, at least for mongoDb, 
if so we could fix it for our $pull implementation.

Dvinyanin: Done

haiodo (approved):

PR #1640, opened 2022-05-05
Tracker: Issue filters - main functionality

agrigorovich-xored:


plugins/tracker-resources/src/components/issues/Issues.svelte
```diff
@@ -50,19 +53,25 @@
...
+  const dispatch = createEventDispatcher()
   const ENTRIES_LIMIT = 200
   const spaceQuery = createQuery()
   const issuesQuery = createQuery()
   const resultIssuesQuery = createQuery()
   const statusesQuery = createQuery()
   const issuesMap: { [status: string]: number } = {}
+
+  let filterElement: HTMLElement | null = null
+  let filters: { [p: string]: any[] } = {}
```
haiodo, line 65: Why not have array of filters?
I would prefer to have arrays with typed DocumentQuery inside.
Also assume you will need to add support for $or as next step.

haiodo (approved):

haiodo (approved):

PR #1650, opened 2022-05-05
Custom attributes

### LINKED ISSUE #1587: Custom attributes

BykhovDenis:
close #1587 
Signed-off-by: Denis Bykhov <80476319+BykhovDenis@users.noreply.github.com>

### DIFF
--- a/packages/model/src/dsl.ts
+++ b/packages/model/src/dsl.ts
@@ -338,96 +338,76 @@ export class Builder {
 /**
  * @public
  */
-export function getType<T extends Type<any>> (
-  _class: Ref<Class<T>>,
-  label: IntlString,
-  attrs: Record<string, any> = {}
-): T {
-  // eslint-disable-next-line
-  return {
-    space: core.space.Model,
-    modifiedBy: core.account.System,
-    modifiedOn: Date.now(),
-    _id: generateId(),
-    _class,
-    label,
-    ...attrs
-  } as T
-}
-
-/**
- * @public
- */
 export function TypeString (): Type<string> {
-  return getType(core.class.TypeString, core.string.String)
+  return { _class: core.class.TypeString, label: core.string.String }
 }
 
 /**
  * @public
  */
 export function TypeNumber (): Type<number> {
-  return getType(core.class.TypeNumber, core.string.Number)
+  return { _class: core.class.TypeNumber, label: core.string.Number }
 }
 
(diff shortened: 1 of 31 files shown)


packages/model/src/dsl.ts
```diff
@@ -335,79 +335,99 @@ export class Builder {
...
+  _class: Ref<Class<T>>,
+  label: IntlString,
+  attrs: Record<string, any> = {}
+): T {
+  // eslint-disable-next-line
+  return {
+    space: core.space.Model,
+    modifiedBy: core.account.System,
+    modifiedOn: Date.now(),
+    _id: generateId(),
```
haiodo: Why we need id?

plugins/activity-resources/src/components/utils.ts
```diff
@@ -37,14 +37,21 @@ async function createPseudoViewlet (
   // Check if it is attached doc and collection have title override.
   const presenter = await getObjectPresenter(client, doc._class, { key: 'doc-presenter' })
   if (presenter !== undefined) {
+    let collection = ''
+    if (dtx.collectionAttribute?.label !== undefined) {
+      if (dtx.collectionAttribute.isCustom) {
```
haiodo: If we have embedded why we need this logic? And it looks wrong

haiodo:
Why we need isCustom?

plugins/view-resources/src/components/CreateAttribute.svelte
```diff
@@ -0,0 +1,106 @@
...
+  export let _class: Ref<Class<Space>>
+  let name: string
+  let type: Type<PropertyType> | undefined
+  let is: AnyComponent | undefined
+  const client = getClient()
+  const hierarchy = client.getHierarchy()
+  const dispatch = createEventDispatcher()
+
+  async function save (): Promise<void> {
+    if (type === undefined) return
```
haiodo, line 33: Please show confirmation dialog and inform about changes will not be possible to perform for type and field name in document.

plugins/activity-resources/src/components/utils.ts
BykhovDenis: It's outdated, Thanks

BykhovDenis:
`Why we need isCustom?` I left it for possible change label in future

haiodo (approved):

PR #1652, opened 2022-05-05
Board: Add CardCover

### LINKED ISSUE #1629: Card: Cover support

Dvinyanin:

### DIFF
--- a/plugins/board/src/index.ts
+++ b/plugins/board/src/index.ts
@@ -59,6 +59,15 @@ export interface CardDate extends Obj {
 /**
  * @public
  */
+export interface CardCover {
+  color?: number
+  image?: string
+  size: 'large' | 'small'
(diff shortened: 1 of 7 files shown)


plugins/board/src/index.ts
```diff
@@ -73,8 +82,7 @@ export interface Card extends Task {
 
   location?: string
 
-  coverColor?: number
```
haiodo, line 76: Do not see migration for this fields.

Dvinyanin: Why should we add migration for this field? 
It has not been used and there was no way to update it

haiodo: We should if it was used before, but ok for now.

haiodo (approved):

PR #1718, opened 2022-05-11
Board: Fix push/pull activity

### LINKED ISSUE #1519: [Bug] Card lables: Adding/removing label from card is not displayed in history (could be because of $push/$pull operations)

Dvinyanin:


plugins/activity-resources/src/components/utils.ts
```diff
@@ -164,11 +166,30 @@ function getHiddenAttrs (client: TxOperations, _class: Ref<Class<Doc>>): Set<str
...
+    added: utx.$push?.[m.key],
+    removed: utx.$pull?.[m.key]
+  }
+  if (value.set !== undefined) {
+    value.set = await getRealValue(value.set)
   }
-  return val
+  if (value.added !== undefined) {
+    value.added = Array.isArray(value.added.$each)
+      ? (value.added.$each as any[]).map(getRealValue)
```
haiodo: Better to findAll with {_id: {$In: array}}

plugins/activity-resources/src/components/utils.ts
```diff
@@ -164,11 +166,30 @@ function getHiddenAttrs (client: TxOperations, _class: Ref<Class<Doc>>): Set<str
...
-  const val = utx[m.key]
-
-  if (client.getHierarchy().isDerived(m._class, core.class.Doc) && typeof val === 'string') {
-    // We have an reference, we need to find a real object to pass for presenter
-    return await client.findOne(m._class, { _id: val as Ref<Doc> })
+  async function getRealValue (value: any): Promise<any> {
+    if (client.getHierarchy().isDerived(m._class, core.class.Doc) && typeof value === 'string') {
+      // We have an reference, we need to find a real object to pass for presenter
+      return await client.findOne(m._class, { _id: value as Ref<Doc> })
+    }
```
haiodo: value should be returned if not derived from Doc

Dvinyanin: Fixed

plugins/activity-resources/src/components/utils.ts
Dvinyanin: Done

plugins/activity-resources/src/components/utils.ts
```diff
@@ -164,11 +166,26 @@ function getHiddenAttrs (client: TxOperations, _class: Ref<Class<Doc>>): Set<str
...
+    removed: utx.$pull?.[m.key]
+  }
+  if (value.set !== undefined) {
+    ;[value.set] = await getAllRealValues([value.set])
   }
-  return val
+  if (value.added !== undefined) {
+    value.added = await getAllRealValues(Array.isArray(value.added.$each) ? value.added.$each : [value.added])
+  }
+  if (value.removed !== undefined) {
```
haiodo: Removed will not be located from snapshot storage and need to be restored from transactions. Please see how document it restored from transactions in activity.

Dvinyanin: It is only for `$pull` operation in `TxUpdateDoc`
But if `attachedDoc` was deleted, presenter will not be shown for all `txes`

haiodo (approved):

PR #1730, opened 2022-05-12
Board: Add labels mode preference

### LINKED ISSUE #1572: Preferences: Labels compact mode

Dvinyanin:


plugins/board/src/index.ts
```diff
@@ -111,6 +112,12 @@ export interface MenuPage extends Doc {
   label: IntlString
 }
 
+/**
+ * @public
+ */
+export interface LabelsCompactMode extends Preference {
```
annano, line 118: I am not quite sure how this preference mechanism is working, but shouldn't we have just a preference for Board application per each user and store all possible preferences there?

Dvinyanin: As I understand, preference works with users out of the box, I just marked boards with compact mode

haiodo: Preference object are individual per user, so any kind of model is possible.

haiodo (approved):

PR #1768, opened 2022-05-16
Telegram: latest messages below. Update AttachmentPreview layout.

SasLord:
Signed-off-by: Alexander Platov <sas_lord@mail.ru>

### DIFF
--- a/plugins/telegram-resources/src/components/Chat.svelte
+++ b/plugins/telegram-resources/src/components/Chat.svelte
@@ -82,7 +82,7 @@
         updateAccountsQuery(accountsIds)
       },
       {
-        sort: { sendOn: SortingOrder.Descending },
+        sort: { sendOn: SortingOrder.Ascending },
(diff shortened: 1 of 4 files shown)


plugins/telegram-resources/src/components/Chat.svelte
```diff
@@ -82,7 +82,7 @@
         updateAccountsQuery(accountsIds)
       },
       {
-        sort: { sendOn: SortingOrder.Descending },
+        sort: { sendOn: SortingOrder.Ascending },
```
BykhovDenis: You should get last 500 messages, not first. It should be `Descending`, just use `messages = res.reverse()` in 77

BykhovDenis (changes requested):

haiodo (approved):

BykhovDenis (approved):

haiodo (approved):

PR #1847, opened 2022-05-24
Update Filters layout

SasLord:


plugins/view-resources/src/components/filter/FilterBar.svelte
```diff
@@ -24,23 +24,26 @@
 
   export let _class: Ref<Class<Doc>>
   export let query: DocumentQuery<Doc>
+  export let filters: Filter[] | undefined
```
BykhovDenis: Why not `export let filters: Filter[] = []` and remove extra checks

BykhovDenis (approved):

PR #1859, opened 2022-05-24
Board: Update actions

### LINKED ISSUE #1848: Fix card actions

annano:


packages/ui/src/types.ts
```diff
@@ -74,7 +75,22 @@ export interface PopupPositionElement {
     h: HorizontalAlignment
   }
 }
+
 export type PopupPosAlignment = 'right' | 'top' | 'float' | 'account' | 'full' | 'content' | 'middle'
+
+export function isPopupPosAlignment (x: any): x is PopupPosAlignment {
+  return (
```
Dvinyanin, line 82: Minor: might be better?
```suggestion
  return ['right', 'top', 'float', 'account', 'full', 'content', 'middle'].includes(x)
```
or even
```suggestion
const POPUP_POS_ALIGMENTS = ['right', 'top', 'float', 'account', 'full', 'content', 'middle'] as const
export type PopupPosAlignment = typeof POPUP_POS_ALIGMENTS[number]
export function isPopupPosAlignment (x: any): x is PopupPosAlignment {
  return POPUP_POS_ALIGMENTS.includes(x)
}
```

haiodo: Let's leave as is.

haiodo (approved):

PR #2124, opened 2022-06-21
Chunter: Open message links without reload

denis-bunakalya:
[tsk-177](https://front.hc.engineering/workbench%3Acomponent%3AWorkbenchApp/tracker%3Aapp%3ATracker/tracker%3Ateam%3ADefaultTeam/projects/629d9558d1a401ec31f59d8e#tracker%3Acomponent%3AEditIssue%7C62b1938e839f85d8716b6525%7Ctracker%3Aclass%3AIssue%7Ccontent)

### DIFF
--- a/packages/presentation/src/components/message/Nodes.svelte
+++ b/packages/presentation/src/components/message/Nodes.svelte
@@ -14,7 +14,7 @@
 // limitations under the License.
 -->
 <script lang="ts">
-  import { CheckBox, Component } from '@anticrm/ui'
+  import { CheckBox, Component, navigate, parseLocation } from '@anticrm/ui'
   import view from '@anticrm/view'
 
   export let nodes: NodeListOf<any>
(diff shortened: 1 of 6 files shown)


packages/presentation/src/components/message/Nodes.svelte
```diff
@@ -63,8 +63,19 @@
     {:else if node.nodeName === 'DIV'}
       <div><svelte:self nodes={node.childNodes} /></div>
     {:else if node.nodeName === 'A'}
-      <a href={node.getAttribute('href')} target={node.getAttribute('target')}
-        ><svelte:self nodes={node.childNodes} /></a
+      <a
+        href={node.getAttribute('href')}
+        target={node.getAttribute('target')}
+        on:click={(e) => {
```
haiodo, line 69: I think we need to check why we handle page new url and reload instead of update in Workbench itself.

denis-bunakalya: It's related to the `target` property, which is `_blank` by default for links created by tiptap, i.e. open in new tab
If we remove that prop, then links will be opened in the same tab but with reload, this is the default behavior of `<a>` tag
That's why we handle click and prevent default behavior if needed

BykhovDenis (approved):

PR #2375, opened 2022-11-14
Allow to create new team

muhtimur:
Signed-off-by: muhtimur <timur.mukhamedishin@xored.com>

# Contribution checklist

## Brief description

Now we can create teams

image

## Checklist

* [x] - Does it well tested?
* [x] - Tested for Chrome.
* [x] - Tested for Safari.

## Related issues

[TSK-108](https://front.hc.engineering/workbench/platform/tracker/my-issues#tracker%3Acomponent%3AEditIssue%7C62a8296e4b1f5cf4dffd018e%7Ctracker%3Aclass%3AIssue%7Ccontent)


packages/core/src/classes.ts
```diff
@@ -271,6 +271,7 @@ export interface Space extends Doc {
   private: boolean
   members: Arr<Ref<Account>>
   archived: boolean
+  icon?: Asset
```
haiodo: We do not need it for generic spaces.

muhtimur: We discussed it with Denis, and he told me that the easiest way to add a custom icon is to create this field in Space interface. But there is one more way -- to create a mixin for Space which will show some presenter with the custom icon. So, is it better to use 2nd way?

haiodo: Is it only will be used for Team? If so please use it there, and have options for Space unified components. No need for mixin here.

haiodo: We should not mix Core and UI in one.

muhtimur: Now it is used only for teams but in future we will have to use it for userpics in direct messages

muhtimur: Added a presenter component for Team

packages/core/src/classes.ts
```diff
@@ -271,6 +271,7 @@ export interface Space extends Doc {
   private: boolean
   members: Arr<Ref<Account>>
   archived: boolean
+  presenter?: AnyAttribute
```
haiodo: Why we need to extend base Space class?

haiodo: Please use some of view.mixin.* if you need presenters.

muhtimur: Corrected it

haiodo (approved):

PR #2400, opened 2022-11-29
TSK-343: Remember unfinished comment per document

disant1:
Draft.
# Contribution checklist

## Checklist

* [ ] - Does it well tested?
* [x] - Tested for Chrome.
* [x] - Tested for Safari.

## Related issues


plugins/chunter-resources/src/components/CommentInput.svelte
```diff
@@ -17,14 +17,84 @@
...
+  async function updateCommentInput (draftComment: Comment | undefined) {
+    inputContent = draftComment ? draftComment.message : ''
+  }
+
+  function commentIsEmpty (message: string, attachments: number): boolean {
+    return !(message.length > 0 || attachments > 0)
+  }
+
+  async function updateDraft (object: Doc) {
+    const drafts = fetchMetadataLocalStorage(chunter.metadata.CommentDrafts) ?? {}
```
haiodo: Better to use svelte writable and do fetch only once and save on draft update.

packages/text-editor/src/components/TextEditor.svelte
```diff
@@ -48,7 +48,11 @@
       dispatch('content', content)
     }
   }
-
+  export function setContent (newContent: string): void {
+    content = newContent
+    editor.commands.setContent(content)
```
haiodo: We will lose Undo history, please update only if content is different.

plugins/attachment-resources/src/components/AttachmentRefInput.svelte
```diff
@@ -39,24 +43,61 @@
 
   const client = getClient()
   const query = createQuery()
+  const drafts = writable<Record<Ref<Doc>, Record<Ref<Attachment>, Attachment>>>()
```
haiodo: Please move out of component, so it could be reused, over-wise it will be alive only then component is alive.

plugins/attachment-resources/src/components/AttachmentRefInput.svelte
```diff
@@ -39,24 +43,61 @@
...
-        attachedTo: objectId
-      },
-      (res) => {
-        originalAttachments = new Set(res.map((p) => p._id))
-        attachments = new Map(res.map((p) => [p._id, p]))
+  $: objectId && getAttachments(objectId)
+
+  async function getAttachments (objectId: Ref<Doc>) {
+    if (shouldUseDraft) {
+      drafts.set(fetchMetadataLocalStorage(attachment.metadata.AttachmentDrafts) ?? {})
```
haiodo: Please cancel query in this case, since if component will be reused with different props it will work incorrectly.

plugins/chunter-resources/src/components/CommentInput.svelte
```diff
@@ -17,14 +17,105 @@
...
   import { createBacklinks } from '../backlinks'
   import chunter from '../plugin'
 
-  const client = getClient()
   export let object: Doc
+  export let shouldUseDraft: boolean = false
+
+  const client = getClient()
   const _class = chunter.class.Comment
+  const drafts = writable<Record<string, Comment>>()
```
haiodo: Better to have a writable with <Id, DraftDoc> and DraftDoc should contain any of document fields and attachments as well, as any other serialisable fields.

So it could be easy reused lately if more document drafts will be required, and chaning only this part will be possible to move all drafts on server DB>

haiodo (changes requested):

packages/text-editor/src/components/TextEditor.svelte
disant1: Fixed, but we have one bug - TipTap's setContent() cuts spaces from content at the beginning and at the end.

plugins/chunter-resources/src/components/CommentInput.svelte
disant1: Done

packages/text-editor/src/components/TextEditor.svelte
disant1: This bug also fixed

plugins/attachment-resources/src/components/AttachmentRefInput.svelte
disant1: Done

haiodo (approved):

PR #2592, opened 2023-02-06
Low level bulk operations

haiodo:
Signed-off-by: Andrey Sobolev <haiodo@gmail.com>

# Contribution checklist

## Brief description

* Rework server storage and adapters to support bulk operations.
* Fix deletion of attached documents + performance speedup.

## Checklist

* [x] - UI test added to added/changed functionality?
* [x] - Does Changelog.md is updated with changes?
* [x] - Does it well tested?
* [x] - Tested for Chrome.
* [x] - Tested for Safari.

## Related issues


plugins/view-resources/src/components/EditDoc.svelte
```diff
@@ -70,7 +70,9 @@
   $: if (_id && _class) {
     query.query(_class, { _id }, (result) => {
       object = result[0]
-      realObjectClass = object._class
+      if (object !== undefined) {
+        realObjectClass = object?._class
```
BykhovDenis: `if` or `?` is extra

dev/storage/src/storage.ts
```diff
@@ -44,8 +44,15 @@ class InMemoryTxAdapter extends DummyDbAdapter implements TxAdapter {
...
 
-  tx (tx: Tx): Promise<TxResult> {
-    return this.txdb.tx(tx)
+  async tx (...tx: Tx[]): Promise<TxResult> {
+    const r: TxResult[] = []
+    for (const t of tx) {
+      r.push(await this.txdb.tx(t))
+    }
+    if (r.length === 0) {
+      return r[0]
```
BykhovDenis, line 53: Undefined? Probably it should be `if (r.length === 1)`, also function return type should be Promise<TxResult | TxResult[]>

server/core/src/storage.ts
```diff
@@ -138,16 +139,50 @@ class TServerStorage implements ServerStorage {
...
+        // Skip unsupported tx
+        console.error('Unsupported transacton', tx)
+        continue
+      }
       const domain = this.hierarchy.getDomain(txCUD.objectClass)
-      const adapter = this.getAdapter(domain)
-      const res = await adapter.tx(txCUD)
-      return res
-    } else {
+      if (part.length > 0) {
```
BykhovDenis, line 164:
```suggestion
     if (lastDomain !== domain && part.length > 0) {
await processPart()
}
lastDomain = domain
part.push(txCUD)
```

server/core/src/storage.ts
```diff
@@ -299,73 +342,140 @@ class TServerStorage implements ServerStorage {
...
+          { sort: { modifiedOn: 1 } }
+        )
+        : []
+    const result: Doc[] = []
+
+    const txesAttach =
+      removeAttachObjectIds.length > 0
+        ? await findAll<TxCollectionCUD<Doc, AttachedDoc>>(
+          ctx,
+          core.class.TxCollectionCUD,
```
BykhovDenis, line 377: I'm not sure that all AttachedDoc CUD ties were used CollectionTx

server/core/src/storage.ts
```diff
@@ -299,73 +342,140 @@ class TServerStorage implements ServerStorage {
...
+        continue
+      }
+      const rtx = actualTx as TxRemoveDoc<Doc>
+      const isAttached = this.hierarchy.isDerived(rtx.objectClass, core.class.AttachedDoc)
+
+      const objTxex = isAttached
+        ? txesAttach.filter((tx) => tx.tx.objectId === rtx.objectId)
+        : txes.filter((it) => it.objectId === rtx.objectId)
+
+      const createTx = isAttached
```
BykhovDenis: We have `TxProcessor.buildDoc2Doc` for next 15 lines

server/core/src/storage.ts
```diff
@@ -299,73 +342,140 @@ class TServerStorage implements ServerStorage {
...
+        ? await findAll<TxCollectionCUD<Doc, AttachedDoc>>(
+          ctx,
+          core.class.TxCollectionCUD,
+          { 'tx.objectId': { $in: removeAttachObjectIds } },
+          { sort: { modifiedOn: 1 } }
+        )
+        : []
+
+    for (const tx of rawTxes) {
+      const actualTx = TxProcessor.extractTx(tx)
```
BykhovDenis: We do it twice, I think rawTxes not big array and it's ok, but probably better create new array for store rtx and isAttached in first loop and use this array for second loop

server/core/src/storage.ts
```diff
@@ -395,39 +512,48 @@ class TServerStorage implements ServerStorage {
...
+    for (const tx of txes) {
+      const actualTx = TxProcessor.extractTx(tx)
+      if (!this.hierarchy.isDerived(actualTx._class, core.class.TxUpdateDoc)) {
+        continue
+      }
+      const rtx = actualTx as TxUpdateDoc<Doc>
+      if (rtx.operations.space === undefined || rtx.operations.space === rtx.objectSpace) {
+        continue
+      }
+      const factory = new TxFactory(core.account.System)
```
BykhovDenis: I think it should be `tx.modifiedBy`

BykhovDenis (approved):

server/core/src/storage.ts
haiodo: They should be, before it was same logic. So better they to be :-)

server/core/src/storage.ts
haiodo: Thanks.

server/core/src/storage.ts
haiodo: Fixed.

server/core/src/storage.ts
haiodo: Fixed.

server/core/src/storage.ts
haiodo: I do no think this check is really required, in place there it is needed it checked.

PR #2612, opened 2023-02-09
Assign project when assigning sprint

ThetaDR:
# Contribution checklist

## Brief description
https://front.hc.engineering/workbench/platform/tracker/TSK-551

## Checklist

* [ ] - Does it well tested?
* [x] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

### DIFF
--- a/plugins/tracker-resources/src/components/CreateIssue.svelte
+++ b/plugins/tracker-resources/src/components/CreateIssue.svelte
@@ -616,12 +616,17 @@
     object = { ...object, project: projectId }
   }
 
-  const handleSprintIdChanged = (sprintId: Ref<Sprint> | null | undefined) => {
+  const handleSprintIdChanged = async (sprintId: Ref<Sprint> | null | undefined) => {
     if (sprintId === undefined) {
       return
     }
+    let projectSprintId: Ref<Project>
+    if (sprintId !== null) {
+      const sprint = await client.findOne(tracker.class.Sprint, { _id: sprintId })
+      projectSprintId = sprint.project
+    } else projectSprintId = null
 
-    object = { ...object, sprint: sprintId }
+    object = { ...object, sprint: sprintId, project: projectSprintId }
   }
 
   function addTagRef (tag: TagElement): void {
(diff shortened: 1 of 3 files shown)


plugins/tracker-resources/src/components/CreateIssue.svelte
```diff
@@ -616,12 +616,17 @@
     object = { ...object, project: projectId }
   }
 
-  const handleSprintIdChanged = (sprintId: Ref<Sprint> | null | undefined) => {
+  const handleSprintIdChanged = async (sprintId: Ref<Sprint> | null | undefined) => {
     if (sprintId === undefined) {
       return
     }
+    let projectSprintId: Ref<Project>
+    if (sprintId !== null) {
```
haiodo: !== undefined or != null to check for both cases.

server-plugins/tracker-resources/src/index.ts
```diff
@@ -336,6 +336,28 @@ async function doIssueUpdate (
...
+      )
+      res.push(
+        control.txFactory.createTxUpdateDoc(updateTx.objectClass, updateTx.objectSpace, updateTx.objectId, {
+          project: sprint.project
+        })
+      )
+    } else {
+      res.push(
+        control.txFactory.createTxUpdateDoc(updateTx.objectClass, updateTx.objectSpace, updateTx.objectId, {
+          project: null
```
haiodo, line 351: Why you remove project for any non sprint change operation?

ThetaDR: `if (Object.prototype.hasOwnProperty.call(updateTx.operations, 'sprint'))` prevents this, so we update project this way only if the sprint was changed

haiodo: Why we need second else in this case?

ThetaDR: In case user sets "No sprint". So that it sets both sprint and project to null.
Is it supposed to set project only when the user sets any existing sprint or when it's set to "No sprint" too?

haiodo (approved):

PR #2727, opened 2023-03-13
Allow to preselect users

denis-bunakalya:
# Contribution checklist

## Checklist

* [x] - Does it well tested?
* [x] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

### DIFF
--- a/packages/presentation/src/components/ObjectPopup.svelte
+++ b/packages/presentation/src/components/ObjectPopup.svelte
@@ -58,11 +58,12 @@
 
   export let create: ObjectCreate | undefined = undefined
   export let readonly = false
+  export let preselect: ((itemId: Ref<Doc>, select: () => void) => void) | undefined = undefined
 
   let search: string = ''
   let objects: Doc[] = []
 
-  $: selectedElements = new Set(selectedObjects)
+  let selectedElements = new Set(selectedObjects)
 
   const dispatch = createEventDispatcher()
   const query = createQuery()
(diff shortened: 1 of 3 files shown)


packages/presentation/src/components/ObjectPopup.svelte
```diff
@@ -58,11 +58,12 @@
 
   export let create: ObjectCreate | undefined = undefined
   export let readonly = false
+  export let preselect: ((itemId: Ref<Doc>, select: () => void) => void) | undefined = undefined
```
haiodo, line 61: Why we need it? It looks very strange, please use selected/selectedObjects or do a custom component.

denis-bunakalya: We need it to perform some checks before we actually mark or unmark an item. In my example I need to show a confirmation dialog and only mark the clicked item if the user confirmed the action in the dialog. 

It's not possible to use `selected/selectedObjects` for that, cause when we open a popup with `showPopup` we just pass the initial props to this function, and then we cannot change them or pass new ones without closing and reopening the popup, which would cause the loss of internal component's data.

A custom component would be an unjustified copypasting of all three components modified in this pr

haiodo: I think it is a bad idea to show some dialog over a popup dialog, since it is completely unexpected by user. It more looks like inplace user list with selection, and in this case you could on click show something. So please just create a new component for your needs.

denis-bunakalya: Done

denis-bunakalya:
Not more actual

PR #2820, opened 2023-03-23
Export UserInfo

wazsone:
Related to https://github.com/Ezthera/ezqms/pull/306


sol-0:
The same change is already present in this PR: https://github.com/hcengineering/anticrm/pull/2819

PR #2824, opened 2023-03-24
Allow to disable particular items in ObjectPopup

denis-bunakalya:
# Contribution checklist

## Checklist

* [x] - Does it well tested?
* [x] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

### DIFF
--- a/packages/presentation/src/components/ObjectPopup.svelte
+++ b/packages/presentation/src/components/ObjectPopup.svelte
@@ -58,7 +58,7 @@
 
   export let create: ObjectCreate | undefined = undefined
   export let readonly = false
-  export let readonlyItemIds: Set<Ref<Doc>> = new Set()
+  export let forbiddenDeselectItemIds: Set<Ref<Doc>> = new Set()
 
   let search: string = ''
   let objects: Doc[] = []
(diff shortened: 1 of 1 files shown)


haiodo:
@denis-bunakalya I do not like this idea at all.

denis-bunakalya:
> @denis-bunakalya I do not like this idea at all.

Any reason? We need it to show all the selected items, but disable editing for some of them. If we would show only editable items, user sometimes won't see other previously selected items, which are not editable for him

haiodo:
So you need to disallow deselection of some of items? 
In this case you need name it differently, since it is confusing.

And potentially it should be shown differently in control.

denis-bunakalya:
> So you need to disallow deselection of some of items? In this case you need name it differently, since it is confusing.
> 
> And potentially it should be shown differently in control.

Done

packages/presentation/src/components/ObjectPopup.svelte
```diff
@@ -58,6 +58,7 @@
 
   export let create: ObjectCreate | undefined = undefined
   export let readonly = false
+  export let forbiddenDeselectItemIds: Set<Ref<Doc>> = new Set()
```
haiodo:
```suggestion
  export let disallowDeselect: Ref<Doc>[]| undefined
```

denis-bunakalya: Done

haiodo (approved):

PR #2944, opened 2023-04-11
TSK-1079 Баг Отображение Неактивных пустых групп в разделе Активных задач

BykhovDenis:
# Contribution checklist

## Checklist

* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues
https://front.hc.engineering/workbench/platform/tracker/TSK-1079

### DIFF
--- a/plugins/view-resources/src/components/list/ListCategories.svelte
+++ b/plugins/view-resources/src/components/list/ListCategories.svelte
@@ -93,7 +93,7 @@
           const f = await getResource(categoryFunc.action)
           const res = hierarchy.clone(await f(_class, query, groupByKey, update, queryId, $statusStore))
           if (res !== undefined) {
-            categories = concatCategories(categories, res)
+            categories = concatCategories(res, categories)
             return
           }
         }
--- a/plugins/view-resources/src/components/list/ListCategories.svelte
+++ b/plugins/view-resources/src/components/list/ListCategories.svelte
@@ -19,7 +19,15 @@
   import { AnyComponent } from '@hcengineering/ui'
   import { AttributeModel, BuildModelKey, CategoryOption, ViewOptionModel, ViewOptions } from '@hcengineering/view'
   import { createEventDispatcher, onDestroy } from 'svelte'
-  import { buildModel, getAdditionalHeader, getCategories, getGroupByValues, getPresenter, groupBy } from '../../utils'
+  import {
+    buildModel,
+    concatCategories,
+    getAdditionalHeader,
+    getCategories,
+    getGroupByValues,
+    getPresenter,
+    groupBy
+  } from '../../utils'
   import { CategoryQuery, noCategory } from '../../viewOptions'
   import ListCategory from './ListCategory.svelte'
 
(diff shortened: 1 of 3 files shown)


plugins/view-resources/src/components/list/ListCategories.svelte
```diff
@@ -85,7 +85,11 @@
           const f = await getResource(categoryFunc.action)
           const res = hierarchy.clone(await f(_class, query, groupByKey, update, queryId, $statusStore))
           if (res !== undefined) {
-            categories = Array.from(new Set([...categories, ...res]))
+            const set = new Set(categories.map((p) => JSON.stringify(p)))
```
haiodo: Why we need this? They already should be uniq.

plugins/view-resources/src/viewOptions.ts
```diff
@@ -131,16 +131,18 @@ export async function showEmptyGroups (
...
       return it.ofAttribute === attr._id
     })
     if (query !== undefined) {
-      statusList = matchQuery<Status>(
-        statusList,
-        query as DocumentQuery<Status>,
-        _class,
-        hierarchy
-      ) as unknown as Array<WithLookup<Status>>
+      const { [key]: st, space } = query
```
haiodo, line 138:
```suggestion
      const { [key]: st, space, ...data } = query
```

plugins/view-resources/src/viewOptions.ts
```diff
@@ -131,16 +131,18 @@ export async function showEmptyGroups (
...
+      const resQuery: DocumentQuery<Status> = {
+        space
```
haiodo, line 140:
```suggestion
        ...data,
        space
```

plugins/view-resources/src/components/list/ListCategories.svelte
BykhovDenis: This is necessary when the function returns a filtered value, but we have an object with a different category field value. It worked fine until we only had the number | string as category, now we have objects (status) and Set can't make it unique.

BykhovDenis: Probably better use something like fastEqual, but I'm not sure

plugins/view-resources/src/viewOptions.ts
BykhovDenis: We don't need it for status filter

plugins/view-resources/src/viewOptions.ts
BykhovDenis: We don't need it for status filter

plugins/view-resources/src/components/list/ListCategories.svelte
haiodo: categories = await getCategories(client, _class, docs, groupByKey, $statusStore) should always return uniq values already.

plugins/view-resources/src/utils.ts
```diff
@@ -648,6 +648,33 @@ export async function getCategories (
...
+      const id = item.name
+      if (!uniqueObjects.has(id)) {
+        uniqueObjects.set(id, item)
+      }
+    } else {
+      uniqueValues.add(item)
+    }
+  }
+
+  return [...uniqueValues, ...uniqueObjects.values()]
```
haiodo, line 675: Order will be lost if we mix StatusValue and other values.

haiodo (approved):

PR #3043, opened 2023-04-21
[TSK-1304] Fix active viewlet if viewlet list is empty

s0gorelkov:
# Contribution checklist
files

## Checklist

* [x] - Does it well tested?
* [x] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

[TSK-1304](https://front.hc.engineering/workbench/platform/tracker/TSK-1304)


plugins/tracker-resources/src/components/sprints/SprintBrowser.svelte
```diff
@@ -84,7 +84,10 @@
 
   async function update (viewlets: WithLookup<Viewlet>[], active: Ref<Viewlet> | null): Promise<void> {
     viewlet = viewlets.find((viewlet) => viewlet._id === active) ?? viewlets[0]
-    setActiveViewletId(viewlet._id)
+
+    if (viewlet) {
+      setActiveViewletId(viewlet._id)
```
haiodo: @BykhovDenis Should it be set to undefined in this case?

BykhovDenis: I think we always should have least one viewlet

s0gorelkov: `viewlets` are `[]` by default. So, we don't have viewlets until the viewlets request is completed.

haiodo: Should it be better to check for viewlets length before calling update in this case?

s0gorelkov: Anyway, I think it's better to move this func to some util to avoid duplication throughout the project.

haiodo (approved):

PR #3101, opened 2023-04-27
[TSK-779] Use txMatch

wazsone:
# Contribution checklist

## Related issues
https://front.hc.engineering/workbench/platform/tracker/TSK-779

## Checklist

* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

### DIFF
--- a/models/server-attachment/src/index.ts
+++ b/models/server-attachment/src/index.ts
@@ -16,6 +16,7 @@
 
 import { Builder } from '@hcengineering/model'
 
+import attachment from '@hcengineering/attachment'
 import core from '@hcengineering/core'
 import serverAttachment from '@hcengineering/server-attachment'
 import serverCore from '@hcengineering/server-core'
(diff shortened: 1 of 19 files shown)


models/server-attachment/src/index.ts
```diff
@@ -24,6 +25,11 @@ export { serverAttachmentId } from '@hcengineering/server-attachment'
 
 export function createModel (builder: Builder): void {
   builder.createDoc(serverCore.class.Trigger, core.space.Model, {
-    trigger: serverAttachment.trigger.OnAttachmentDelete
+    trigger: serverAttachment.trigger.OnAttachmentDelete,
+    txMatch: {
+      _class: core.class.TxCollectionCUD,
+      'tx.objectClass': attachment.class.Attachment,
```
BykhovDenis, line 31: We also have attachment.class.Photo and document.class.CollaboratorDocument which derived from Attachment. We need think about good solution for it. @haiodo do you have any idea? Probably we should have another field to check object class.

BykhovDenis (changes requested):

models/server-attachment/src/index.ts
haiodo: Best ideas will be to make platform for every Ref<Class<Doc>> field treat them as class and all derived from it in case of matching. We do it manually for findAll for _class field, so it we do so for every reference to Class, it should simplify places like this. But we will need type for tx, and we do not have it right now, so right now in triggers we need to update query and if we have filter for tx.objectClass, we should extend it to any derived from class properly.

@wazsone Please modify query inside server/core/src/triggers.ts, before passing it to matchQuery for fields of tx.objectClass. Be aware of $in/$nin/$ne

wazsone: @BykhovDenis @haiodo Done
https://github.com/hcengineering/anticrm/pull/3101/files#diff-17bd82836258e5117f7b0823518ecf8a2887e58868f5d014760c479cd76f8fcc

BykhovDenis (approved):

PR #3174, opened 2023-05-12
[TSK-1370] Show selected filter values on top 

s0gorelkov:
# Contribution checklist

## Checklist

* [x] - Does it well tested?
* [x] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

[TSK-1370](https://front.hc.engineering/workbench/platform/tracker/TSK-1370)


plugins/contact-resources/src/components/ChannelFilter.svelte
```diff
@@ -46,6 +44,11 @@
     } else {
       selected = [...selected, element._id]
     }
+
+    filter.value = [...selected]
```
haiodo: Why we need to update filter here? It is not expected by method contract. We should rename it, also I propose to set a timer with like 100ms delay before we will start apply, with cancel of previous timer, so if user will click multiple times, we will not start filtering immediately.

plugins/view-resources/src/filter.ts
```diff
@@ -294,3 +294,18 @@ export function getFilterKey (_class: Ref<Class<Doc>>): string {
...
+ * Returns a new array where the selected values have been moved to the beginning
+ *
+ * @export
+ * @template T
+ * @param {T[]} values
+ * @param {(value: T) => boolean} checkIsSelected
+ * @returns {readonly T[]}
+ */
+export function sortFilterValues<T> (values: T[], checkIsSelected: (value: T) => boolean): readonly T[] {
+  const checkIsNotSelected = (value: T): boolean => !checkIsSelected(value)
```
haiodo: Let's have two arrays and one for with one call to checkIsSelected instead of two.

plugins/contact-resources/src/components/ChannelFilter.svelte
s0gorelkov: Done.

plugins/view-resources/src/filter.ts
s0gorelkov: Done.

plugins/view-resources/src/utils.ts
```diff
@@ -63,6 +63,7 @@ import { get, writable } from 'svelte/store'
 import plugin from './plugin'
 import { noCategory } from './viewOptions'
 
+export { default as debounce } from 'lodash.debounce'
```
haiodo: I'm not happy with this.

s0gorelkov: I'm not sure about writing custom implementation of exactly the same functionality. We also have this dependency in the "Collaborator" pod.

haiodo: But it is not used.

Let's switch to just a 
```js
clearTimeout(timeout); timeout = setTimeout(refresh, 200)
```

and it will be enough for now.

s0gorelkov: Done.

plugins/view-resources/package.json
```diff
@@ -45,7 +46,8 @@
     "@hcengineering/notification": "^0.6.12",
     "@hcengineering/presentation": "^0.6.2",
     "@hcengineering/setting": "^0.6.7",
+    "@hcengineering/text-editor": "^0.6.0",
     "fast-equals": "^2.0.3",
-    "@hcengineering/text-editor": "^0.6.0"
+    "lodash.debounce": "~4.0.8"
```
haiodo: do we still need it?

haiodo (approved):

PR #3186, opened 2023-05-15
UBER-135/TSK-1430: allow changing image in PDFViewer through arrow-keys (keyboard)

ThetaDR:
# Contribution checklist

## Checklist

* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues
https://front.hc.engineering/workbench/platform/tracker/TSK-1430


packages/ui/src/popups.ts
```diff
@@ -32,6 +32,16 @@ export interface PopupResult {
 
 export const popupstore = writable<CompAndProps[]>([])
 
+export function updatePopup (id: string, props: any): void {
```
haiodo:
```suggestion
export function updatePopup (id: string, props: CompAndProps): void {
```
?

plugins/attachment-resources/src/components/AttachmentPreview.svelte
```diff
@@ -34,11 +38,13 @@
...
     on:click={() => {
       closeTooltip()
-      showPopup(
+      if (listProvider !== undefined) listProvider.updateFocus(value)
+      const popupInfo = showPopup(
         PDFViewer,
         { file: value.file, name: value.name, contentType: value.type },
         value.type.startsWith('image/') ? 'centered' : 'float'
       )
+      dispatch('popupOpened', popupInfo.id)
```
haiodo:
```suggestion
      dispatch('open', popupInfo.id)
```

plugins/attachment-resources/src/components/AttachmentStyledBox.svelte
```diff
@@ -113,6 +126,7 @@
         }
       )
     }
+    console.log(attachments)
```
haiodo:
```suggestion
```

packages/ui/src/popups.ts
ThetaDR: Then we would have to get existing CompAndProps for the popup and pass all of them as an argument. This way we can just use popupId and new props for popup's component

BykhovDenis: You can use Partial

ThetaDR: Made it partial

haiodo (approved):

PR #3228, opened 2023-05-21
[UBER-178] New way of All/Active/Backlog should be in URI

wazsone:
# Contribution checklist

## Brief description
We need to respect selection of item of Active etc in URI, so filters will be properly handled.

## Related issues
https://front.hc.engineering/workbench/platform/tracker/UBER-178

## Checklist

* [x] - Does it well tested?
* [x] - Tested for Chrome.
* [x] - Tested for Safari.

### DIFF
--- a/plugins/tracker-resources/src/components/issues/Issues.svelte
+++ b/plugins/tracker-resources/src/components/issues/Issues.svelte
@@ -13,11 +13,12 @@
 // limitations under the License.
 -->
 <script lang="ts">
-  import { createEventDispatcher, onMount } from 'svelte'
+  import { createEventDispatcher } from 'svelte'
   import { DocumentQuery, Ref } from '@hcengineering/core'
   import { Issue, Project } from '@hcengineering/tracker'
   import { IntlString } from '@hcengineering/platform'
   import { createQuery } from '@hcengineering/presentation'
+  import { resolvedLocationStore } from '@hcengineering/ui'
 
   import { IModeSelector } from '@hcengineering/ui'
   import tracker from '../../plugin'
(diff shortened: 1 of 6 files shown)


plugins/tracker-resources/src/components/issues/Issues.svelte
```diff
@@ -30,6 +31,8 @@
     ['active', tracker.string.Active, {}],
     ['backlog', tracker.string.Backlog, {}]
   ]
+  let [[mode]] = config
+  const loc = parseLocation(new URL(`${window.location.href}/${mode}`))
```
haiodo: Should we use a #fragment for mode? @BykhovDenis

haiodo: Also I think we need to pass extra path from plugins/workbench-resources/src/components/Workbench.svelte to individual components and pass back options to put into path, instead of adding into each individual component

wazsone: @haiodo Thx, done

plugins/workbench-resources/src/components/Workbench.svelte
```diff
@@ -690,6 +676,13 @@
           <Component
             is={specialComponent.component}
             props={{ model: navigatorModel, ...specialComponent.componentProps, currentSpace, visibileNav }}
+            on:action={(e) => {
+              if (e?.detail) {
+                const loc = getCurrentLocation()
+                loc.query = { ...loc.query, mode: e.detail }
```
haiodo: I suppose we need more generic query update here, not only mode.

wazsone: @haiodo thank you, done

haiodo (approved):

BykhovDenis (approved):

PR #3536, opened 2023-07-31
UBER-619: StatusPopup for creating/renaming

ThetaDR:
# Contribution checklist

## Checklist

* [ ] - Does it well tested?
* [x] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues
https://front.hc.engineering/workbench/platform/tracker/UBER-619


plugins/task-resources/src/components/CreateStatePopup.svelte
```diff
@@ -0,0 +1,45 @@
...
+  import task from '../plugin'
+  import presentation, { Card } from '@hcengineering/presentation'
+
+  const dispatch = createEventDispatcher()
+  export let existingNames: string[] = []
+  export let value = ''
+
+  $: canSave = !existingNames.includes(value)
+  function save () {
+    dispatch('close', value)
```
haiodo: You need to perform create/update operation here with apply and check if there is no such names,

ThetaDR: Why though? It leads to more props to provide (class in case of new status, value in case of the existing one)

plugins/task-resources/src/components/CreateStatePopup.svelte
```diff
@@ -0,0 +1,87 @@
...
+      if (!hierarchy.isDerived(_class, task.class.DoneState)) {
+        newDoc = {
+          ofAttribute: task.attribute.State,
+          name: value.trim(),
+          color: 9,
+          rank: calcRank(lastOne, undefined),
+          ...attachedTo
+        }
+      }
+      const ops = client.apply('create-update-state').notMatch(_class, {
```
haiodo: Better to use space, so few operations for different spaces will go in parallel

server/core/src/storage.ts
```diff
@@ -196,7 +196,7 @@ class TServerStorage implements ServerStorage {
       return result[0]
     }
     if (result.length === 0) {
-      return [{}, false]
```
haiodo, line 199: Why we need this change here?

server/mongo/src/storage.ts
```diff
@@ -737,7 +737,7 @@ class MongoAdapter extends MongoAdapterBase {
       return (await this.getOperations(txes[0])?.raw()) ?? {}
     }
     if (result.length === 0) {
-      return {}
+      return false
```
haiodo, line 740: Why we need this change here?

server/core/src/storage.ts
ThetaDR: For ApplyIf. It's expected to return boolean, but in case with non-complete request it returns array, which we check as boolean and it's always true (cause it's an array) even though should've been false

server/mongo/src/storage.ts
haiodo: ok

server/core/src/storage.ts
haiodo: ok

haiodo (approved):

plugins/task-assets/lang/ru.json
```diff
@@ -73,6 +73,9 @@
     "TodoItems": "Todos",
     "Dashboard": "Дашборд",
     "AllTime": "Все время",
-    "RelatedIssues": "Связанные процессы"
+    "RelatedIssues": "Связанные процессы",
+    "StatusName": "Имя статуса",
+    "StatusPopupTitle": "Создание статуса и изменение имени существующего",
+    "NameAlreadyExists": "Данное имя уже исползутеся другим статусом этого типа"
```
aonnikov, line 79: typo: `исползутеся` -> `используется`

PR #3618, opened 2023-08-22
UBER-772: require having employee mixin to allow Staff mixin

ThetaDR:
# Contribution checklist

## Checklist

* [ ] - Does it well tested?
* [x] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues
https://front.hc.engineering/workbench/platform/tracker/UBER-772

### DIFF
--- a/plugins/view-resources/package.json
+++ b/plugins/view-resources/package.json
@@ -47,7 +47,6 @@
     "@hcengineering/setting": "^0.6.9",
     "@hcengineering/text-editor": "^0.6.0",
     "@hcengineering/query": "^0.6.7",
-    "@hcengineering/hr": "^0.6.10",
     "fast-equals": "^2.0.3"
   }
 }
--- a/plugins/view-resources/package.json
+++ b/plugins/view-resources/package.json
@@ -47,6 +47,7 @@
     "@hcengineering/setting": "^0.6.9",
     "@hcengineering/text-editor": "^0.6.0",
     "@hcengineering/query": "^0.6.7",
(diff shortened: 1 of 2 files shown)


plugins/view-resources/package.json
```diff
@@ -47,6 +47,7 @@
     "@hcengineering/setting": "^0.6.9",
     "@hcengineering/text-editor": "^0.6.0",
     "@hcengineering/query": "^0.6.7",
+    "@hcengineering/hr": "^0.6.10",
```
haiodo: Nope, not allowed.

plugins/view-resources/src/components/EditDoc.svelte
```diff
@@ -104,7 +105,10 @@
         m.kind === ClassifierKind.MIXIN &&
         !ignoreMixins.has(m._id) &&
         (hierarchy.hasMixin(object, m._id) ||
-          (showAllMixins && hierarchy.isDerived(realObjectClass, hierarchy.getBaseClass(m._id))))
+          (showAllMixins &&
```
haiodo, line 107: Base class in our case is Person, not Employee, so we need to filter all mixins added to class extends our base class.

haiodo: Ouch, Employee is mixin for now, we need to check for child mixins, if document is already had it, and only in this case we should allow child mixins.

ThetaDR: Should be fixed now

haiodo (approved):

PR #3630, opened 2023-08-25
[UBERF-41] Chat notification fixes

sol-0:
# Contribution checklist

## Brief description
https://front.hc.engineering/workbench/platform/tracker/UBERF-41

## Checklist

* [ ] - Does it well tested?
* [x] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

### DIFF
--- a/plugins/chunter-resources/src/components/Channel.svelte
+++ b/plugins/chunter-resources/src/components/Channel.svelte
@@ -105,11 +105,11 @@
   function newMessagesStart (messages: Message[], docUpdates: Map<Ref<Doc>, DocUpdates>): number {
     if (space === undefined) return -1
     const docUpdate = docUpdates.get(space)
-    const lastView = docUpdate?.txes?.[0]?.modifiedOn
+    const lastView = docUpdate?.txes?.findLast((tx) => !tx.isNew)
     if (docUpdate === undefined || lastView === undefined) return -1
     for (let index = 0; index < messages.length; index++) {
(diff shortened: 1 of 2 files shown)


sol-0:
@BykhovDenis While searching for the root cause for this issue, I noticed that the logic in couple of places assumed that `NotificationClientImpl.docUpdatesStore` stores only new updates.
I've adjusted the logic to properly show chat notifications, but haven't changed the store logic itself. So, the questions is: shall I change the logic of populating of the store instead? (store only updated spaces and only new txes for those spaces)
In my opinion, having all the updates in the store is more flexible, but we might need to adjust the code in other places also in case there are any similar issues with notifications.

BykhovDenis:
> @BykhovDenis While searching for the root cause for this issue, I noticed that the logic in couple of places assumed that `NotificationClientImpl.docUpdatesStore` stores only new updates.
> I've adjusted the logic to properly show chat notifications, but haven't changed the store logic itself. So, the questions is: shall I change the logic of populating of the store instead? (store only updated spaces and only new txes for those spaces)
> In my opinion, having all the updates in the store is more flexible, but we might need to adjust the code in other places also in case there are any similar issues with notifications.

It was old implementation, we added all ties instead only new for have easy way to show `people` mode, because it show not only new changes. I think we shouldn't currently change store logic.

BykhovDenis (approved):

haiodo (approved):

PR #3643, opened 2023-08-30
UBER-810 Split mixin update to doc and mixin

aonnikov:
# Contribution checklist

## Checklist

* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

### DIFF
--- a/packages/core/src/operations.ts
+++ b/packages/core/src/operations.ts
@@ -17,7 +17,6 @@ import { Client } from './client'
 import core from './component'
 import type { DocumentQuery, FindOptions, FindResult, TxResult, WithLookup } from './storage'
 import { DocumentClassQuery, Tx, TxCUD, TxFactory, TxProcessor } from './tx'
-import { splitObjectAttributes } from './utils'
 
 /**
  * @public
@@ -220,7 +219,7 @@ export class TxOperations implements Omit<Client, 'notify'> {
     return this.client.tx(tx)
   }
 
-  update<T extends Doc>(
+  async update<T extends Doc>(
     doc: T,
     update: DocumentUpdate<T>,
     retrieve?: boolean,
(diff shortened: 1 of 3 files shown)


packages/core/src/operations.ts
```diff
@@ -229,9 +230,18 @@ export class TxOperations implements Omit<Client, 'notify'> {
     const hierarchy = this.client.getHierarchy()
     const mixClass = Hierarchy.mixinOrClass(doc)
     if (hierarchy.isMixin(mixClass)) {
-      // TODO: Rework it is wrong, we need to split values to mixin update and original document update if mixed.
+      const { docUpdate, mixinUpdate } = this.splitMixinUpdate(update, mixClass)
```
haiodo: Let's use apply to pass them all to server and execute instead of multiple requests.

packages/core/src/operations.ts
```diff
@@ -229,9 +230,18 @@ export class TxOperations implements Omit<Client, 'notify'> {
...
+
       const baseClass = hierarchy.getBaseClass(doc._class)
-      return this.updateMixin(doc._id, baseClass, doc.space, mixClass, update, modifiedOn, modifiedBy)
+
+      const result: Promise<TxResult>[] = []
+      if (Object.keys(docUpdate).length > 0) {
+        result.push(this.updateDoc(baseClass, doc.space, doc._id, docUpdate, retrieve, modifiedOn, modifiedBy))
+      }
+      if (Object.keys(mixinUpdate).length > 0) {
```
haiodo: also please expect, it could be multiple mixins in one update.

Example:

Person -> Employee(mixin) -> MyEmployee(mixin)

So if we update MyEmployee with fields from it and Employee and Person it should be 3 calls

packages/core/src/operations.ts
```diff
@@ -322,6 +332,40 @@ export class TxOperations implements Omit<Client, 'notify'> {
...
+    update: DocumentUpdate<T>,
+    mixClass: Ref<Class<T>>
+  ): {
+      docUpdate: DocumentUpdate<T>
+      mixinUpdate: DocumentUpdate<T>
+    } {
+    const docUpdate: DocumentUpdate<T> = {}
+    const mixinUpdate: DocumentUpdate<T> = {}
+
+    const mixinAttrs = this.getHierarchy().getOwnAttributes(mixClass)
```
haiodo: Please use getAllAttributes and split based on attribute of value.

Something like in sync.ts of bitrix plugin.

```ts
     const attrs = getAllAttributes(client, recruit.mixin.Candidate)

        // Update candidate operations
        for (const u of mapping?.updateCandidate ?? []) {
          const attribute = attrs.get(u.attr)
          if (attribute === undefined) {
            console.error('failed to fill attribute', u.attr)
            continue
          }
          if (client.getHierarchy().isMixin(attribute.attributeOf)) {
            doc.mixins[attribute.attributeOf] = { ...(doc.mixins[attribute.attributeOf] ?? {}), [u.attr]: u.value }
          } else {
            ;(doc.document as any)[u.attr] = u.value
          }
        }
```

packages/core/src/operations.ts
aonnikov: Fixed

packages/core/src/operations.ts
aonnikov: Fixed in more generic way

packages/core/src/operations.ts
aonnikov: thanks, fixed

packages/core/src/operations.ts
```diff
@@ -229,9 +229,18 @@ export class TxOperations implements Omit<Client, 'notify'> {
...
     if (hierarchy.isMixin(mixClass)) {
-      // TODO: Rework it is wrong, we need to split values to mixin update and original document update if mixed.
       const baseClass = hierarchy.getBaseClass(doc._class)
-      return this.updateMixin(doc._id, baseClass, doc.space, mixClass, update, modifiedOn, modifiedBy)
+
+      const byClass = this.splitMixinUpdate(update, mixClass, baseClass)
+      const ops = this.apply(doc._id)
+      for (const it of byClass) {
+        if (hierarchy.isMixin(it[0])) {
+          void ops.updateMixin(doc._id, baseClass, doc.space, it[0], it[1], modifiedOn, modifiedBy)
```
haiodo: It looks better with await

packages/core/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "./node_modules/@hcengineering/platform-rig/profiles/default/tsconfig.json",
 
   "compilerOptions": {
+    "target": "ES2019",
```
haiodo: May be we should change it in platform-rig/profiles/default/tsconfig.json?

haiodo (approved):

packages/core/tsconfig.json
aonnikov: changed

packages/core/src/operations.ts
aonnikov: replaced with await

haiodo (approved):

PR #3655, opened 2023-09-04
UBER-564: Add sound notification and settings

mixerka:
# Contribution checklist

## Brief description

Show notification preferences if there are such

## Checklist

* [x] - Does it well tested?
* [x] - Tested for Chrome.
* [x] - Tested for Safari.

## Related issues

### DIFF
--- a/plugins/notification-resources/src/components/NotificationSettings.svelte
+++ b/plugins/notification-resources/src/components/NotificationSettings.svelte
@@ -14,7 +14,8 @@
 -->
 <script lang="ts">
   import { Ref } from '@hcengineering/core'
-  import type { NotificationGroup, NotificationSetting, NotificationType } from '@hcengineering/notification'
+  import { getResource } from '@hcengineering/platform'
+  import type { NotificationGroup, NotificationPreferencesGroup, NotificationSetting, NotificationType } from '@hcengineering/notification'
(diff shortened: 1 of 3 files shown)


plugins/notification-resources/src/components/NotificationSettings.svelte
```diff
@@ -44,6 +45,19 @@
   })
 
   let group: Ref<NotificationGroup> | undefined = undefined
+  let currentPreferenceGroup: NotificationPreferencesGroup | undefined = undefined
+  
+  client.findAll(notification.class.NotificationPreferencesGroup, {}).then((res) => {
```
haiodo, line 55: Please use query instead, since you could have multiple tabs/browsers, so you should receive notifications.

mixerka: I considered this. My intuition was like that: those items are pretty static, they are for preferences items as such:

Screenshot 2023-09-04 at 09 56 50

And they should be pretty static, like we are creating them on workspace setup. Do we really want to use query for such static content too?

haiodo: ok, is for groups.

haiodo (approved):

PR #3667, opened 2023-09-07
Fix ezqms upgrade

annano:
# Contribution checklist

## Brief description
Api was removed in this PR: https://github.com/hcengineering/anticrm/pull/3626

## Checklist

* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

### DIFF
--- a/packages/text-editor/src/components/extension/nodeHighlight.ts
+++ b/packages/text-editor/src/components/extension/nodeHighlight.ts
@@ -1,4 +1,4 @@
-import { Extension, getMarkRange, mergeAttributes } from '@tiptap/core'
+import { Extension, getMarkRange, mergeAttributes, Range } from '@tiptap/core'
 import { Plugin, TextSelection } from 'prosemirror-state'
 import { NodeUuidExtension, NodeUuidOptions } from './nodeUuid'
 
@@ -10,6 +10,11 @@ export enum NodeHighlightType {
 interface NodeHighlightExtensionOptions extends NodeUuidOptions {
   getNodeHighlightType: (uuid: string) => NodeHighlightType | undefined | null
   isHighlightModeOn: () => boolean
+}
+
+// eslint-disable-next-line @typescript-eslint/no-invalid-void-type
+function isRange (range: Range | undefined | null | void): range is Range {
+  return range !== null && range !== undefined
 }
 
 /**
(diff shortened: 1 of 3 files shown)


haiodo:
@annano nope, you need to move this functionality to server.

annano:
> @annano nope, you need to move this functionality to server.

Removed, can you please review again?

packages/text-editor/src/components/extension/nodeHighlight.ts
```diff
@@ -29,7 +34,7 @@ export const NodeHighlightExtension = Extension.create<NodeHighlightExtensionOpt
 
             const range = getMarkRange(doc.resolve(pos), schema.marks[NodeUuidExtension.name])
 
-            if (range === null || range === undefined) {
+            if (!isRange(range)) {
```
haiodo: May be?
```suggestion
            if (range == null) {
```
will be enough?

packages/text-editor/src/components/extension/nodeHighlight.ts
```diff
@@ -12,6 +12,11 @@ interface NodeHighlightExtensionOptions extends NodeUuidOptions {
   isHighlightModeOn: () => boolean
 }
 
+// eslint-disable-next-line @typescript-eslint/no-invalid-void-type
+function isRange (range: Range | undefined | null | void): range is Range {
```
haiodo: why not boolean response?

annano: I need this type guard because `getMarkRange` return type is `Range | void` and then I get the typescript error later that `Range | void` type does not have properties `to` and `from`.

This type guard just to tell typescript it is `Range` type (not void), as I can not add `&& range !== void`

haiodo: But return type is boolean?

haiodo: May be you have unintended upgrade of typescript in QMS?

lexiv0re: It's this concept https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards and it's been in TS for a while. It helps you to narrow the object type down. Very convenient and makes the code clearer for cases where you need to conditionally cast object to some type.

haiodo: Some very weird concept actually, I do not think we need it here.

annano: @haiodo Removed, can you please review?

BykhovDenis (approved):

PR #3706, opened 2023-09-15
Demo autotest structure

jakson-b1ank:
First of all, I add Page Object Model and made small refactoring. I described a number of elements (pages, windows, tabs and sidebars) as well as actions with them. I also add fixture, but it needs to be improved(multi-authorization and remaining structural elements of the pages)

## Added:
- Common POM fixture
- Recruit page, windows, table, left sidebar, locators and methods
- In file "actions.spec.ts" refactored first test just for example, how I see it
- Moved some files for correct fixture

## TODO:
- Add i18n (for multi-language text-locators)
- Add other pages and elements to POM fixture
- Add in fixture multi-user (with multi-thread) authorization 
- Сhange the approach to testing - create and delete data before and after the test, without using scripts, only with playwright tools and API, without state saving


tests/sanity/tsconfig.json
```diff
@@ -1,9 +0,0 @@
-{
```
haiodo, line 1: Why it is removed? It is required for proper function of project.

aplatoff:
@trix7xter please make all checks passing, especially `uitest` :)

tests/sanity/tests/actions.spec.ts
```diff
@@ -10,19 +10,39 @@ test.describe('actions tests', () => {
...
 
-    await page.click('text=Talents')
-    await expect(page).toHaveURL(`${PlatformURI}/workbench/sanity-ws/recruit/talents`)
+  test('Add new candidate', async ({ page, authorizedMainPage, recruitPage }) => {
+    await test.step(`Open recruit application`, async () => {
+      await authorizedMainPage.leftSidebar.openRecruitApplication();
+    });
 
-    await page.click('td:has-text("Frontend Engineer")')
+    await test.step(`Open talents tab`, async () => {
```
aonnikov, line 19: Wrapping every statement in a separate step seem to be excessive to me. It makes the code less readable to my taste, I would prefer larger steps.

tests/sanity/tests/actions.spec.ts
```diff
@@ -10,19 +10,39 @@ test.describe('actions tests', () => {
...
+      await recruitPage.recruitsPanel.openTalentsTab();
+    });
 
-    await page.press('body', 'Meta+k')
-    await page.fill('[placeholder="type\\ to\\ filter\\.\\.\\."]', 'Talent')
-    expect(await page.locator('div.selectPopup :text("New Talent")').count()).toBe(1)
-    await page.click('div.selectPopup :text("New Talent")')
-    await page.click('button#card-close')
+    await test.step(`Check talents url`, async () => {
+      await expect(page).toHaveURL(`${PlatformURI}/workbench/sanity-ws/recruit/talents`)
```
aonnikov, line 24: nit: I know that it was in the original test, but I don't see a purpose of checking URL in the test that verifies a candidate creation.

I see the following option:
1. Remove this check
2. If URLs are important for some reason, we can add a separate test specific to navigation
3. Or move this check into `openTalentsTab` so we check URL each time we navigated.

tests/sanity/tsconfig.json
jakson-b1ank: OK, but this did not affect the operation of autotests. In general, it is enough to specify other directories (to which I moved the root directory and out dir).

tests/sanity/tests/actions.spec.ts
jakson-b1ank: I saved it because it was in original test. These kind of checks needs to be in other Assert classes, and if I were you I would use these checks. Maybe later I'll send example.

PR #3729, opened 2023-09-21
UBER-845: Add NotificationPresenter to send rich text notifications

mixerka:
# Contribution checklist

## Checklist

* [x] - Does it well tested?
* [x] - Tested for Chrome.
* [x] - Tested for Safari.

## Related issues


packages/text/src/plain-text.ts
```diff
@@ -0,0 +1,98 @@
...
+  'video'
+]
+const ELLIPSIS_CHAR = '…'
+const WHITESPACE = ' '
+
+/**
+ * @public
+ */
+export function stripTags (htmlString: string, textLimit = 0): string {
+  const parser = new SaxesParser()
```
haiodo: Why we need this? we have already facilities to parse html using tiptap and prosemirror

plugins/notification/src/index.ts
```diff
@@ -169,6 +179,10 @@ export interface DocUpdateTx {
   modifiedBy: Ref<Account>
   modifiedOn: Timestamp
   isNew: boolean
+  title?: IntlString
```
haiodo, line 184: Why they are optional? And we discussed rename to InboxItem

plugins/notification/src/index.ts
```diff
@@ -95,6 +95,16 @@ export interface NotificationTemplate {
   subjectTemplate: string
 }
 
+/**
+ * @public
+ */
+export interface NotificationPresentation {
+  title: IntlString
+  body: IntlString
+  intlParams: Record<string, string | number>
```
haiodo, line 104: should it be titleParams and bodyParams? Semantic not clear right now

plugins/notification/src/index.ts
```diff
@@ -95,6 +95,16 @@ export interface NotificationTemplate {
...
+  intlParamsNotLocalized?: Record<string, IntlString>
```
haiodo, line 105: Should sender be a part of it?

haiodo: Or Icon if it is a service send.

packages/text/src/plain-text.ts
haiodo: Better to do parse with tiptap and render to text using it.

server-plugins/notification-resources/src/index.ts
```diff
@@ -442,18 +448,90 @@ async function isShouldNotify (
   }
 }
 
+async function findPersonForAccount (control: TriggerControl, personId: Ref<Person>): Promise<Person | null> {
```
haiodo: Please prefer undefined.

server-plugins/notification/src/index.ts
```diff
@@ -112,14 +112,32 @@ export interface TypeMatch extends NotificationType {
   func: TypeMatchFunc
 }
 
+/**
+ * @public
+ */
+export type IntlFullfilmentFunction = (
```
haiodo: Let's think about better name

server-plugins/tracker-resources/src/index.ts
```diff
@@ -69,6 +75,83 @@ export async function issueTextPresenter (doc: Doc, control: TriggerControl): Pr
   return issueName
 }
 
+function isTheSamePerson (control: TriggerControl, assignee: Ref<Person>, target: Ref<Account>): boolean {
```
haiodo:
```suggestion
function isSamePerson (control: TriggerControl, assignee: Ref<Person>, target: Ref<Account>): boolean {
```

server-plugins/tracker-resources/src/index.ts
```diff
@@ -69,6 +75,83 @@ export async function issueTextPresenter (doc: Doc, control: TriggerControl): Pr
...
+  const targetAccount = control.modelDb.getObject(target) as PersonAccount
+  return assignee === targetAccount?.person
+}
+
+const NOTIFICATION_BODY_SIZE = 50
```
haiodo, line 83: I do no think we need to strip it here, inbox presenter should be smart enough to render it properly no matter of size.

packages/text/src/plain-text.ts
aonnikov: You can find them here:
https://github.com/hcengineering/anticrm/blob/main/packages/text/src/html.ts

server-plugins/chunter-resources/src/index.ts
```diff
@@ -536,6 +537,40 @@ export async function IsThreadMessage (
   return space !== undefined
 }
 
+const NOTIFICATION_BODY_SIZE = 50
+/**
+ * @public
+ */
+export async function getChunterFullfilmentPrarams (
```
aonnikov: typo: `Prarams` -> `Params`

plugins/notification/src/index.ts
```diff
@@ -95,6 +95,16 @@ export interface NotificationTemplate {
   subjectTemplate: string
 }
 
+/**
+ * @public
+ */
+export interface NotificationPresentation {
```
aonnikov: maybe rename this to `NotificationBody` or `NotificationContent`?

plugins/notification/src/index.ts
aonnikov: I suppose that it is both title + body because translation will ignore parameters that do not present in the translated string, but I would prefer separate sets of parameters.

packages/text/src/plain-text.ts
mixerka: No problem, will do this with tiptap! 

However this approach of making a plain text from HTML based on plain js sax parser implementation has its benefits. We are not using tiptap to show activity messages, only to edit and so we are rely on HTML representation. This approach is quick and doesn't require any additional assumptions.

plugins/notification/src/index.ts
mixerka: I made it optional to mitigate issues with a previous existing data and migration. We've postpone the idea of renaming things. We've started from the idea to change Inbox implementation too, but we've decided not continue with it and just add fields to create more beautiful desktop notifications. So no other impact than desktop notification with all of this.

plugins/notification/src/index.ts
mixerka: I have an idea to use intlParams for both of them. The reason for this is to be able to change the view of result notification only by changing the translations. You can easily move things from title to body back and forward and to see which options work better and this doesn't require to change where to store params to fulfil the translation

plugins/notification/src/index.ts
mixerka: senderName for some notification is a part of intlParams. sender itself stored in modifiedBy of transaction as was before. Again we decided to continue only with adding localization properties to current implementation.

server-plugins/tracker-resources/src/index.ts
mixerka: this is not for inbox presenter, this is for native notifications. So the idea behind this in a tracker is the following. When someone put  a comment in issue activity we want to show the start of the message. However the message itself could be really big, it could have headers and formatting, even table. We have no controls for it, but we allow to paste it. And for whole message having it plain text is might be not the best idea, because we loose rich formatting. But for native notifications we need just plain text.

In this case I think it's just too much to preserve whole message inside a inbox item. So I propose to cut it here for activity messages.

plugins/notification/src/index.ts
mixerka: I love NotificationContent

packages/text/src/plain-text.ts
haiodo: Markdown is hold in html tiptap/prosemirror presentation, so it is not full html syntax, only subset. we use it on server for backlinks processing. So natural will be to use it here

server-plugins/tracker-resources/src/index.ts
haiodo: I'm not sure why we have to distinguish them, better to have one solid InboxItem and have a notification based on it shown to user, or email was send/etc.

packages/text/src/html.ts
```diff
@@ -33,3 +34,48 @@ export function parseHTML (content: string, extensions: Extensions): ProseMirror
...
+const WHITESPACE = ' '
+
+/**
+ * @public
+ */
+export function stripTags (htmlString: string, textLimit = 0, extensions: Extensions | undefined = undefined): string {
+  const effectiveExtensions = extensions ?? defaultExtensions
+  const parsed = parseHTML(htmlString, effectiveExtensions)
+
+  const textParts: string[] = []
```
haiodo, line 48: Why not use `parsed.textContent`?

plugins/notification/src/index.ts
haiodo: I think we always intend to make our code base better, and not hold unattended stuff.

packages/text/src/html.ts
mixerka: Imagine you have block elements at start of your comment. Such as LI or P. In this case `parsed.textContent` gives you plain text without spaces between such blocks. Text from different block elements will be concatenated and glued together. 

For input 
`<h1 class="proseHeading">Header</h1><p>Paragraph 1 text</p><ol><li><p>Item 1</p></li><li><p>Item 2</p></li></ol>`

You will get the following `parsed.textContent`
` HeaderParagraph 1 textItem 1Item 2`

Or simpler, if you will have very short first paragraph.

I handled such cases.

mixerka:
So seems that I've fixed all issues with the code is there anything I still need to fix?

BykhovDenis (approved):

PR #3774, opened 2023-10-02
Add isEmpty extension to remove on:content event from collaborator editor

annano:
# Contribution checklist

## Checklist

* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues


packages/text-editor/src/components/extension/validateContent.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export interface ValidateContentStorage {
+  isEmpty: boolean
+}
+
+const handleChange = (editor: Editor, options: ValidateContentOptions, storage: ValidateContentStorage): void => {
+  if (options.onEmpty !== undefined && options.onEmpty !== null) {
+    if (storage.isEmpty !== editor.isEmpty) {
+      storage.isEmpty = editor.isEmpty
```
aonnikov: Should we update storage even when not `onEmpty` callback was passed?

packages/text-editor/src/components/extension/validateContent.ts
```diff
@@ -0,0 +1,43 @@
...
+      options.onEmpty(storage.isEmpty)
+    }
+  }
+
+  if (options.onValidate !== undefined && options.onValidate !== null) {
+    options.onValidate(editor.getHTML()) // TODO: might need to call async ?
```
aonnikov: Yes, maybe async ... 
When I tried to edit large documents `getHTML` was slow. Maybe we should pass editor so the calling code will decide how to perform validation (check html, prosemirror json, or text).

packages/text-editor/src/components/extension/validateContent.ts
lexiv0re: Is `editor.isEmpty` = `true` for `<p></p>` value in the editor?

aonnikov: It depends on the schema. It checks whether content matches the default content. If the content is block, then yes, `editor.isEmpty` will be `true` for `<p></p>`

https://github.com/ueberdosis/tiptap/issues/580#issuecomment-826398844
https://github.com/ueberdosis/tiptap/blob/b359f137e9777200e2483b1754ec1987e7d59765/packages/core/src/Editor.ts#L452

annano: Yes

annano: Updated

lexiv0re: ok, great

packages/text-editor/src/components/extension/validateContent.ts
annano: I moved the extension for only `isEmpty` state, validation extension should probably be more complicated.

aonnikov (approved):

PR #3798, opened 2023-10-05
UBER-911: Mentions without second input and tabs

mixerka:
# Contribution checklist

## Brief description

Screenshot 2023-10-05 at 20 23 33

## Checklist

* [x] - Does it well tested?
* [x] - Tested for Chrome.
* [x] - Tested for Safari.

## Related issues


packages/presentation/src/types.ts
```diff
@@ -34,15 +34,28 @@ export interface ObjectCreate {
...
+ */
+export type ObjectSearchContext = 'search' | 'mention'
+
 /**
  * @public
  */
 export interface ObjectSearchCategory extends Doc {
   label: IntlString
   icon: Asset
+  title: IntlString
```
aonnikov, line 67: You will need to update Uberflow as well.

packages/text-editor/src/components/MentionPopup.svelte
```diff
@@ -0,0 +1,173 @@
...
+
+  let list: ListView
+  let scrollContainer: HTMLElement
+  let selection = 0
+
+  function dispatchItem (item: ObjectSearchResult): void {
+    dispatch('close', item)
+  }
+
+  export function onKeyDown (key: KeyboardEvent): boolean {
```
aonnikov, line 57: I have a problem with closing this popup when no results found. Maybe we need to add better handling for Esc, Enter, Tab, RightArrow buttons. Currently only adding double spaces works.

haiodo:
@mixerka Could you please rebase? Thanks.

packages/core/src/storage.ts
```diff
@@ -208,6 +208,42 @@ export type FindResult<T extends Doc> = WithLookup<T>[] & {
...
+/**
+ * @public
+ */
+export type FulltextQueryOptions = any
+
+/**
+ * @public
+ */
+export interface FulltextSearchResult {
+  suggest?: any
```
haiodo: May we elaborate on this a bit more?

packages/core/src/storage.ts
```diff
@@ -217,5 +253,13 @@ export interface Storage {
...
   ) => Promise<FindResult<T>>
+
   tx: (tx: Tx) => Promise<TxResult>
 }
+
+/**
+ * @public
+ */
+export interface FulltextStorage {
+  searchFulltext: (query: FulltextQuery, options: FulltextQueryOptions) => Promise<FulltextSearchResult>
```
haiodo: May be we just name it search? and SearchQuery? and SearchOptions?

plugins/tracker-resources/src/index.ts
```diff
@@ -152,28 +152,31 @@ export { default as AssigneeEditor } from './components/issues/AssigneeEditor.sv
 
 export { CreateProject, IssuePresenter, TitlePresenter }
 
+const DEFAULT_LIMIT = 200
 export async function queryIssue<D extends Issue> (
```
haiodo: I think all of this will be removed after we switch to new search?

server/elastic/src/__tests__/adapter.test.ts
```diff
@@ -38,7 +37,22 @@ describe('client', () => {
     console.log(hits)
   })
 
-  // it('should find document', async () => {
-  //   const adapter = await createElasticAdapter('http://localhost:9200/', 'ws1')
-  // })
+  it('should find document with raw search', async () => {
```
haiodo, line 42: We do not start elastic on tests, so we need to add it before we could write tests for elastic.

server/elastic/src/adapter.ts
```diff
@@ -100,6 +104,36 @@ class ElasticAdapter implements FullTextAdapter {
     return this._metrics
   }
 
+  async searchRaw (query: FulltextQuery, options: FulltextQueryOptions): Promise<FulltextSearchResult> {
```
haiodo: Better to have some FulltextQuery as structured query with all groupings  and limits and have it return a proper results.

server/middleware/src/spaceSecurity.ts
```diff
@@ -400,6 +405,54 @@ export class SpaceSecurityMiddleware extends BaseMiddleware implements Middlewar
...
+    ctx: SessionContext,
+    query: FulltextQuery,
+    options: FulltextQueryOptions
+  ): Promise<FulltextSearchResult> {
+    const newQuery = justClone(query)
+    const account = await getUser(this.storage, ctx)
+    let spaces: string[] = []
+    if (!isSystem(account)) {
+      spaces = await this.getAllAllowedSpaces(account)
+      const spacesFilter = {
```
haiodo: I'm not happy with unknown Elastic like syntax here, we need a structured/typed query and it should be handled inside adapter itself.

plugins/tracker-resources/src/index.ts
mixerka: Yes, it's true! And I could move it out from this PR too if we would like just accept the part with elastic changes

packages/core/src/storage.ts
mixerka: Do you mean naming or the structure of results?

server/elastic/src/__tests__/adapter.test.ts
mixerka: Oh I saw we had one test and just thought it would be nice to add more. Are there any problems with elastic in test runner? Or should I just add it there?

haiodo: You need to continue work on search API, since right now it is pure elastic, but it should be our minimal APIs suitable for our search. And in case we could like to replace elastic in future we will be able to do so with just adapter replacement.

server/middleware/src/spaceSecurity.ts
mixerka: I have looked at https://github.com/elastic/elasticsearch-js/blob/main/src/api/types.ts and thought that having all the possible typing for query it a bit too much. Maybe I wrong and we should take it? Or we could say that we only want to support some subset of what is possible, but I think it's against initial idea to be able to use full power of Elastic with specific request. 

Or I solve it quick way by passing spaces to adapter and the query will be not so typed, but the merge of spaces inside a query would happen in adapter and not here. 

Which approach do you like more?

plugins/tracker-resources/src/index.ts
haiodo: We should remove and switch to new API,

Also you need to continue work on search API, since right now it is pure elastic, but it should be our minimal APIs suitable for our search. And in case we could like to replace elastic in future we will be able to do so with just adapter replacement.

server/elastic/src/__tests__/adapter.test.ts
haiodo: No problem, we just not added it.

server/middleware/src/spaceSecurity.ts
haiodo: For me search API looks like following:

`search(mySearchString, options): SearchResult[]` and thats all.

Options could contain a limit on documents we would like to search, groupping and limits per individual classes etc.

packages/core/src/storage.ts
```diff
@@ -208,6 +208,49 @@ export type FindResult<T extends Doc> = WithLookup<T>[] & {
...
+  [key: string]: any
+}
+
+/**
+ * @public
+ */
+export interface SearchQuery {
+  query: string
+  aggregateBy?: string
+  filter?: {
```
haiodo: Is it document attributes? How it will be applied, could it be 'or' or 'and'? Could it be from multiple classes? like Employee and Organization fields?

packages/core/src/storage.ts
```diff
@@ -208,6 +208,49 @@ export type FindResult<T extends Doc> = WithLookup<T>[] & {
...
+    [key: string]: any
+  }
+}
+
+/**
+ * @public
+ */
+export interface SearchOptions {
+  limit?: number
+  offset?: number
```
haiodo: Why we need offset?

haiodo: Should we have sorting?

packages/core/src/storage.ts
```diff
@@ -217,5 +260,13 @@ export interface Storage {
...
   ) => Promise<FindResult<T>>
+
   tx: (tx: Tx) => Promise<TxResult>
 }
+
+/**
+ * @public
+ */
+export interface FulltextStorage {
+  searchFulltext: (query: SearchQuery, options: SearchOptions) => Promise<SearchResult>
```
haiodo, line 267: Should we design it with some streaming capabilities? Like we start search, receive some results and immediately display them, and send more requests in context like to have more items of particular type, or if some more data is arriving?

packages/core/src/storage.ts
mixerka: Right now It's elastic document attributes not our Doc attribute and it only could be 'and'. So as we are dealing with elastic documents there is no knowledge about our classes and if the user want to make filters on class attributes he has to use docKey or something on his own. Right now this is about what we need for mentions now. 

And we could improve this in several ways, just let's discuss what do we want to have right now.

What could we improve:
1. Make more clear structure to be able to make more complex conditions. We even could make it mongo-like conditions. Or we could just add more like "contains" and "contains not". Do work on this I would need to make some decent tests for elastic because I cannot see scenarios for this in mentions right now.
2. We could introduce docKey like mapping inside elastic adapter. I'm not sure how deep we want to get the knowledge about our hierarchy there, but this is obviously possible solution.

packages/core/src/storage.ts
mixerka: I put offset there to be able to build pagination for results. I'm not sure that we would need it in mentions or spotlight, but it might be useful if we will want to introduce "show more".

I thought about sorting, but I believe in search we always want to use scoring so we would have a question how to tune the scoring in general and how to make search more relevant.

packages/core/src/storage.ts
mixerka: It would be interesting. I assume you are talking about Scroll API. I never saw using of it in user-facing search scenarios, only for large amounts of data such as logs or something. So I don't see any usage of it for mentions and spotlight, but if you could give some clue, some usage example for our system, I would be happy to build such API.

packages/core/src/storage.ts
haiodo: I'm not sure about usage scenario to pass _class filter here. Probable we need to have it in parameters. And we need to think about more examples to use filters to fully be sure about do we need it or not.

packages/core/src/storage.ts
haiodo: I think we do not need a pagination, assume we have many results, user will type more to find a goal.

packages/core/src/storage.ts
mixerka: But now search is available only for certain categories https://github.com/hcengineering/anticrm/blob/main/models/contact/src/index.ts#L711
We have ObjectSearchCategory in a model and it gives us necessary labels and icons. And in PR it gives also a component to render result. 

If we wouldn't filter by _class how can we ensure that we only asking for results that we need? Example: Backlinks. They are in index for some reason but we do not want to search for them. I could image there might be more object which shouldn't be available through the common search, or which shouldn't be mentioned. How to deal with it without filter by _class?

packages/core/src/storage.ts
mixerka: No problem, I will remove it for now, it's not hard to add it if we need it

packages/core/src/storage.ts
```diff
@@ -208,6 +208,47 @@ export type FindResult<T extends Doc> = WithLookup<T>[] & {
...
+  limit?: number
+}
+
+/**
+ * @public
+ */
+export interface SearchResult {
+  // suggest?: any
+  // aggregations?: any
+  docs: IndexedDoc[]
```
haiodo: Why we need IndexedDoc here? I supposed we had a clear structure what is visible on client for mentions and spotlight.

mixerka: We do! Thanks, will fix this

packages/core/src/storage.ts
```diff
@@ -208,6 +208,72 @@ export type FindResult<T extends Doc> = WithLookup<T>[] & {
...
+ */
+export interface SearchOptions {
+  limit?: number
+}
+
+/**
+ * @public
+ */
+export interface SearchResultDoc {
+  id: Ref<Doc>
```
haiodo, line 233: I expected we will have same as for Inbox, like a record of parameters and a labels with IntlString, so UI will render them in proper places and will not decide about details.

So all after space should be ommited, except icon and probable color.

In this case no IndexedReaders and presenters will be required.

Thoughts?

mixerka: For notifications a presentation was relatively simple because it was a plain text. Now every component do something with fields before show them. Even name need to be processed before rendering. And there are avatar, issue's status which gives different icon and etc.

Now it's obvious which fields could be in SearchResultDoc. We definitely could replace this strict definition with JSON-like key-value structure, but what we could actually improve by that? I mean the way we retrieve value from IndexedDoc would be the same, and how we using it in component too. 

Or maybe do you mean we should change the way how we store data in index? But even then there would be an issue about what you need to know to show fields properly.

packages/core/src/storage.ts
```diff
@@ -208,6 +208,68 @@ export type FindResult<T extends Doc> = WithLookup<T>[] & {
...
+  _class: Ref<Class<Doc>>
+  space: Ref<Space>
+  icon?: string
```
haiodo: Asset? or something other?

packages/core/src/storage.ts
```diff
@@ -208,6 +208,68 @@ export type FindResult<T extends Doc> = WithLookup<T>[] & {
...
+  iconComponent?: string
+  iconProps?: { [key: string]: string }
+  objectId?: string
+  title?: string
```
haiodo, line 238: IntlString and params?

packages/core/src/utils.ts
```diff
@@ -331,3 +339,39 @@ export class RateLimitter {
     await await Promise.race(this.processingQueue.values())
   }
 }
+
+/**
+ * @public
+ */
+export interface IndexedReader {
```
haiodo: Do we still need it here?

packages/model/src/dsl.ts
```diff
@@ -244,6 +246,13 @@ export function UX<T extends Obj> (
   }
 }
 
+export function SEARCH<T extends Obj> (searchConfig: ClassSearchConfig) {
```
haiodo: Why it is Upper case?

packages/core/src/storage.ts
```diff
@@ -208,6 +208,68 @@ export type FindResult<T extends Doc> = WithLookup<T>[] & {
...
+ * @public
+ */
+export interface SearchResultDoc {
+  id: Ref<Doc>
+  _class: Ref<Class<Doc>>
+  space: Ref<Space>
+  icon?: string
+  iconComponent?: string
+  iconProps?: { [key: string]: string }
+  objectId?: string
```
haiodo: What is difference between id and objectId?

packages/core/src/storage.ts
```diff
@@ -208,6 +208,68 @@ export type FindResult<T extends Doc> = WithLookup<T>[] & {
...
+  title?: string
+}
+
+/**
+ * @public
+ */
+export interface SearchResultRaw {
```
haiodo: Why we need this one in API?

packages/core/src/storage.ts
```diff
@@ -208,6 +208,68 @@ export type FindResult<T extends Doc> = WithLookup<T>[] & {
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface TxResult {}
 
+/**
+ * @public
+ */
+export interface IndexedDoc {
```
haiodo: Why we need this one in API?

plugins/attachment-resources/src/components/AttachmentStyleBoxEditor.svelte
```diff
@@ -86,7 +86,13 @@
       haveUnsavedChanges = false
     }
 
-    await descriptionBox.createAttachments()
+    /*
```
haiodo: Why we need this changes in this PR?

packages/model/src/dsl.ts
mixerka: I looked at the @ UX and it looked as a nice idea, will fix this :)

packages/core/src/storage.ts
mixerka: Asset, will fix it

packages/core/src/storage.ts
mixerka: No needed for objectId and title. I've added them in IndexedDoc and fulfil them before storing the document in elastic here 
https://github.com/hcengineering/anticrm/pull/3798/files#diff-ffc4287c479a62d55793cdcd4550255db0216f7bb73bf5ada79c9811236e25dfR152

Maybe objectId is a bad naming, but I mean human readable object ID. After this issue https://github.com/hcengineering/anticrm/issues/3913 I thought that objectId is a really nice name for this.

packages/core/src/utils.ts
mixerka: You are right, not here, will move it to the only place where I'm using it

packages/core/src/storage.ts
mixerka: Must be really not a good naming, yes. objectId is "TSK-15" . Like human readable one.

packages/core/src/storage.ts
mixerka: You are right, we don't need it here anymore, will move

packages/core/src/storage.ts
mixerka: Will move too

plugins/attachment-resources/src/components/AttachmentStyleBoxEditor.svelte
mixerka: My bad, I just reproduced this A LOT when tested mention links. And I thought it would be merged soon anough. Will move to a separated PR

packages/core/src/utils.ts
```diff
@@ -129,10 +131,16 @@ export function docUpdKey (name: string, opt?: IndexKeyOptions): string {
  */
 export function docKey (name: string, opt?: IndexKeyOptions): string {
   const extra = opt?.extra !== undefined && opt?.extra?.length > 0 ? `#${opt.extra?.join('#') ?? ''}` : ''
-  return (
+  let key =
     (opt?.docId !== undefined ? opt.docId.split('.').join('_') + '|' : '') +
     (opt?._class === undefined ? name : `${opt?._class}%${name}${extra}`)
-  )
+  if (opt?.refAttribute !== undefined) {
```
haiodo, line 137: do we still need this changes?

haiodo (approved):

PR #3882, opened 2023-10-24
EZQMS-306: add extensions for chunter message version

annano:
# Contribution checklist

## Checklist

* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues


plugins/chunter/src/index.ts
```diff
@@ -62,6 +62,13 @@ export interface ChunterMessage extends AttachedDoc {
   reactions?: number
 }
 
+/**
+ * @public
+ */
+export interface ChunterMessageExtension extends ChunterMessage {
+  mixinClass: Ref<Class<Doc>>
```
lexiv0re: Will it be simpler to put `presenter` here directly instead of a class that is then used to get the presenter?

aonnikov: Presenters should be part of the model, we should not store ref to class / presenter in every object.

annano: If we store presenter here, in case we want to change the presenter we would need to migrate all existing objects in the database.

plugins/chunter-resources/src/components/Message.svelte
```diff
@@ -83,6 +82,16 @@
 
   $: isEditing = false
 
+  let extensions: Ref<Mixin<ChunterMessageExtension>>[] = []
```
haiodo, line 85: Should ObjectPresenter works fine with Hiearchy.findMixinMixins?

annano: I am not sure I understand, but the `ObjectPresenter` in this case showing presenter for main message class `_class` and not for the mixin.

plugins/view-resources/src/components/MixinPresenter.svelte
```diff
@@ -0,0 +1,53 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+-->
+<script lang="ts">
+  import type { Class, Doc, Ref } from '@hcengineering/core'
+  import { getClient } from '@hcengineering/presentation'
+  import { AttributeModel } from '@hcengineering/view'
+  import { getObjectPresenter } from '../utils'
+
+  export let mixinClass: Ref<Class<Doc>>
```
lexiv0re: Minor: should it be `Ref<Mixin<Doc>>`?

plugins/chunter/src/index.ts
lexiv0re: Yes, you are right.

plugins/view-resources/src/components/MixinPresenter.svelte
lexiv0re, line 1: Can we just pass `_class` parameter here https://github.com/hcengineering/anticrm/blob/25086b8348981af99698c95f234b41115be6dac7/plugins/view-resources/src/components/ObjectPresenter.svelte#L59 (if present) instead of `doc._class` when some new property (e.g. `exact`) is set to `true`? Otherwise, it feels unnecessarily duplicated.

annano: Updated

plugins/view-resources/src/components/MixinPresenter.svelte
annano: Removed

lexiv0re (approved):

haiodo:
@lexiv0re Why we merged it?

annano:
> @lexiv0re Why we merged it?

I have merged it, because it blocks my PR into qms.
These are changes into chunter, and it is not used nor supported as I understand.

PR #3895, opened 2023-10-26
UBER-1027: don't update issue space in kanban view

kristina-fefelova:
# Contribution checklist

## Checklist

* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

### DIFF
--- a/plugins/tracker-resources/src/components/issues/KanbanView.svelte
+++ b/plugins/tracker-resources/src/components/issues/KanbanView.svelte
@@ -245,8 +245,7 @@
       return
     }
     return {
-      [groupByKey]: groupValue,
-      space: doc.space
+      [groupByKey]: groupValue
     }
   }
 
--- a/plugins/tracker-resources/src/components/issues/KanbanView.svelte
+++ b/plugins/tracker-resources/src/components/issues/KanbanView.svelte
@@ -245,7 +245,8 @@
       return
     }
     return {
-      [groupByKey]: groupValue
+      [groupByKey]: groupValue,
+      space: doc.space
     }
   }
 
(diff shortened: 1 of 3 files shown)


plugins/tracker-resources/src/components/issues/KanbanView.svelte
```diff
@@ -245,8 +245,7 @@
       return
     }
     return {
-      [groupByKey]: groupValue,
-      space: doc.space
```
haiodo: We need to check if space update is not required. You could try diffUpdate

BykhovDenis: @haiodo which case it can be required?

kristina-fefelova: @haiodo I couldn't find a case where we need to update `space`, so I decided that we can just get rid of it 

If I use `diffUpdate` in generic `Kanban`, this also work, thanks

haiodo: It could be in case of all issues, and move between spaces occur, but I've not sure how we handle it here.

haiodo: So probable we need to add one more issue to check it

haiodo (approved):

plugins/tracker-resources/src/components/issues/KanbanView.svelte
BykhovDenis: Now it will be broken, because diffUpdate skip space changes...

haiodo: heh

haiodo (approved):

PR #3960, opened 2023-11-09
UBERF-4229: Fix createAttachments runtime error

mixerka:
# Contribution checklist

## Checklist

* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

### DIFF
--- a/plugins/attachment-resources/src/components/AttachmentStyleBoxEditor.svelte
+++ b/plugins/attachment-resources/src/components/AttachmentStyleBoxEditor.svelte
@@ -20,7 +20,7 @@
   import { navigate } from '@hcengineering/ui'
   import view from '@hcengineering/view'
   import { getObjectLinkFragment } from '@hcengineering/view-resources'
-  import { createEventDispatcher, onDestroy } from 'svelte'
+  import { createEventDispatcher } from 'svelte'
   import AttachmentStyledBox from './AttachmentStyledBox.svelte'
 
   export let object: Doc
@@ -74,6 +74,8 @@
       return
     }
 
+    descriptionBox.createAttachments()
+
     const old = getAttribute(client, object, key)
     if (description !== old) {
       await updateAttribute(client, object, object._class, key, description)
@@ -85,11 +87,6 @@
     } else {
       haveUnsavedChanges = false
     }
-
-    /*
-      Delayed save can be invoked after dismount. In this case descriptionBox would be null
-    */
-    await descriptionBox?.createAttachments()
   }
 
   let saveTrigger: any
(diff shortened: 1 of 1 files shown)


plugins/attachment-resources/src/components/AttachmentStyleBoxEditor.svelte
```diff
@@ -86,7 +86,12 @@
       haveUnsavedChanges = false
     }
 
-    await descriptionBox.createAttachments()
+    /*
+      Delayed save can be invoked after dismount. In this case descriptionBox would be null
+    */
+    if (descriptionBox !== undefined && descriptionBox !== null) {
+      await descriptionBox.createAttachments()
```
haiodo:
```suggestion
      await descriptionBox?.createAttachments()
```

haiodo: Just ? is required without if

mixerka: Fixed!

plugins/attachment-resources/src/components/AttachmentStyleBoxEditor.svelte
```diff
@@ -86,7 +86,10 @@
       haveUnsavedChanges = false
     }
 
-    await descriptionBox.createAttachments()
+    /*
+      Delayed save can be invoked after dismount. In this case descriptionBox would be null
+    */
+    await descriptionBox?.createAttachments()
```
haiodo: One more think about it, on dismount we need to force save instead to not loose state potentially.

mixerka: Thanks, added a handler for that too

haiodo: Nope, you need in case of haveUnsavedChanges, do clearTimeout and call save(saveObject, saveDescription)

plugins/attachment-resources/src/components/AttachmentStyleBoxEditor.svelte
```diff
@@ -102,6 +105,17 @@
     saveTrigger = setTimeout(() => save(saveObject, saveDescription), 2500)
   }
 
+  onDestroy(async () => {
+    // If change text and use mention link inside TextArea
```
haiodo: Nope, you need in case of haveUnsavedChanges call save(saveObject, saveDescription)

mixerka: Nope, we are not in THAT case. The reason this case was that save is async at the start descriptionBox is valid reference, but while await updateAttribute it become invalid one. save already when needed. 

But, I still came up with cleaner solution, because the order is not relevant we could just call createAttachments while reference is still valid.

haiodo (approved):

PR #4011, opened 2023-11-18
Added search customization to DropdownLabels

sol-0:
# Contribution checklist

## Checklist

* [x] - Does it well tested?
* [x] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

### DIFF
--- a/packages/ui/src/components/DropdownLabels.svelte
+++ b/packages/ui/src/components/DropdownLabels.svelte
@@ -42,6 +42,8 @@
   export let minW0 = true
   export let disabled: boolean = false
 
+  export let withSearch: boolean = true
+
   let container: HTMLElement
   let opened: boolean = false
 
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/DropdownLabels.svelte
```diff
@@ -42,6 +42,8 @@
   export let minW0 = true
   export let disabled: boolean = false
 
+  export let withSearch: boolean = true
```
haiodo: Why not use the enableSearch?

sol-0: To make it consistent with `Dropdown` which has `withSearch` prop :)
Renamed to `enableSearch` here anyway, thanks!

aonnikov (approved):

packages/ui/src/components/DropdownLabels.svelte
haiodo: In this case we need to make it consistent in all places.

PR #4211, opened 2023-12-17
[WIP] TSK-1668: Surveys

anotherpit:
# Contribution checklist

## Brief description

First steps building Surveys plugin. Work In Progress. Ready for early code review. But not ready for merge into `main` yet.

Screenshot 2023-12-18 at 23 38 08

## Checklist

* [x] - Does a local formatting is applied (rush format)
* [x] - Does a local svele-check is applied (rush svelte check)
* [ ] - Does it well tested?
* [x] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

## Contributor requirements


plugins/survey-resources/src/functions/surveyCreate.ts
```diff
@@ -0,0 +1,24 @@
...
+  const owner = getCurrentAccount()
+  const id = generateId<Survey>()
+
+  return await getClient().createDoc(
+    survey.class.Survey,
+    core.space.Space,
+    {
+      ...object,
+      questions: 0,
+      description: `Survey ${object.name}`,
```
aonnikov: why overwrite description?

models/survey/src/index.ts
```diff
@@ -14,25 +14,146 @@
...
+  TRadioButtons,
+  TSurvey,
+  TTypeRank
+} from './types'
 
+export { surveyOperation } from './migration'
 export { surveyId } from '@hcengineering/survey'
 export { default } from './plugin'
 
+// TODO: Should we place it into plugin itself, e.g. survey.routingParts.surveyId?
```
anotherpit: I definitely don't like routes or route parts hardcoded as raw strings all over the place, when you can't Ctrl+Click it, find all usages or refactor. So enum or string constants are must. The only question is how we should expose it to other packages. I suggest make this decision later, once we need it outside for the very first time.

models/survey/src/types.ts
```diff
@@ -13,6 +13,162 @@
...
+/**
+ * @public
+ */
+@UX(survey.string.Rank)
+@Model(survey.class.Rank, core.class.Type)
+export class TTypeRank extends TType {}
+
+// FIXME: Currently, classes that extend TSpace and represent private spaces
+//  cannot be stored in their own domain, as it breaks SpaceSecurityMiddleware.
+//  @Model(survey.class.Survey, core.class.Space, DOMAIN_SURVEY)
```
anotherpit: The issue I've mentioned during last daily meeting

models/survey/src/types.ts
```diff
@@ -13,6 +13,162 @@
...
+
+  @Prop(ArrOf(TypeCheckboxesOption()), survey.string.Options, { defaultValue: [] })
+    options: RadioButtonsOption[] = []
+}
+
+/**
+ * @public
+ */
+@Model(survey.class.Info, survey.class.QuestionData)
+export class TInfo extends TQuestionData implements Info {
```
anotherpit: Not supported in UI yet. I need this type as an example of non-interactive and non-assessible question data type

packages/core/src/operations.ts
```diff
@@ -103,7 +103,7 @@ export class TxOperations implements Omit<Client, 'notify'> {
     space: Ref<Space>,
     attachedTo: Ref<T>,
     attachedToClass: Ref<Class<T>>,
-    collection: string,
+    collection: Extract<keyof T, string> | string,
```
anotherpit, line 106: Strictly speaking, it should be just `Extract<keyof T, string>`, without `| string`, but I could not fix it quickly because it affected too many places, and brought too many warnings/errors. So for now it only helps IDE autocomplete better, but doesn't help TS to do type check

packages/core/src/utils.ts
```diff
@@ -235,7 +235,8 @@ export function fillDefaults<T extends Doc> (
   for (const attribute of attributes) {
     if (attribute[1].defaultValue !== undefined) {
       if ((object as any)[attribute[0]] === undefined) {
-        ;(object as any)[attribute[0]] = attribute[1].defaultValue
+        // Clone default value as it might be an object (e.g. array)
```
anotherpit, line 238: Any objections?

plugins/survey-resources/src/components/OptionsQuestionDataEditor.svelte
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+   * Declared $$Props help TypeScript ensure that your component is a valid {@link QuestionDataEditorComponent}
+   * @see https://raqueebuddinaziz.com/blog/svelte-type-events-slots-and-props/#restprops-props
+   */
+  interface $$Props extends QuestionDataEditorComponentProps<Data> {}
+
+  export let object: Data
+  export let editable = true
+  export let submit: (data: Data) => Promise<void>
```
anotherpit: 1. Submit is passed as a callback prop here (instead of event), because it's a part of common interface for all existing and future question data editors, and some of them might need to call this callback inside to wait for its result

2. Also discussed with @aonnikov that for the first implementation we are OK to submit the whole `QuestionData` object in one transaction. If any question data editor will ever need a more fine-grained update transactions, we have a solution without breaking this interface

plugins/survey-resources/src/components/QuestionCollectionEditor.svelte
```diff
@@ -0,0 +1,185 @@
...
+
+<div class="antiSection">
+  <div class="antiSection-header">
+    <span class="antiSection-header__title">
+      <Label label={survey.string.Questions} />
+    </span>
+  </div>
+
+  {#if questions.length > 0}
+    {#each questions as question, index (question._id)}
```
anotherpit: @aonnikov You were right, `Table` and `ListView` do have fixed hardcoded height by design, so we cannot easily use them here

plugins/survey-resources/src/components/QuestionEditor.svelte
```diff
@@ -0,0 +1,88 @@
...
+  let dataEditorComponent: Promise<ComponentType>
+  $: if (object.data._class !== dataClass?._id) {
+    dataClass = hierarchy.getClass<Data>(object.data._class)
+    dataEditorComponent = getResource(hierarchy.as(dataClass, survey.mixin.QuestionDataEditor).editor)
+  }
+
+  let data: Data = object.data
+  $: if (!deepEqual(data, object.data)) {
+    data = object.data
+  }
```
anotherpit: Tricks to make live editing (without explicit Save/Cancel) work in question data editors without losing focus and excessive re-renders. We do not re-render nested question data editor if it already has the same (i.e. deeply equal) actual data, which typically means that data was changed by question data editor itself. But if we receive from server a _new_ data it'll trigger re-render of question data editor.

plugins/survey-resources/src/plugin.ts
```diff
@@ -13,7 +13,20 @@
...
+import { mergeIds, type Resource } from '@hcengineering/platform'
+import { type ComponentType } from 'svelte'
 
-export default mergeIds(surveyId, survey, {})
+export default mergeIds(surveyId, survey, {
+  component: {
+    OptionsQuestionDataEditor: '' as QuestionDataEditorComponentTypeRef<Checkboxes | RadioButtons>,
+    SurveyCreator: '' as Resource<ComponentType>,
+    SurveyNamePresenter: '' as Resource<ComponentType>,
+    QuestionCollectionEditor: '' as Resource<ComponentType>
```
anotherpit: Technically, `Resource<ComponentType> === AnyComponent`, but `AnyComponent` is declared in `@hcengineering/ui`, which it's not publishable (and `Resource` and `ComponentType` are declared in `svelte`), which causes transpilation errors saying something like `Cannot name an inferred type of exported default without referencing @hcengineering/ui/src/...`

plugins/survey/src/index.ts
```diff
@@ -14,90 +14,73 @@
...
- * Interactive survey item, e.g. a question
- */
-export interface SurveyInteractiveItem extends SurveyItem {
-  title: string
-}
+export * from './types'
 
 /**
  * @public
- *
```
anotherpit, line 53: BTW, what are these `@public` JS-doc sections for? Do we really need'em? Not a big deal, just curious

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,114 @@
...
+export type QuestionDataEditorComponentTypeRef<Q extends QuestionData> = Resource<
+ComponentType<QuestionDataEditorComponent<Q>>
+>
+
+/**
+ * @public
+ */
+export interface QuestionDataEditor<Q extends QuestionData = QuestionData> extends Class<Doc> {
+  editor: QuestionDataEditorComponentTypeRef<Q>
+}
```
anotherpit: Some TS magic so that TS could help us control typings in question data editors – they must implement pretty specific and exact interface, and that's how we achieve this

models/survey/src/index.ts
aonnikov: We can start with constants, but eventually will have to do something more smart, depending on use cases. For example, we can expose navigation function in the plugin, so other plugins have convenient interface for navigation but don't know details.

models/survey/src/types.ts
aonnikov: I would prefer to keep surveys in `survey` domain, but unfortunately that's not possible now.

models/survey/src/types.ts
```diff
@@ -13,6 +13,162 @@
...
+  @Prop(TypeBoolean(), core.string.Archived, { defaultValue: false })
+  @Hidden()
+  declare archived: boolean
+
+  @Prop(TypeBoolean(), core.string.Private, { defaultValue: true })
+  @Hidden()
+  declare private: boolean
+
+  @Prop(Collection(survey.class.Question), survey.string.Questions, { defaultValue: 0 })
+    questions: CollectionSize<Question> = 0
```
aonnikov: as long as we define description as `Markup`, we will need  some collection (`attachments`?) for attached files.

plugins/survey-assets/lang/ru.json
```diff
@@ -1,7 +1,22 @@
 {
   "string": {
+    "Checkboxes": "Чекбоксы",
```
aonnikov: Please change to something like `Множественный выбор`  and something similar for RadioButtons

models/survey/src/types.ts
aonnikov: We need to discuss this again, I feel like it is not right to mix assessible/non-assessible questions together.

plugins/survey/src/types.ts
aonnikov: nice idea, sometimes we have issues with component interfaces

plugins/survey/src/index.ts
aonnikov: it was for api doc builder, we never really used it but had to put these doc sections.

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,114 @@
...
+  rank: Rank
+  data: Data
+}
+
+/**
+ * @public
+ *
+ * Base marker interface for every question data implementation
+ */
+export interface QuestionData extends Obj {}
```
aonnikov: We will need to have answers somewhere

plugins/survey-resources/src/components/SurveyCreator.svelte
aonnikov, line 1: nit: we conventionally name such components `CreateSurvey` to make it easier to find them.

plugins/survey-resources/src/components/QuestionCollectionEditor.svelte
```diff
@@ -0,0 +1,185 @@
...
+    label: clazz.label,
+    icon: clazz.icon
+  }))
+  const query = createQuery()
+
+  let questions: Question[] = []
+  let isMoving = false
+
+  $: if (object.questions > 0) {
+    query.query<Question>(
```
aonnikov, line 54: remove `if` and just use query, otherwise this will create new query each time the `object` is updated

plugins/survey-resources/src/components/QuestionCollectionEditor.svelte
```diff
@@ -0,0 +1,185 @@
...
+  async function onClickMove (index: number, up: boolean): Promise<void> {
+    const question1 = questions[index]
+    const question2 = questions[index + (up ? -1 : 1)]
+
+    isMoving = true
+    const ops = client.apply(object._id)
+    await questionUpdate(ops, question1, { rank: question2.rank })
+    await questionUpdate(ops, question2, { rank: question1.rank })
+    await ops.commit()
+    isMoving = false
```
aonnikov, line 99: when you moved a question, what question do you see under your cursor, the same (moved) or new question? 😄 it might be expected that when you move a question up, you scroll list of questions up as well, to keep the same question focused so you can press Up button multiple times.

plugins/survey-resources/src/functions/surveyCreate.ts
anotherpit: It's not an overwrite. Description field is inherited from Space, but hidden in Survey, user doesn't provide it in UI

plugins/survey/src/types.ts
anotherpit: Yes, we will. Will be be added in further commits

plugins/survey-resources/src/functions/surveyCreate.ts
aonnikov: I missed this somehow. I thought that object contains description. In either case, we will need to show some content in survey before all questions to explain what this survey is about, provide some information, etc

anotherpit: That's exactly why we want `Info` questions – to be able to provide information blocks, and not only in the beginning, but also in any part of the survey

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,153 @@
...
+  weight: number
+}
+
+/** @public */
+// eslint-disable-next-line @typescript-eslint/no-empty-interface, @typescript-eslint/no-unused-vars
+export interface AnswerData<Q extends Question> {}
+
+/** @public */
+export interface QuestionOption {
+  label: Markup
```
aonnikov: not sure if we really need markup for answers, plain text should be enough

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/** @public */
+export interface SingleChoiceQuestion extends Question {
+  options: QuestionOption[]
+  shuffle: boolean
+}
+
+/** @public */
+export interface SingleChoiceAnswerData extends AnswerData<SingleChoiceQuestion> {
+  selection?: number
```
aonnikov: shouldn't it be required? it is a single choice, so there should be one and only one selection

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+/** @public */
+export interface MultipleChoiceAnswerData extends AnswerData<MultipleChoiceQuestion> {
+  selections: number[]
+}
+
+/** @public */
+export interface MultipleChoiceAssessmentData extends AssessmentData<MultipleChoiceQuestion> {
+  correctAnswer: MultipleChoiceQuestion
```
aonnikov: here and below, should it be answer data instead of question?

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/** @public */
+export interface ReorderAssessmentData extends AssessmentData<ReorderQuestion> {
+  correctAnswer: ReorderQuestion
+}
+
+/** @public */
+export interface SurveyRequest extends AttachedDoc<Survey, 'requests', Survey> {
+  // TODO: Should it be Employee, no?
+  // TODO: Should it be an array of assignees instead?
```
aonnikov: I guess we need one request per surveyed person

plugins/survey/src/types.ts
anotherpit: Well, it's definitely optional for an answer given by user. In case of authoring the question, yes, _typically_ it is required, but we can imagine a tricky question which is displayed as a single choice (e.g. checkboxes), but the correct is answer is no answer. I'll mark it with TODO and come back to it when I have the whole flow implemented.

plugins/survey/src/types.ts
anotherpit: Absolutely! I would bump into it when implementing UI for this, but it's nice to have an attentive reviewer, too! Thank you

plugins/survey/src/types.ts
anotherpit: Yeah, for now that's my thinking, too. Uberflow implementation has it as an array though. I'll leave this TODO until I have a complete UI flow implemented, and see how it plays out.

packages/core/src/utils.ts
aonnikov: nope

aonnikov:
overall LGTM, the current design should be good enough, if needed we can make adjustments in the model while working on UI

plugins/survey/src/types.ts
lexiv0re: One of the use cases in qms is when a training manager assigns a training to a group of people and wants to track its overall progress e.g. 18/29 passed the training. For this it might be easier to have it as an array.

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,151 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import { AttachedDoc, Class, CollectionSize, Doc, Markup, Ref, Space } from '@hcengineering/core'
```
lexiv0re: Minor: prefer importing types as `type`

plugins/survey/src/types.ts
lexiv0re: If we support zero answers as a right answer I'd expect it to be in a multiple choice question (with checkboxes where you can uncheck a checkbox) rather than in a single choice question (with radio buttons where if you selected one you cannot deselect it anymore but only change your selection).

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,151 @@
...
+}
+
+/** @public */
+export interface SingleChoiceAnswerData extends AnswerData<SingleChoiceQuestion> {
+  selection?: number
+}
+
+/** @public */
+export interface SingleChoiceAssessmentData extends AssessmentData<SingleChoiceQuestion> {
+  correctAnswer: SingleChoiceAnswerData
```
lexiv0re: Did you consider having one field in the `AssessmentData` that'd hold the correct answer structure for all questions and specific questions will only define the structure itself (define its type)? Otherwise, we may end up having different fields with correct answers for different questions which might prevent us from having some generic code (e.g. it may be handy somewhere to exclude the correct answer field from being transferred from the DB to the UI by excluding this field from the requested projections).

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,151 @@
...
+
+/** @public */
+export interface SingleChoiceQuestion extends Question {
+  options: QuestionOption[]
+  shuffle: boolean
+}
+
+/** @public */
+export interface SingleChoiceAnswerData extends AnswerData<SingleChoiceQuestion> {
+  selection?: number
```
lexiv0re: 1. Is it the selected option index? Can we make it clear by making the name more specific or adding a comment?
2. It might become complicated to work with indices when `shuffle` option is on. If I'm getting it right it will shuffle the options for the trainee so the indices in his training UI will be different from the indices in the original question. We may also need to store these indices so once they are shuffled for one trainee they are stable every time he opens his training. So it might be confusing which index this answer is referring to.

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,151 @@
...
+}
+
+/** @public */
+export interface ReorderAssessmentData extends AssessmentData<ReorderQuestion> {
+  correctAnswer: ReorderAnswerData
+}
+
+/** @public */
+export interface SurveyRequest extends AttachedDoc<Survey, 'requests', Survey> {
+  // TODO: Should it be Employee, no?
```
lexiv0re: Potentially we might want to be able to expose some questionnaires to external people later but might be `Employee` for now.

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,151 @@
...
+  results: CollectionSize<SurveyResult>
+  // TODO: maxAttempts?: number
+  // TODO: dueDate?: Timestamp
+}
+
+export interface SurveyResult extends AttachedDoc<SurveyRequest, 'results', Survey> {
+  answers: CollectionSize<Answer<any>>
+}
+
+export interface Answer<Q extends Question> extends AttachedDoc<SurveyResult, 'answers', Survey> {
```
lexiv0re: Do we need the answer to extend `Doc` and be a standalone entity in the DB? Feels like it might complicate the process of passing a training where instead of just updating the result you'll need to CRUD individual answers.

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,151 @@
...
+  questionClass: Ref<Class<Q>>
+  question: Ref<Q>
+  answer: AnswerData<Q>
```
lexiv0re: Minor: having the same name with the interface might be tricky here. One will usually address this field as `answer.answer`.

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,151 @@
...
+  score?: Fraction
```
lexiv0re: Minor: just a matter of preference but would it be simpler to think of scores in percentages rather than in fractions?

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,151 @@
...
+  // TODO: Should it be Employee, no?
+  // TODO: Should it be an array of assignees instead?
+  assignee: Ref<Person>
+
+  results: CollectionSize<SurveyResult>
+  // TODO: maxAttempts?: number
+  // TODO: dueDate?: Timestamp
+}
+
+export interface SurveyResult extends AttachedDoc<SurveyRequest, 'results', Survey> {
```
lexiv0re: We probably need something like `submitDate` to track whether the result has been submitted by the trainee as we'll need to create the result in the DB before it's submitted (the user might not finish the training right away and get back to it another day so we need to keep the intermediate answers meanwhile).

lexiv0re: Once submitted it might be handy to calculate and store the score for the whole result and whether the user passed (result compared agains the threshold) so we don't need to calculate it every time. Also, might then be simpler to have some filters/etc for this conditions for the results.
Is this applicable to basic surveys?

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+export type Rank = string
+
+/**
+ * ∈ [0, 1]
+ */
+export type Fraction = number
+
+/** @public */
+export interface Survey extends Space {
```
lexiv0re: Do we need some simple description for surveys? It might be handy to show in some previews/notifications.

plugins/survey/src/types.ts
```diff
@@ -0,0 +1,151 @@
...
+import { Person } from '@hcengineering/contact'
+import { ComponentType, SvelteComponent } from 'svelte'
+import { Resource } from '@hcengineering/platform'
+
+/**
+ * @public
+ *
+ * @link https://www.npmjs.com/package/lexorank
+ */
+export type Rank = string
```
lexiv0re: Minor: should it be defined in more common place? It's effectively reused across a lot of various plugins.

plugins/survey/src/types.ts
anotherpit: yeah, should have been highlighted by IDE. i'll double check it, thanks

plugins/survey/src/types.ts
anotherpit: yeah, i did consider it, but didn't want to make it obligatory and put any restrictions on form of it for now. at least until we have some UI for passing surveys/assessments implemented. we'll probably end up with either a standard general field for correct answers that we'll be able to cut unconditionally, or a way to provide a question-specific function to cut out that information. we'll see when we have the complete flow implemented

plugins/survey/src/types.ts
anotherpit: 1. it is. any suggestions on better naming?
2. not too complicated, and all the complexity is gonna be encapsulated within a question-specific component that'll be responsible for displaying a question to assignee and properly storing his answers. i do see how it can be implemented. let me provide some implementation for it first, and then we'll see how much of a complexity it is

plugins/survey/src/types.ts
anotherpit: might be also an application-specific thing. let's get back to it a bit later, when we build an app for eQMS

plugins/survey/src/types.ts
anotherpit: yeah, i see your point. but d'you have better naming alternatives?

plugins/survey/src/types.ts
anotherpit: i'm thinking about different scenarios how an assignee can take a survey/assessment. and some of them might imply showing only one question per time on the screen and not letting an assignee change his answer once it's given. in this case working with answers individually might be more convenient (while still allowing other scenarios)

plugins/survey/src/types.ts
anotherpit: yeah, may be

plugins/survey/src/types.ts
anotherpit: 100%. and maybe similar field should be in answer

anotherpit: absolutely. but we might want to add this denormalization data on later stages.

plugins/survey/src/types.ts
anotherpit: i still can't decide between a dedicate description field, or a separate question type that allows you put arbitrary rich text in any place of the survey. another option is to include description in a higher level entity (something like `Training` in eQMS that extends eQMS' `TDocument`)

plugins/survey/src/types.ts
anotherpit: well, i wouldn't call them _a lot_, but yeah, it's already used in several places. maybe it's time to move it to core

plugins/survey/src/types.ts
aonnikov: I would rather add a separate answer option with "no right answer" text. Otherwise, we will need to somehow mark questions that allow no answer, and be able to distinguish not answered and answered as "no right answer" questions.

PR #4271, opened 2023-12-26
TSK-1668: Side changes from Surveys

anotherpit:
# Contribution checklist

## Brief description

Side changes from https://github.com/hcengineering/platform/pull/4211 without Survey plugins

## Checklist

* [ ] - Does a local svele-check is applied (rush svelte check)
* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

## Contributor requirements


packages/core/src/classes.ts
```diff
@@ -83,10 +83,14 @@ export interface UXObject extends Obj {
 /**
  * @public
  */
-export interface AttachedDoc extends Doc {
-  attachedTo: Ref<Doc>
-  attachedToClass: Ref<Class<Doc>>
-  collection: string
+export interface AttachedDoc<
+  Parent extends Doc = Doc,
+  Collection extends Extract<keyof Parent, string> | string = Extract<keyof Parent, string> | string,
```
aonnikov, line 88: I think there can be some places where we don't create collection attributes for some reasons. Don't we have any problems with this now?

aonnikov (dismissed):

packages/core/src/classes.ts
haiodo: It should be for any non system class. So let's try :-)

plugins/view/src/index.ts
```diff
@@ -435,9 +435,23 @@ export type ViewActionInput = 'focus' | 'selection' | 'any' | 'none'
...
+export type ViewActionFunction<TDoc extends Doc = Doc, TParams = Record<string, any>> = (
+  doc: TDoc | TDoc[] | undefined,
+  evt: Event,
+  params?: TParams
+) => Promise<void>
+
+/**
+ * @public
+ */
+export type ViewActionAvailabilityFunction<TDoc extends Doc = Doc> = (
```
haiodo: Let's use not a TDoc, since it conflict with our model definition classes and could confuse somebody.

server/middleware/src/spaceSecurity.ts
```diff
@@ -95,6 +95,11 @@ export class SpaceSecurityMiddleware extends BaseMiddleware implements Middlewar
   }
 
   private async init (ctx: MeasureContext): Promise<void> {
+    // FIXME: The following line looks up for private spaces in DOMAIN_SPACE only.
+    //  Thus, if you have a model class that represents private spaces,
+    //  and try to store it in its own domain, e.g. `@Model(MyModel, Space, MY_DOMAIN)`,
```
haiodo: Nope, we disallow to change domain for derived classes, so we only need to extend dsl to throw an error if such case is detected.

plugins/view/src/index.ts
anotherpit: `<TSmth extends Smth>` is kinda standard for generics in TS when you need longer names, but I see how it clashes with our convention for class types, so let's come up with an alternative convention. How about `<GSmth extends Smth>` (_G_ standing for _generic_)?

packages/core/src/classes.ts
anotherpit: @aonnikov 

> Don't we have any problems with this now?

We have, and it's not a problem, it's a designed behavior covered by tests. We will add a missing collection attribute even if it's not present on the model class and only added by mixin.

That's exactly why we have a trailing `| string` part here. The first leading part will let IDE suggest you the most appropriate options, while the second part will still let you type whatever you actually need.

server/middleware/src/spaceSecurity.ts
anotherpit: Might be a good idea to implement this check here, and get rid of this comment then.

> we disallow to change domain for derived classes

Is it meant only for Space and its descendants? Or any class that has already declared its domain?

aonnikov: if the domain has already been defined in a class, descendant classes should not redeclare it.

plugins/view/src/index.ts
anotherpit: Nevermind, in this exact place we don't even need longer names

anotherpit: Fixed

server/middleware/src/spaceSecurity.ts
anotherpit: Implemented check (which already caught non-critical violation in `model-activity`), and dropped this comment

aonnikov (approved):

haiodo (approved):

PR #4412, opened 2024-01-23
EZQMS-527: Introduced `ActionButton` component

anotherpit:
# Contribution checklist

## Brief description

Introduced `ActionButton` component that lets you present registered view actions as buttons in UI keeping them DRY, i.e. without copy-pasting `label`, `icon`, `action`/`on:click`, `visibilityTester`/`{#if}`, etc.

## Checklist

* [ ] - Does a local svele-check is applied (rush svelte-check)
* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

## Contributor requirements


plugins/view-resources/src/actions.ts
```diff
@@ -50,6 +50,21 @@ export function getSelection (focus: FocusSelection, selection: SelectionStore):
...
+export function queryAction (
+  query: LiveQuery,
+  client: Client,
+  doc: Doc | Doc[],
+  id: Ref<Action>,
+  callback: (action: Action | undefined) => void | Promise<void>,
+  derived: Ref<Class<Doc>> = core.class.Doc,
+  mode: ViewContextType = 'context'
+): boolean {
+  return query.query(view.class.Action, { 'context.mode': mode, _id: id }, async (actions) => {
```
BykhovDenis: Why we need LiveQuery here, action is model object and shouldn't change without page reloading

anotherpit: but action's `visibilityTester` is a function whose result might change without a page reload. for example, imagine having a ‘send to review’ action button on issue page that is only available for issues in status `in progress`. once your click this button, it should move the issue to different status, and the button should disappear

BykhovDenis: Yes, but query doesn't know about it, and it will call own callback (which contain check visibility) only when this action modified (never I think)

anotherpit: yeah, got it, you're right

BykhovDenis (approved):

PR #4490, opened 2024-02-01
Fixes ugly looked image selection

### LINKED ISSUE #4466: [safari] Ugly image selection in text editor
In Safari there is an ugly looking selection around selected image in text editor. The issue is reproducible in Safari after updating to 17.3 (OSX Sonoma). In other browsers selection looks good.

1. It seem to be fixable with `-webkit-user-selection: none;` style applied to the `<p>` element containing the image but this element is controlled by Tiptap.
2. <p> element seem to be redundant here and it appears because for some reason Image extension is configured with `inline` parameter set to true.

bogdan-ts:
# Contribution checklist

## Brief description

Fixes #4466 The new pseudo-class :has is used because only latest version of Safari is affected by the error. If support is required in older browsers, I can add the css-has-pseudo plugin for postcss.

## Checklist

* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Contributor requirements

* [ ] - Sign-off is provided. [DCO](https://github.com/apps/dco)
* [ ] - GPG Signature is provided. [GPG](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

### DIFF
--- a/packages/theme/styles/_text-editor.scss
+++ b/packages/theme/styles/_text-editor.scss
@@ -193,6 +193,12 @@
   }
 }
 
+@supports (selector(:has(.text-editor-image-container))) {
+  p:has(> .text-editor-image-container) {
+    user-select: none;
+  }
+}
+
 .text-editor-image-container {
   display: flex;
   flex-direction: row;


aonnikov:
It will probably work, but the problem is deeper than just fixing styles. 
1. images probably don't need to be inlined. that would solve the problem, but this would probably break existing content
2. image selection should not lead to selection of the entire paragraph

bogdan-ts:
> It will probably work, but the problem is deeper than just fixing styles.
> 
> 1. images probably don't need to be inlined. that would solve the problem, but this would probably break existing content
> 2. image selection should not lead to selection of the entire paragraph

I chose most safer way to do that, because it not affected to existing content, but of course I can try to remove "inline: true" property from Image Node, but maybe you can help me in potentially problematic cases with images

haiodo:
Hi @ComixRu I think PR is good for short term fix. Could you please rebase and sign all commits. thanks.

haiodo (approved):

aonnikov:
> > It will probably work, but the problem is deeper than just fixing styles.
> > 
> > 1. images probably don't need to be inlined. that would solve the problem, but this would probably break existing content
> > 2. image selection should not lead to selection of the entire paragraph
> 
> I chose most safer way to do that, because it not affected to existing content, but of course I can try to remove "inline: true" property from Image Node, but maybe you can help me in potentially problematic cases with images

one of problematic cases that I can think of, is an existing document that was created with `inline` property set to `true`. I suppose that we set `inline` to `false`, the images won't be displayed in the document properly anymore.

bogdan-ts:
> Hi @ComixRu I think PR is good for short term fix. Could you please rebase and sign all commits. thanks.

Hi @haiodo, Done

Issue #4630, opened 2024-02-14
In development mode exist user unfriendly error message


haykeminyan:
ENV Development mode How to reproduce this error:

1. `sh ./scripts/fast-start.sh`
2. `cd dev/prod && rushx dev-server`
3. stop all docker-containers
4. try to enter with credentials "user1/1234"

While you want to enter the system with credentials that describe there user/customer could not understand the main issue

Current error message:

**Unknown error: JSON.parse: unexpected character at line 1 column 1 of the JSON data**

My offer to use the more precise description of the problem/bug with the current status:

**Unknown error: Failed to fetch: 504 Gateway Timeout**

haykeminyan:
Hello everyone! I just created a PR with fixing this small bug
here it is :=)

https://github.com/hcengineering/platform/pull/4635

haiodo:
[haykeminyan ](https://front.hc.engineering/browse/?workspace=platform&_class=contact%3Aclass%3APerson&_id=65ccb0f760b2a36a8d3cb629&label=haykeminyan%20) Thanks a lot for a contribution, could you please add Sign off and sign commit. Thanks you.

PR #4635, opened 2024-02-14
added more precise try/catch error message while logging in platform

haykeminyan:
I have added more precise try/catch emulating error, which will be more informative user-friendly for users when server will be down for example with 504 error


haykeminyan:
task description

https://github.com/hcengineering/platform/issues/4630

haiodo:
@haykeminyan Thanks a lot for a contribution, could you please add Sign off and sign commit. Thanks.

PR #4854, opened 2024-03-01
feat(planner): improve and reuse `Chip`

euaaaio:
# Contribution checklist

## Brief description

image

## Checklist

* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

## Contributor requirements

* [ ] - Sign-off is provided. [DCO](https://github.com/apps/dco)
* [ ] - GPG Signature is provided. [GPG](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

### DIFF
--- a/packages/ui/src/components/Chip.svelte
+++ b/packages/ui/src/components/Chip.svelte
@@ -14,9 +14,13 @@
 -->
 <script lang="ts">
   import { createEventDispatcher } from 'svelte'
-  import { IconClose, ButtonIcon } from '..'
+  import { IconClose, ButtonIcon, LabelAndProps, tooltip } from '..'
 
   export let label: string
+  export let size: 'small' | 'min' = 'small'
+  export let isRemovable: boolean = false
+  export let backgroundColor: string | undefined = undefined
+  export let showTooltip: LabelAndProps | undefined = undefined
 
   const dispatch = createEventDispatcher()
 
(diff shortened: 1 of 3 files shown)


kristina-fefelova:
Is it correct size of cross icon? Could you please check it?

packages/ui/src/components/Chip.svelte
```diff
@@ -14,9 +14,13 @@
...
 <script lang="ts">
   import { createEventDispatcher } from 'svelte'
-  import { IconClose, ButtonIcon } from '..'
+  import { IconClose, ButtonIcon, LabelAndProps, tooltip } from '..'
 
   export let label: string
+  export let size: 'small' | 'min' = 'small'
+  export let isRemovable: boolean = false
+  export let backgroundColor: string | undefined = undefined
+  export let showTooltip: LabelAndProps | undefined = undefined
```
kristina-fefelova: Maybe `tooltipProps` or something like?
now it sounds like it should be a function or boolean

packages/ui/src/components/Chip.svelte
```diff
@@ -33,44 +37,69 @@
...
+
+      &:active:after {
+        background-color: var(--global-ui-active-OverlayColor);
+      }
     }
 
-    &:hover:after {
-      background-color: var(--global-ui-hover-OverlayColor);
+    &.min {
+      height: auto;
```
kristina-fefelova: Should we set up fixed `min` height?

euaaaio:
> Is it correct size of cross icon? Could you please check it?

It’s ok. Same icon size for all sizes of Chip.

packages/ui/src/components/Chip.svelte
euaaaio: I think it would be better to name it something like just `tooltip`. Buy we already named it `showTooltip` for `Button`, `StyleButton` and 24 other components.

packages/ui/src/components/Chip.svelte
euaaaio: That’s an interesting question. 
We use `--global-min-Size: 20px (1.25rem)` for height in our Figma layouts. 
But in css variables we have `--global-min-Size: 1rem`

I suggest we keep it as it is for now. We will come back to this component and variable soon anyway.

packages/ui/src/components/Chip.svelte
kristina-fefelova: Let's rename to `tooltip`

euaaaio: Maybe it would be better to make another PR where we rename all tooltip properties in all components for consistency?

I can do it very quickly.

kristina-fefelova (approved):

Issue #4858, opened 2024-03-01
When creating a task, a project step is not created


kulik8v:
Problem: When I create a task in a project I try to specify the stage to which this task will belong. Since the required stage was not entered beforehand, I try to enter it as part of the task creation. But the stage is not created

Procedure:

1. I press "Create Task" in the project
2. Fill in the task name and description
3. click on "Stage" - the system offers to choose from the existing ones or create a new one.
4. I enter the name of the new stage
5. I press "enter"

   ![image](https://github.com/hcengineering/platform/assets/101204005/dafc6c9a-f12a-47f7-8e5a-288ed515a35d)
6. Nothing happens - the stage is not created

   ![image](https://github.com/hcengineering/platform/assets/101204005/ef948727-472d-479a-8ba9-725c5bdb20cb)

Expectations: When entering the name of a new stage and confirming the creation with the "enter" key - the stage should be created and applied to the task

Comment: If you exit from task creation and create a step in the corresponding area, and then select it in the task, it is applied

![image](https://github.com/hcengineering/platform/assets/101204005/15e13e03-70af-41cd-ab52-5a7334f9cbbb)

<sub>View in Huly <a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NWUyNTNmYjMwMWI3NWQxNzFmODVmN2MiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.byWq1Bq0qVcll-NxCwdsZhuLPpTB2v3cEWEnJrupREU">UBERF-5860</a></sub>

mmilad75:
I will be working on this issue

mmilad75:
I have a question regarding getting the current space. How should I get it? @kulik8v

mmilad75:
I'm facing an issue with re-rendering the modal when a new milestone is added. And it's my first time working with Svelte. Can you help me with it? @kulik8v

kulik8v:
> I'm facing an issue with re-rendering the modal when a new milestone is added. And it's my first time working with Svelte. Can you help me with it? @kulik8v 

hi :) unfortunately no, because I am not a developer.

mmilad75:
[...]

PR #4879, opened 2024-03-05
Improve TagsPopup

gsdstr:
# Contribution checklist

## Brief description

Quick Add action Selecting a created item On enter action Fill label title from search input

Improvements are suitable as a reference for #4858

## Checklist

* [ ] - Does it well tested?
* [ ] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

#4858

## Contributor requirements

<sub>View in Huly <a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NWU2ZjI3Yzk1ZmRhMGNjMDg5OWZhZWIiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.kjObU5OGBLxWlKcXM8-PD1NlEqFd88TI5XYi81hxO3g">UBERF-5883</a></sub>

### DIFF
--- a/plugins/tags-resources/src/components/CreateTagElement.svelte
+++ b/plugins/tags-resources/src/components/CreateTagElement.svelte
@@ -35,8 +35,10 @@
 
   export let targetClass: Ref<Class<Doc>>
   export let keyTitle: string = ''
+  export let prefillTitle: string = ''
+  export let quick: boolean = false
 
-  let title = ''
+  let title: string = prefillTitle
   let description = ''
   let color: number = 0
 
@@ -52,8 +54,13 @@
     color = getColorNumberByText(title)
   }
 
-  $: if (!categoryWasSet && categories.length > 0) {
-    category = findTagCategory(title, categories)
+  $: {
+    if (!categoryWasSet && categories.length > 0) {
+      category = findTagCategory(title, categories)
+      if (quick) { // TODO dirty solution, move all logic to separate class
+        void createTagElement()
+      }
+    }
   }
 
   export function canClose (): boolean {
@@ -78,7 +85,7 @@
     categoryItems = newItems
   })
 
-  async function createTagElenent () {
+  async function createTagElement () {
     const tagElement: Data<TagElement> = {
       title,
       description,
@@ -87,8 +94,8 @@
       category: category ?? tags.category.NoCategory
     }
 
-    await client.createDoc(tags.class.TagElement, tags.space.Tags, tagElement, tagElementId)
-    dispatch('close')
+    const res = await client.createDoc(tags.class.TagElement, tags.space.Tags, tagElement, tagElementId)
+    dispatch('close', res)
   }
   const showColorPopup = (evt: MouseEvent) => {
     showPopup(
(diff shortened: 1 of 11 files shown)


plugins/tags-resources/src/components/CreateTagElement.svelte
```diff
@@ -52,8 +54,13 @@
     color = getColorNumberByText(title)
   }
 
-  $: if (!categoryWasSet && categories.length > 0) {
-    category = findTagCategory(title, categories)
+  $: {
+    if (!categoryWasSet && categories.length > 0) {
+      category = findTagCategory(title, categories)
+      if (quick) { // TODO dirty solution, move all logic to separate class
```
aonnikov: Yeah, this is dirty. In fact, we don't need to use this component for "quick" label creation, it is enough to have an utility function for that.

plugins/tags-resources/src/components/CreateTagElement.svelte
```diff
@@ -35,8 +35,10 @@
 
   export let targetClass: Ref<Class<Doc>>
   export let keyTitle: string = ''
+  export let prefillTitle: string = ''
```
aonnikov: not the best name for this variable. Can we use `keyTitle` or `title` instead?

plugins/tags-resources/src/components/TagsPopup.svelte
```diff
@@ -101,6 +115,16 @@
     }
     return r
   }
+
+  async function onSearchKeydown (ev: KeyboardEvent): Promise<void> {
+    if (ev.code !== 'Enter') return
+    if (objects.length < 1) {
```
aonnikov: minor: there is a chance that by the time user hits Enter, the query did not return new objects and `objects.length` is 0, but the the label exists. As a result, we create duplicate label.

aonnikov (commented):
Thanks for your contribution! The idea is great, but need to do some additional changes.
And I'm not completely sure that we need "quick" label creation. In this approach there is no way for user to specify color for the label. Maybe instead of creating a label right away we need to show the user label creation dialog with pre populated label title. If he is ok with other label properties, he hit Enter once again and create the label. Think about it.

gsdstr:
> And I'm not completely sure that we need "quick" label creation.

I think it's necessary to give the user the ability to edit created labels. This will speed up issues creation and keep the ability to choose colors. Like this:
Screenshot 2024-03-07 at 13 21 56

aonnikov:
> > And I'm not completely sure that we need "quick" label creation.
> 
> I think it's necessary to give the user the ability to edit created labels. This will speed up issues creation and keep the ability to choose colors. Like this: Screenshot 2024-03-07 at 13 21 56

We definitely need to do this, but definitely not here. Tags/labels have larger scope than just tracker so there should be a more generic solution. Let's not focus on this right now. I propose to either go with "quick" creation or open tag creation dialog when user hits Enter.

aonnikov (dismissed):

aonnikov (approved):

PR #5047, opened 2024-03-24
Fix login form issue when using password auto-fill

vlad-tim:
# Contribution checklist

### Non-technical issue description

This PR fixes a minor UI issue with the login form when using Firefox password auto-fill feature. This video illustrates the issue.

https://github.com/hcengineering/platform/assets/11474041/20c093c1-0c07-4610-94c4-7fbdc61eb503

**Steps to reproduce:**

1. Use Firefox password manager to remember the login credentials.
2. Log in using the saved credentials.
3. Log out.

**Expected behavior:**

"Log in" button is enabled and allows to log in again using the saved credentials.

**Observed behavior:**

"Log in" button is disabled. It is possible to make it enabled again by editing the password field, e.g. by entering a dummy character and then removing it.

**Impact:**

This issue causes a minor inconvenience for a user who is affected and aware of it, and may confuse/block a new user who encounters it for the first time. I was not able to reproduce the issue in Chrome, and had no possibility to test in Safari. The issue might also affect users who use browser extensions to fill the password instead of the built-in browser functionality.

### Technical issue description

Input is validated using function `validate` https://github.com/hcengineering/platform/blob/18e8dcab5782ce159c2754f3bcd794c8143ed1d5/plugins/login-resources/src/components/Form.svelte#L189

This function is asynchronous, and prone to a race condition. The race condition happens when the password manager fills username and then fills the password. This creates two simultaneous validation events: the first one validates the form where the password is still empty, whereas the second validation call validates the complete form with non-empty password. Unfortunately, the first validation finishes after the second validation and flags the form as invalid https://github.com/hcengineering/platform/blob/18e8dcab5782ce159c2754f3bcd794c8143ed1d5/plugins/login-resources/src/components/Form.svelte#L75-L77

### Proposed solution

The proposed solution introduces a `Mutex` class that allows to synchronize async functions. In this case the Mutex is applied to the form validation function. It eliminates the race condition and fixes the problem described and illustrated above.

### Further thoughts

As a follow-up, the introduced class may be extracted to a better place to be reused by other components.

As an alternative solution, I considered using the existing library https://www.npmjs.com/package/async-mutex .

## Checklist

* [x] - UI test added to added/changed functionality?
* [x] - Does a local formatting is applied (rush format)
* [x] - Does a local svelte-check is applied (rush svelte-check)
* [x] - Does a local UI tests are executed [UI Testing](../tests/readme.md)
* [x] - Does it well tested?
* [x] - Tested for Chrome.
* [ ] - Tested for Safari.

## Related issues

None

## Contributor requirements

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjAwN2Q2MzdmZTg3NDA1MTFhYWVkNGYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.WMzGROWiIYEwfNk41HZFjWiidve9_knIB5WrsENw18A">Huly&reg;: <b>UBERF-6159</b></a></sub>

### DIFF
--- a/plugins/login-resources/src/components/Form.svelte
+++ b/plugins/login-resources/src/components/Form.svelte
@@ -31,7 +31,7 @@
   import { onMount } from 'svelte'
   import { BottomAction, getHref } from '..'
   import login from '../plugin'
-  import { Mutex } from '../mutex'
+  import { makeSequential } from '../mutex'
   import Providers from './Providers.svelte'
 
   interface Field {
@@ -67,19 +67,7 @@
 
   $: $themeStore.language && validate($themeStore.language)
 
-  const validationMutex = new Mutex()
-
-  // This function wraps around `validateAsync` to make its executions sequential / synchronized
-  async function validate (language: string): Promise<boolean> {
-    const unlockMutex = await validationMutex.lock()
-    try {
-      return await validateAsync(language)
-    } finally {
-      unlockMutex()
-    }
-  }
-
-  async function validateAsync (language: string): Promise<boolean> {
+  const validate = makeSequential(async function validateAsync (language: string): Promise<boolean> {
     if (ignoreInitialValidation) return true
     for (const field of fields) {
       const v = object[field.name]
@@ -114,7 +102,7 @@
     }
     status = OK
     return true
-  }
+  })
   validate($themeStore.language)
 
   let inAction = false
--- a/plugins/login-resources/src/components/Form.svelte
+++ b/plugins/login-resources/src/components/Form.svelte
@@ -31,6 +31,7 @@
   import { onMount } from 'svelte'
   import { BottomAction, getHref } from '..'
   import login from '../plugin'
+  import { Mutex } from '../mutex'
   import Providers from './Providers.svelte'
 
   interface Field {
@@ -66,7 +67,19 @@
 
   $: $themeStore.language && validate($themeStore.language)
 
+  const validationMutex = new Mutex()
+
+  // This function wraps around `validateAsync` to make its executions sequential / synchronized
   async function validate (language: string): Promise<boolean> {
+    const unlockMutex = await validationMutex.lock()
+    try {
+      return await validateAsync(language)
+    } finally {
+      unlockMutex()
+    }
+  }
+
+  async function validateAsync (language: string): Promise<boolean> {
     if (ignoreInitialValidation) return true
     for (const field of fields) {
       const v = object[field.name]
--- a/plugins/login-resources/package.json
+++ b/plugins/login-resources/package.json
@@ -6,10 +6,12 @@
   "license": "EPL-2.0",
   "scripts": {
     "build": "compile ui",
+    "test": "jest --passWithNoTests --silent",
     "build:docs": "api-extractor run --local",
     "format": "format src",
     "build:watch": "compile ui",
     "_phase:build": "compile ui",
+    "_phase:test": "jest --passWithNoTests --silent",
     "_phase:format": "format src",
     "_phase:validate": "compile validate"
   },
--- a/plugins/login-resources/src/__tests__/mutex.test.ts
+++ b/plugins/login-resources/src/__tests__/mutex.test.ts
@@ -1,0 +1,28 @@
+import { makeSequential } from '../mutex'
+
+describe('mutex', () => {
+  let results: number[]
+
+  beforeEach(() => {
+    results = []
+  })
+
+  async function waitAndPushResult (millis: number): Promise<void> {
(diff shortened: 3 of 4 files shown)


plugins/login-resources/src/components/Form.svelte
```diff
@@ -66,7 +67,19 @@
 
   $: $themeStore.language && validate($themeStore.language)
 
+  const validationMutex = new Mutex()
```
haiodo: Hi, thanks for figure this out.

I think we need a bit more generic solution, since we potentially have more cases like this one, there we should ignore results or be sore operations are completed one but one, but proposed solution looks a bit more complicated.

We usually use local Promise variable and add await inside update method, but missed in this particular place, if you have some use-full generic idea it would be great.

Thanks.

vlad-tim: Thanks for looking into this, I am happy to work on improving the proposed solution, but I need a few clarifications:

> 1. proposed solution looks a bit more complicated

Is it particularly about  _implementation_ of Mutex or its _usage_?

> 2. there we should ignore results or be sore operations are completed one but one

I didn't fully understand this part (probably there were some typos), could you rephrase?

> 3. We usually use local Promise variable and add await inside update method

Could you link an example?

---

Also, what would be the right package/place in this repository to put a generic utility like Mutex into?

haiodo: Hi @vlad-timofeev sorry for delay with response.

1. Usage one.

I suppose we should evolve your idea to something like this.

```typescript
# Some there in platform
  export function oneByOne<T extends (...args: Parameters<T>) => Promise<void>> (t: T): (...args: Parameters<T>) => Promise<void> {
    let previousCall: Promise<any> | undefined
    return async function (...args: Parameters<T>): Promise<void> {
      if (previousCall !== undefined) {
        await previousCall
      }
      previousCall = t(...args)
      return await previousCall
    }
  }
# A component code
  let a = 20
  const myUpdate = oneByOne(async (x: number) => {
    console.log('called with', x)
    await new Promise<void>((resolve) => {
      setTimeout(resolve, x)
    })
    a = x
    console.log('returned with', x)
  })
  myUpdate(200)
  myUpdate(100)
```
So oneByOne would hold a promise of previous execution and before calling myUpdate will wait for previous to complete, so if variables are changing in different order and we have some asyncs inside update function, we will use results of last one arguments. Because assume `something` for some previous arguments could work longer.

haiodo: Nope, this variant is not working as expected, this one is correct.

```typescript
 function reduceCalls<T extends (...args: Parameters<T>) => Promise<void>> (t: T): (...args: Parameters<T>) => Promise<void> {
    let nextCalls: (() => Promise<void>)[] = []
    return async function (...args: Parameters<T>): Promise<void> {
      const myOp = async (): Promise<void> => {
        await t(...args)
        nextCalls.splice(0, 1)
        const last = nextCalls[nextCalls.length - 1]
        nextCalls = [last]
        if (last !== undefined) {
          void last()
        }
      }
      nextCalls.push(myOp)
      if (nextCalls.length === 1) {
        await myOp()
      }
    }
  }

  let a = 20
  const myUpdate = reduceCalls(async (x: number) => {
    console.log('called with', x)
    await new Promise<void>((resolve) => {
      setTimeout(resolve, x)
    })
    a = x
    console.log('returned with', x)
  })
  myUpdate(500)
  myUpdate(100)
  myUpdate(50)
```

vlad-tim: Thanks for that. I will re-work this PR and push a new revision asap.

vlad-tim: @haiodo , I tested your variants locally , the second one printed a different result from what I expected though:

```
[user@huly] ➜  ~/workshops/promise_test npx tsx from_mr_ace.ts                                                                                                                                                                 INSERT
called with 500
returned with 500
called with 50
returned with 50
[user@huly] ➜  ~/workshops/promise_test                                                                                                                                                                                       
```

I used the function-wrapper interface that you proposed, and encapsulated the existing Mutex implementation inside. I appended some code to `mutex.ts` file in the 2nd commit to expose this interface and as a result it is just as simple to use as in your examples. I also added tests to verify that this functionality works as expected.

aonnikov (approved):

PR #5120, opened 2024-04-01
feat(tests): write categories creat then delete ui test

Ngoc0905:
- New ui test of categories page that creates then deletes a category


aonnikov:
Thanks for the contribution. I'm closing this because test code coverage increase is not exactly what we are looking for.

PR #5127, opened 2024-04-01
Feature: Creatable dropdown feature with create category in product

sarthak212:

Screenshot 2024-04-02 at 12 58 01 AM

Added functionality for creating new element in the dropdown if doesn't exist, it can be used for creating thing which only require single string like a category, tags, skills and so on.

* Added creatable flag in DropdownLabelsPopup which add option to add new item based on specific condition when creatable flag is passed as props, search has some value and if that element already doesn't exist in the dropdown
* Updated DropdownLabels with same flag creatable and modified the logic to update the selected item and also trigged the newoption binding if new element is created i.e parent element can handle it as we are only passing single string in dropdown.
* Updated CreateProduct component to pass the createable flag and created the category in case it's created from dropdown when creating new product.

There can still be multiple bugs in this, so please guide me through it if you find any.

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjBiMGU0OGU1MGZjYjI5NThiMDU3MDQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.gDCGNbLdVKJiTf-M8lo_F9ecBf1__pLFlfQ8N7Xt7RE">Huly&reg;: <b>UBERF-6260</b></a></sub>

### DIFF
--- a/packages/ui/src/components/DropdownLabels.svelte
+++ b/packages/ui/src/components/DropdownLabels.svelte
@@ -29,6 +29,7 @@
   export let placeholder: IntlString | undefined = ui.string.SearchDots
   export let items: DropdownTextItem[]
   export let multiselect = false
+  export let creatable = false
   export let selected: DropdownTextItem['id'] | Array<DropdownTextItem['id']> | undefined = multiselect ? [] : undefined
   export let allowDeselect: boolean = false
   export let showDropdownIcon: boolean = false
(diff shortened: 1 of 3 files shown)


aonnikov:
Hello! The idea is good, but I do not like the way it is implemented.

sarthak212:
Hi @aonnikov , by implementation i think you mean for dropdownlabel and dropdownlabelpop rather than createProduct component because creation of respective new element will be handled in parent component rather than dropdown.

For Dropdown i wasn't sure about it , i was thinking of returning whole object back from dropdownpop and similar with dropdown component because i way that we are only returing id of a dropdown and in case of new element there should be a flag to identify so that's why i created another bind for that flag, to not disrupt the current implementation as it's used in multiple places.

another proposition for this could be seperate component to handle creatable and rather than just returning/binding id we can return a object which may contain this format 
`{ id: string, label: string, isNew?: boolean}`
it may be benefitial in some case because then in dropdown component we don't have to find the label by looping over each element as tha value is present in variable itself and we can pass additional info about selected dropdown too like in case of currency dropdown we need currency code, currency icon($) and so on, this way format is straigtforward.

let me know if you think this may be better implementation, i am up for another PR or i may be wrong  so let me know if you mean anything else in previous implementation part which should change.

aonnikov:
Hey @sarthak212 
I think there already were several pull requests that tried to address the same problem. We even accepted one that implemented fast creation for labels. But even for labels, there was an issue - labels allow to specify not only label text, but also color. We don't have much objects that are as simple as labels. Usually dropdowns are used for more complex objects - statuses, employees, etc. These objects can't be created in one click, user will need to specify more parameters.

That being said, I'm sure that in general the feature might be useful, but it should initiate new object creation, not create it right away. Additionally, sometimes dropdowns allow to choose only values that match some condition. It may happen, that object created does not match these conditions and can't be used 😉 

Regarding the implementation, I don't like bound value. I would rather used a callback or a  special item passed from the outside that indicates that user wants to create a new item. I also don't like that CreateProduct is responsible for creation of these objects, if we user categories dropdown somewhere else, we will have to do something to avoid code duplication.

sarthak212:
Okay @aonnikov , thanks for the review , i'll try to keep these in mind for next PR

PR #5150, opened 2024-04-02
adding email validation

zapata124:
Simple PR to add email validation in CreateEmployee component. As of now a user is able to input any string greater than zero.

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjBjMzEwM2U1MGZjYjI5NThiMDVmNWUiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.Mrg2TcY5GvhYpPXO9xFE_zhaFknmz1IQAo74pFQLmaE">Huly&reg;: <b>UBERF-6291</b></a></sub>

### DIFF
--- a/plugins/contact-resources/src/components/CreateEmployee.svelte
+++ b/plugins/contact-resources/src/components/CreateEmployee.svelte
@@ -115,6 +115,14 @@
     }
     channels = channels
   }
+  function emailValidator (email: string): boolean {
+    return (
+      email.length > 0 &&
+      !!email.match(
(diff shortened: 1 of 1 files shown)


aonnikov:
Thanks for good attempt. This regular expression covers most of valid emails but not all 😉 
From my point of view, it does not make much sense to implement ideally precise email validation. There is always a chance that we use inaccurate regular expression and do not let user enter his correct email. It is less of a problem if we let them enter invalid emails.
It would be enough if we validated that it looks like an email, i.e. there is `@`, local part, and domain.

PR #5195, opened 2024-04-04
Single z-index for tooltips and popups

ThetaDR:
Override zIndex for cases the popup opens from tooltip (ex.: comments)

image

### DIFF
--- a/packages/ui/src/components/Popup.svelte
+++ b/packages/ui/src/components/Popup.svelte
@@ -38,7 +38,7 @@
     element={popup.element}
     onClose={popup.onClose}
     onUpdate={popup.onUpdate}
-    zIndex={(i + 1) * 500}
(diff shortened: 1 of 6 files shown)


haiodo (dismissed):

aonnikov:
to be honest, I don't like this approach. It is good as a temporary solution, but eventually overrides will produce more mess because we will have to manually manage all z-indexes.

Ideally, we need to get rid of library that is used in editor for popups and replace them with our popups. Alternatively, we can expose z-index counter or methods for accessing it and use it in text editor.

haiodo:
I think we need unified z-layer order allocator between popups and tooltips.

haiodo:
Or unify them under one place.

ThetaDR:
Discussed with @aonnikov, going to use svelte `use` directive for it, it should make it more stable so it should prevent some of the bugs compared to this approach

PR #5214, opened 2024-04-05
feat(help): added find bug button for easy navigation

### LINKED ISSUE #4816: Add "Find a bug?" button to huly
When I find a bug while using Huly, I have to search for this repository separately to create an issue

It would be convenient if there was a special button in the web application that would lead here

aeswibon:
# Bug Fix

## Proposed Change

- [x] Closes #4816
- [x] Adds a button for easy navigation to file a bug
- [x] Ensure that the changeset adheres to the [DCO guidelines](https://github.com/apps/dco).

## Screenshot
image


aonnikov:
I was about to reject this, but it looks like several person find this feature very useful. Maybe this is a good idea, indeed.
The only thing I'm concerned of, is the location of the button, because It might be difficult to find it. And we probably need to do something with the button's style and text.  Would it be better to name it `Report a bug`?

aeswibon:
Sure, I will update that.

aeswibon:
@aonnikov Also, what style and where should I place the button? Maybe we can move the button, in the workbench, right above the help & support button...

haiodo (dismissed):

aonnikov (approved):

PR #5228, opened 2024-04-07
Fix alignment of `AssigneeBox` button

breizeway:
After I got the Huly platform up and running locally, I quickly noticed an alignment bug on the `AssigneeBox` component on the "All Projects" in the "Tracker" section.

image

After doing a little digging, it looks like this bug was introduced in [PR #5013](https://github.com/hcengineering/platform/pull/5013) with the addition of the `h-full` class. This bug goes away if I remove that class, but--assuming that new class is necessary--it can be fixed in this instance by adding `flex items-center`. The size of the button was also inconsistent compared to others in the table, so I bumped it up to `large`.

image

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjExZjYzOTY4ZWY5NjM0ZDNmMDQyMjUiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.IY54VwCs1Ga4Ix5s8EgAZ7vlwsHCX1IsjD5HiKhgjps">Huly&reg;: <b>UBERF-6396</b></a></sub>

### DIFF
--- a/plugins/contact-resources/src/components/AssigneeBox.svelte
+++ b/plugins/contact-resources/src/components/AssigneeBox.svelte
@@ -76,7 +76,7 @@
 
   const client = getClient()
 
-  async function updateSelected (value: Ref<Person> | null | undefined) {
+  async function updateSelected(value: Ref<Person> | null | undefined) {
     selected = value
       ? $personByIdStore.get(value) ?? (await client.findOne(contact.class.Person, { _id: value }))
       : undefined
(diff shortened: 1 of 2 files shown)


aonnikov:
Looks good to me. Could you sign-off your commits to pass DCO check?

breizeway:
> Looks good to me. Could you sign-off your commits to pass DCO check?

Ah, of course. It should be ok now.

aonnikov:
It looks like there are formatting issues. Please resolve them when you have a chance.

PR #5240, opened 2024-04-08
TSK-1682: Introduced Schema module

anotherpit:
For upcoming implementation of Recruit Script feature we need a way to:

* let one group of users compose and configure forms of typed fields
* let other groups of users fill those forms with their data

Think of something like Google Forms or Typeform. And also think of JSON schema as a comprehensive and standard language to describe forms.

So here I introduce Schema module which provides a set of context-agnostic building blocks to let any other Huly module incorporate forms built by one users and filled by others. The first module to utilize Schema will be Recruit, but this will happen in the next PR.


aplatoff:
Just curious: why do not use `platform`'s model for that? Is there any limitations and it can not cover some cases?

models/schema/src/plugin.ts
```diff
@@ -0,0 +1,20 @@
...
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import schema, { schemaId } from '@hcengineering/schema-resources/src/plugin'
+import { mergeIds } from '@hcengineering/platform'
+
+export { schemaId } from '@hcengineering/schema'
+export default mergeIds(schemaId, schema, {})
```
anotherpit, line 20: So, `schema` plugin is effectively empty. Should we drop it and convert `schema-resources` into package, maybe? But keep in mind we still need `model-schema` plugin to register mixin

plugins/schema/src/types.ts
```diff
@@ -0,0 +1,23 @@
...
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import { type Class, type Obj, type Ref } from '@hcengineering/core'
+import type { JSONSchema7 } from 'json-schema'
+
+/** @public */
+export interface Schema extends Obj, JSONSchema7 {}
```
anotherpit, line 20: @aplatoff Let's move here, so we could resolve the thread later.

> Just curious: why do not use `platform`'s model for that? Is there any limitations and it can not cover some cases?

What exactly do you mean? The module does use `platform`'s native data structures like `Obj`, `Doc` and `Mixin`. No limitations there, they do perfectly fit our needs 👍🏽

aplatoff: I mean that since both the platform and JSON Schema are focused on modeling data structures, why do we need two different mechanisms for modeling?

aplatoff: Any comments on this? :) @anotherpit

anotherpit: Discussed on call. The suggested idea is to not reinvent the wheel and utilize `Class`es and `Attribute`s and to have a flow similar to that of the Class Settings screen

image

PR #5243, opened 2024-04-08
Fixed bug related to Invalid Image retrievals on multiple tabs.

CodeQueenTosin:
Bug: When opening multiple workspace tabs in the browser and trying to view an issue with an attached image in one of the tabs, the images are not displayed. This occurs because the Image API requires a token stored in the presentation-metadata-Token cookie, but when opening another workspace tab, the cookie gets overridden with the token associated with the new workspace. As a result, the first tab starts using the wrong token, leading to image retrieval failures.

Solution: To address the issue, I implemented a solution that ensures the correct token is used when fetching images. This involved modifying the fetchContent function to retrieve the token from the presentation-metadata-Token cookie and passing it along with the image request to the Image API. This solution ensures that images are displayed correctly, even when multiple workspace tabs are open simultaneously.

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjEzZWIwMzc5NmRkMzAxMWUwZTQ0NzUiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.rvAHCx4fXUGBOExCnOhXPsw2XPEZRqU0ZQfoU2bPy0c">Huly&reg;: <b>UBERF-6418</b></a></sub>

### DIFF
--- a/packages/text-editor/src/provider/minio.ts
+++ b/packages/text-editor/src/provider/minio.ts
@@ -26,9 +26,10 @@ async function fetchContent (doc: YDoc, name: string): Promise<void> {
   for (let i = 0; i < 2; i++) {
     try {
       const frontUrl = getMetadata(presentation.metadata.FrontUrl) ?? window.location.origin
+      const token = getCookieToken(); 
 
       try {
-        const res = await fetch(concatLink(frontUrl, `/files?file=${name}`))
+        const res = await fetch(concatLink(frontUrl, `/files?file=${name}&token=${token}`))
 
         if (res.ok) {
           const blob = await res.blob()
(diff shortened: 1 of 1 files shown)


aonnikov (changes requested):
I don't understand how it is suppose to work. The changes affect only provider that is used to fetch document content from minio directly. But even in this case these changes are not sufficient. The problem occurs not only in text editor, but with any images requested from `/front`.

CodeQueenTosin:
> I don't understand how it is suppose to work. The changes affect only provider that is used to fetch document content from minio directly. But even in this case these changes are not sufficient. The problem occurs not only in text editor, but with any images requested from `/front`.

Okay thanks, changes will be extended for all cases.

haiodo:
Hi @CodeQueenTosin 

I will change schema for resource retrieval in near future and will include workspace id into requests, 
so it will be possible to properly define cookies.

So let's leave current solution for now, until work with resource retrieval will be finished.

Thanks for your contribution.

PR #5246, opened 2024-04-08
Fixed typo (Planing -> Planning)

tanishgupta-git:
Press the search icon at right and scroll to the bottom. In the commands list, there is a typo.

- Planing —> Planning


nickywilliams114 (approved):
Just curious if GotoTimePlaning and GotoTimeTeamPlaning also need to be changed to Planning

tanishgupta-git:
yaah we can also update them if we want

PR #5256, opened 2024-04-08
UBERF-NA-App: Add number increment buttons & ADA aria-labels on Button comp

LoganEllsworth:
* Added up / down buttons on `EditBox` component for `type="number"` fields, increments value by 1.
* Added `aria-label` to the `Button` component as prop `aria` for ADA compliance on buttons. Only added `aria` prop to the above mentioned buttons currently, as well as raw aria-labels to `type="password"` and text input fields in `EditBox`.

Screenshot from 2024-04-08 15-09-41


haiodo:
Closing as outdated.

Issue #5140, opened 2024-04-02
workSpaceUrl is unique


iraj720:
https://github.com/hcengineering/platform/blob/c8a58d542805be956067a211dd7616bf53d15946/server/account/src/index.ts#L753-L760

workspace is unique and workspaceUrl contains it.
so workspaceUrl is also unique we can create an unique index for it for simplification

there are so many string manipulation which is not needed in this function and so in the database

we can easily put w-{md5(email, workspaceName)} as the workspace without any for loop and redundant db calls bcz it will be always unique and also put workspaceUrl same as workspace to decrease the confusion in the code and db and user can update wsUrl whenever he/she wants

Issue #5139, opened 2024-04-02
create workspace confusion


iraj720:
a user can create multiple workspaces with the same name and it will be confusing for him/her
so easily we can reject that by :

db.collection('workspace').createIndex({ workspaceName: 1, createdBy: 1}, { unique: true })

but it needs some cleaning in db call functions
also we can show the creator of a workspace in select workspace page to fix this confusion

Issue #5138, opened 2024-04-02
search promise is locking the proccess


iraj720:
https://github.com/hcengineering/platform/blob/c8a58d542805be956067a211dd7616bf53d15946/server/account/src/index.ts#L795

this promise is used few lines below it and we are awaiting twice on it

if any error happens in the generateWorkspaceRecord function it will not assign it to searchPromise and over and over again it will happen untill you restart the pod and that global variable will be free

so further requests will always fail no matter they are correct or not bcz it will run the former promise not the new one

<sub>View in Huly <a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjBiZGQyZGU1MGZjYjI5NThiMDVhYTEiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.KdeK7SVcG_xXV9uFpdZWzuNjJYtG52lTdPXWu4YTyJE">UBERF-6274</a></sub>

PR #5277, opened 2024-04-09
fix creating workspace

iraj720:
https://github.com/hcengineering/platform/issues/5140 
https://github.com/hcengineering/platform/issues/5139 
https://github.com/hcengineering/platform/issues/5138

this is a fix for these 3 issues just tested it. its working correct.

also we can add a layer for interpreting mongodb error codes


iraj720:
@haiodo 
this is the correct one

PR #5297, opened 2024-04-10
UBERF-6409: Turn notifications filter into an "Unread" toggle

riltonfranzonee:


aonnikov:
To me this toggle has several problems:
1. Naming. It probably should be reversed and named `Read` to indicate that read messages are shown when the toggle is on. Or it should have two labels, `All` to the left and `Unread` to the right.
2. For small screens the toggle is worse option than button with dropdown menu, it consumes more space so I have to increase width of the inbox.

kristina-fefelova:
@aonnikov 
1.  I think that with naming we can keep like this, since two labels will take up too much space.
Also in slack there is the same control and this may be familiar to users. What do you think?
Screenshot from 2024-04-11 10-33-21
Screenshot from 2024-04-11 10-36-00

2. Perhaps the toggle and text should be made a little smaller and/or turned into button (as was before) when panel has small width

plugins/notification-resources/src/components/Filter.svelte
```diff
@@ -13,36 +13,27 @@
...
-    },
-    {
-      id: 'unread',
-      isSelected: filter === 'unread',
-      label: notification.string.Unread
-    }
-  ]
+<div class="toggle-wrapper">
+  <ToggleWithLabel 
+    label={"Unread"}
```
kristina-fefelova: Please use translated strings

plugins/notification-resources/src/components/Filter.svelte
```diff
@@ -13,36 +13,27 @@
 // limitations under the License.
 -->
 <script lang="ts">
-  import { IconFilter, SelectPopup, eventToHTMLElement, showPopup, ButtonIcon } from '@hcengineering/ui'
-  import notification from '../plugin'
-
-  export let filter: 'all' | 'read' | 'unread' = 'all'
+  import ToggleWithLabel from "@hcengineering/ui/src/components/ToggleWithLabel.svelte"
+  export let filter: 'all' | 'unread' = 'all'
```
kristina-fefelova: Can we use type  `InboxNotificationsFilter ` ?

plugins/notification-resources/src/components/People.svelte
```diff
@@ -25,7 +25,7 @@
 
   import PeopleNotificationView from './PeopleNotificationsView.svelte'
 
-  export let filter: 'all' | 'read' | 'unread' = 'all'
+  export let filter: 'all' | 'unread' = 'all'
```
kristina-fefelova: Can we use type  `InboxNotificationsFilter ` ?
Please check all places

kristina-fefelova:
thanks a lot  for your contribution, but we already have that changes with another pull request

PR #5449, opened 2024-04-24
FIX: Optimize Emoji load and highlight Symbols icon on scroll

nathangemis:
* Change the way that we generate emojis (loop function)
* When we scroll the emojis, Symbols icon category was not highlighted, Now we move to next category when we are at 90% scroll of the precedent category div size

### DIFF
--- a/packages/ui/src/components/EmojiPopup.svelte
+++ b/packages/ui/src/components/EmojiPopup.svelte
@@ -105,7 +105,7 @@
 
   let currentCategory = categories[0]
 
-  function getEmojis(startCode: number, endCode: number, postfix?: number[]): Array<string | undefined> {
+  function getEmojis(startCode: number, endCode: number, postfix?: number[]): Array<string> {
     const result = []
     for (let i = startCode; i <= endCode; i++) {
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/EmojiPopup.svelte
```diff
@@ -107,12 +105,16 @@
 
   let currentCategory = categories[0]
 
-  function getEmojis (startCode: number, endCode: number, postfix?: number[]): Array<string | undefined> {
-    return [...Array(endCode - startCode + 1).keys()].map((v) => {
-      const str = postfix ? fromCodePoint(v + startCode, ...postfix) : fromCodePoint(v + startCode)
-      if ([...str.matchAll(regex)].length > 0) return str
-      return undefined
-    })
+  function getEmojis(startCode: number, endCode: number, postfix?: number[]): Array<string | undefined> {
```
aonnikov: The returned value should be changed because `undefined` is not possible now.

haiodo:
Closing since outdated.

kristina-fefelova:
Closing as outdated.

PR #5524, opened 2024-05-06
Fix locator for flaky test

JasminMus:

fix locator for flaky test

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjM4ZjNhZWM2MTdkNTcxZjcwZTM5MDYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.WLoLu_0yUophUgNh06yr4HpZzDXasmgecVpkJOHH0yc">Huly&reg;: <b>UBERF-6812</b></a></sub>

### DIFF
--- a/tests/sanity/tests/model/recruiting/talents-page.ts
+++ b/tests/sanity/tests/model/recruiting/talents-page.ts
@@ -59,6 +59,8 @@ export class TalentsPage extends CommonRecruitingPage {
   openOtherSkills = (): Locator => this.page.getByText('Other')
   skillsLink = (): Locator => this.page.locator('text=Skills')
   newSkillButton = (): Locator => this.page.locator('button:has-text("Skill")')
+  emailContact = (): Locator =>
+    this.page.locator('div[class^="popupPanel-body"] div.horizontal button[id="gmail:string:Email"]')
 
   async clickAddApplication (): Promise<void> {
     await this.addApplicationButton().click()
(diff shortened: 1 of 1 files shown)


tests/sanity/tests/model/recruiting/talents-page.ts
```diff
@@ -59,6 +59,8 @@ export class TalentsPage extends CommonRecruitingPage {
   openOtherSkills = (): Locator => this.page.getByText('Other')
   skillsLink = (): Locator => this.page.locator('text=Skills')
   newSkillButton = (): Locator => this.page.locator('button:has-text("Skill")')
+  emailContact = (): Locator =>
+    this.page.locator('div[class^="popupPanel-body"] div.horizontal button[id="gmail:string:Email"]')
```
ThetaDR: `.horizontal` is generic class in our Scroller component, so I think it would be better to use `.popupPanel-body__header` (or another class that is used only in the header) instead of `.horizontal`. It'll make sure that we are trying to check email in the header and not in any other place in the panel.

ThetaDR (changes requested):

tests/sanity/tests/model/recruiting/talents-page.ts
JasminMus: I don't think it should be a problem, since horizontal is looking that exact same area.
image

but since you know components a lot better than I do, I will make an update

haiodo (approved):

ThetaDR (approved):

PR #5532, opened 2024-05-07
Add livness check for login and workspace

JasminMus:

Add two spec of tests that will run against the https://front.hc.engineering/  to ensure that the production is working.

NOTE:
These tests should not run on your PR build because they are excluded in the package.json script.

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjM5ZmRlZmM2MTdkNTcxZjcwZTNlMDQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.Iakj8ZqacmMSH0MvhjvU-O6Ig1w96vv1pM52Fo_p_fU">Huly&reg;: <b>UBERF-6830</b></a></sub>


tests/sanity/tests/liveness/login.spec.ts
```diff
@@ -0,0 +1,55 @@
+import { test } from '@playwright/test'
```
haiodo: @JasminMus Let's move Liveness checks for another folder or probable event repository. Since we plan to have multiple demo workspaces, and it will be hard to maintain in this repository. Probable we could for now move it into our second repository with a real deployment. Thoughts?

JasminMus: @haiodo sure we can move this to another repository, I left it here because I didn't wanna duplicate things such as locators, playwright setup etc, we would just need a new workflow that will do the install, not even use docker and hit the production.
But I am fine with moving it to another repository, I will just need a bit of guidance to where.

tests/sanity/.env
```diff
@@ -4,4 +4,6 @@ PLATFORM_USER='user1'
 PLATFORM_USER_SECOND='user2'
 PLATFORM_TOKEN='eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJlbWFpbCI6InVzZXIxIiwid29ya3NwYWNlIjoic2FuaXR5LXdzIn0.hfUCqePHO-WNps2by4B-CYGKIpDpLG0WVCUUtU-SVI4'
 SETTING=storage.json
-SETTING_SECOND=storageSecond.json
\ No newline at end of file
+SETTING_SECOND=storageSecond.json
+
+LIVENESS_URI='https://front.hc.engineering'
```
aonnikov: I don't see why we need a separate URI for these tests. Liveness tests are just a subset of all tests that run against some deployment. 
Ideally, we should be able to run them against local setup or CI as well. In order to be able to do this we need only two things:
1. Properly tag these tests so we can filter / filter them out
2. Provide separate tests configuration

An alternative would be to create separate folder with tests.

And we need to esnure that is not possible to accidentally run these tests on production environment.

JasminMus: this will be moved to uberflow repo, the grep tag is already there @livness which won't run any tests on CI that are marked with @livness... 
Basically the idea is exclude these types of tests from a regular run, but to create a separate Cron workflow which will check if the production is working or not every lets say 4 hours or once a day.

JasminMus: But I will just modify this PR a bit so we have these tests cover our current tests as well

JasminMus: @aonnikov ok the PR is now set for regular CI check

haiodo (approved):

PR #5577, opened 2024-05-10
Statuses fix

ThetaDR:
* Migration for statuses
* fix `$in` query to model

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjNlMjRjMzI1NjAzZjg0ZDA0ZWM0ZTMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.YZsKJFUHjy0ai9fSrYRRNPIL55zyOxFjA7cd71Q6VmE">Huly&reg;: <b>UBERF-6907</b></a></sub>

### DIFF
--- a/models/tracker/src/migration.ts
+++ b/models/tracker/src/migration.ts
@@ -504,6 +504,35 @@ async function restoreTaskTypes (client: MigrationClient): Promise<void> {
   }
 }
 
+async function migrateIssueStatuses (client: MigrationClient): Promise<void> {
+  await client.update(
+    DOMAIN_TX,
+    {
+      objectClass: task.class.TaskType,
+      'attributes.ofClass': tracker.class.Issue,
+      'attributes.statusClass': core.class.Status
+    },
+    {
+      $set: {
+        'attributes.statusClass': tracker.class.IssueStatus
+      }
+    }
+  )
+
+  await client.update(
+    DOMAIN_STATUS,
+    {
+      _class: core.class.Status,
+      ofAttribute: tracker.attribute.IssueStatus
+    },
+    {
+      $set: {
+        _class: tracker.class.IssueStatus
+      }
+    }
+  )
+}
+
 export const trackerOperation: MigrateOperation = {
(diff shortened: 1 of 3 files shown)


packages/core/src/memdb.ts
```diff
@@ -58,7 +58,7 @@ export abstract class MemDb extends TxProcessor implements Storage {
       const obj = this.objectById.get(query._id) as T
       if (obj !== undefined && this.hierarchy.isDerived(obj._class, _class)) result.push(obj)
     } else if (query._id?.$in !== undefined) {
-      const ids = query._id.$in
+      const ids = [...new Set(query._id.$in)]
```
BykhovDenis, line 61: You don't need to spread set, you can make for on Set

BykhovDenis (approved):

haiodo (approved):

models/tracker/src/migration.ts
```diff
@@ -504,6 +504,35 @@ async function restoreTaskTypes (client: MigrationClient): Promise<void> {
...
+    },
+    {
+      $set: {
+        'attributes.statusClass': tracker.class.IssueStatus
+      }
+    }
+  )
+
+  await client.update(
+    DOMAIN_STATUS,
```
lexiv0re, line 523: Statuses are in the model now. `DOMAIN_STATUS` was kept just in case if migration goes wrong. We'll drop it soon.

haiodo: It is 239 before statuses are in model actually.

lexiv0re: Ah, ok. Should be fine then.

PR #5617, opened 2024-05-19
Table options

andrewkuryan:
* Add an options button with Delete and Duplicate options inside row and column handlers when selected:

Снимок экрана 2024-05-21 в 01 09 41

Снимок экрана 2024-05-21 в 01 09 59

Multiple deletions and duplications are supported as well:

Снимок экрана 2024-05-21 в 01 10 23

* Fix a bug where the [`tableCellDecorationPlugin`](https://github.com/hcengineering/platform/blob/main/packages/text-editor/src/components/extension/table/tableCell.ts#L40) function was called twice, leading to the double creation of the decorators list. Proof:

image

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjRhNzM5NjAwNGU2ZmM4NTdkZjcyNzciLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.JZuuquPtquBIo2lcHG6OY8x-vXsx9krOcAZwC4dCO70">Huly&reg;: <b>UBERF-6987</b></a></sub>


packages/text-editor/src/components/extension/table/tableCell.ts
```diff
@@ -28,7 +28,7 @@ import { rowHandlerDecoration } from './decorations/rowHandlerDecoration'
 
 export const TableCell = TiptapTableCell.extend({
   addProseMirrorPlugins () {
-    return [...(this.parent?.() ?? []), tableCellDecorationPlugin(this.editor)]
+    return [tableCellDecorationPlugin(this.editor)]
```
aonnikov, line 31: Was it the root cause of duplicated elements? I don't remember exactly, but I think that we need parent plugins because there is some useful prosemirror's stuff.

packages/text-editor/src/components/extension/table/decorations/optionsButton.ts
```diff
@@ -0,0 +1,30 @@
...
+import { optionsSvg } from './icons'
+
+export interface OptionItem {
+  id: string
+  icon: AnySvelteComponent
+  label: string
+  action: () => void
+}
+
+export function createOptionsButton (editor: Editor, options: OptionItem[], className?: string): HTMLElement {
```
aonnikov: This won't work for us. Our tables already have massive decorations and cause problems with displaying because we have to keep about 2rem padding on top and left sides of the table.
I think we can use either col/row drag indicator or entire header. But in this case we will need to properly distinguish between click and drag.

aonnikov (commented):
Thanks for the very valuable contribution. Please look at couple of comments from me.

packages/text-editor/src/components/extension/table/tableCell.ts
andrewkuryan: @aonnikov If you add logging to the `addProseMirrorPlugins` method, you will see that it is called twice. The first time, the `this.parent?.()` call returns `undefined`, and the second time, it returns an array consisting of the `table-cell-decoration-plugin` instance, which is the result of the first call:

Снимок экрана 2024-05-20 в 21 59 15

This behavior is consistent with the source code of the `@tiptap` module, as the `extend` function is also called twice (the second time inside `configure`). During the first call, the `parent` field is assigned `this`:

image

packages/text-editor/src/components/extension/table/decorations/optionsButton.ts
andrewkuryan: @aonnikov Moved the options button inside indicators. The button becomes available only when a row or column is selected. The options menu opens on the `click` event, not on `mousedown`, so it doesn't interfere with dragging. Updated the PR description to match

aonnikov (dismissed):

aonnikov:
@andrewkuryan could you please resolve formatting issues?

aonnikov (approved):

PR #5622, opened 2024-05-20
Feat: add breakpoints to issues

TathevikJS:
### Pull Request Description: Adding Breakpoint Feature to Issues

**Summary:**
This pull request introduces a new feature to our issue tracking system: the ability to add, set, and edit a difficulty breakpoint for each issue. The difficulty level ranges from 0 to 5. This enhancement aims to help organize tasks more effectively among employees by dividing tasks based on their difficulty levels.

**Key Features:**

1. **Add Difficulty Breakpoint:**
   * Users can now specify a difficulty level for each issue when creating it.
   * Difficulty levels range from 0 (least difficult) to 5 (most difficult).
2. **Edit Difficulty Breakpoint:**
   * Users have the ability to edit the difficulty level of an issue after it has been created.
   * This allows for flexibility in case the assessment of the task difficulty changes over time.
3. **Task Organization and Assignment:**
   * Helps in the distribution of tasks among employees by matching the difficulty of tasks with the appropriate skill levels.
   * Ensures a balanced workload by assigning tasks of varying difficulties evenly across all employees.

**Screenshots**
Screen Shot 2024-05-20 at 17 53 11

Screen Shot 2024-05-20 at 17 53 18

Screen Shot 2024-05-20 at 17 53 24

Screen Shot 2024-05-20 at 17 53 49

Screen Shot 2024-05-20 at 17 53 57


aonnikov:
@TathevikJS thanks for the contribution.
I understand the idea and that someone may need breakpoints in their everyday use, but it is not something that everyone needs. One can come up with handful attributes that might be useful, but we won't be able to add them all into the base implementation.

Instead, in the platform we already provide an extension point that allows to add custom attributes to any class in the system. Say, almost the same result you can achieve by just configuration, with zero programming.

Example:
Screenshot 2024-05-21 at 11 10 47

I say "almost" because the implementation is far from being ideal and may require some enhancements such as:
- ability to add icon for the field (not necesaary, though)
- ability to show custom fields in the creation dialog
- enhance custom fields support in filters

Any contributions that enhance user experience for the custom attributes  would be welcome.

PR #5624, opened 2024-05-20
Feat add issue dependencies

TathevikJS:
**Summary:**
This pull request introduces a new feature to our issue tracking system: the ability to set dependencies between issues. This feature ensures that an issue cannot be marked as complete until all its dependency is resolved, enhancing task management and project workflow.

Key Features:

1. Add Issue Dependencies:

Users can set dependency for an issue, linking it to other issues that must be resolved first.
This can be done during issue creation and also by editing an existing issue.

1. Dependency Validation:

An issue cannot be marked as complete if dependency is still open.

1. Dependency Display:

The user interface has been updated to show dependency for each issue, providing clear visibility of task prerequisites.


aonnikov:
@TathevikJS first of all, thanks for the contribution!

We already have something similar to the issue dependencies, named "Related issue", maybe it is better to look into evolving this feature rather than adding another type of dependencies between issues. I also think that dependencies between issues should allow to configure type of dependencies, like issue links in Jira so that is easier to understand what depends on what and why. And one more thing, it is absolutely necessary to be able to configure this dependencies for existing issues.

PR #5661, opened 2024-05-24
UBERF-7011: Switch to Ref<Blob>

haiodo:
* Use direct download link for s3 if allowed.
* Improve local preview, avif, webp formats added.

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjUwN2E4N2QxMmEwOTk1ZmI4MzA0ZDYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.MRa6K57KOmMMbh63Md_VreeonbcW4LE_s2BAUKQU0VY">Huly&reg;: <b>UBERF-7057</b></a></sub>


packages/text/src/nodes/image.ts
aonnikov, line 1: we also have a file extension that allows to attach inline files, should we modify it in the same way?

aonnikov (approved):

BykhovDenis (approved):

PR #5715, opened 2024-06-03
Security and performance improvements on authentication

sgmakgg:
Hi there!

* pbkdf2Sync from 'node:crypto' was changed to callback version - pbkdf2
* Buffer.compare() replaced  with timingSafeEqual() from 'node:crypto'
  [*The crypto.timingSafeEqual() function is used to determine whether two variables are equal without exposing timing information that may allow an attacker to guess one of the values. A constant-time algorithm underpins it.](https://nodejs.org/api/crypto.html#cryptotimingsafeequala-b)
* hashWithSalt() function was promisified and renamed to 'hashPassword'
* verifyPassword() function was promisified
* try...catch blocks added in places where functions are called

Signed-off-by: Makarov Sergey [sgmakgg@gmail.com](mailto:sgmakgg@gmail.com)


haiodo:
@lexiv0re Please check if this is valid for new global account and use proposed changes, thanks.

PR #5717, opened 2024-06-03
UBERF-4725 Migrate collaborative content

aonnikov:
<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NjVkZGZhMTI4ZDQ2ZTc0NDQ3Y2NhMjUiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.XB9cCxX8pcXX8Fg3GQlyZv_Gc_l1PTHQ5i0Shp_pwLY">Huly&reg;: <b>UBERF-7136</b></a></sub>


server-plugins/activity-resources/src/utils.ts
```diff
@@ -315,6 +323,15 @@ export async function getTxAttributesUpdates (
       }
     }
 
+    // we don't want to show collaborative documents in activity
+    // instead we show their content as Markup
+    // TODO this should be generalized via activity extension
+    if (attrClass === core.class.TypeCollaborativeDoc) {
```
kristina-fefelova: are you planning to do it  in this pull request or is it a note for the future?

aonnikov: let's decide on review how we handle this
I can do in this PR but It would be also good to have async trigger because file loading may affect the trigger performance.

kristina-fefelova: we can make activity generation as asynс trigger, I thought about that too

I would prefer to do it async in a separate pull request, and merge this pull request for now if it works
You can assign this task to me

aonnikov: I figured out that I need to do a bit more changes to support mixins here. Will be done tomorrow.

BykhovDenis (approved):

PR #5825, opened 2024-06-14
Added Emoji Support in Inline and Side Menu in Collaborative Text Editor

namanvats:
What Does this PR Do?
It adds support in Collaborative Text Editor to add Emjois using inline command `/` or side menu

Earlier only 4 options were there in the menu :

* Image
* Table
* Code block
* Separator Line

Added a new option : Emoji , later can be extended easily to support Emoji popup

Screenshot
Inline Command
Screenshot 2024-06-14 at 6 13 08 PM

Side Menu
Screenshot 2024-06-14 at 6 13 19 PM

### DIFF
--- a/packages/text-editor/src/components/CollaborativeTextEditor.svelte
+++ b/packages/text-editor/src/components/CollaborativeTextEditor.svelte
@@ -32,6 +32,7 @@
     themeStore
   } from '@hcengineering/ui'
   import view from '@hcengineering/view'
+  import activity from '../../../../plugins/activity'
   import { AnyExtension, Editor, FocusPosition, mergeAttributes } from '@tiptap/core'
   import Collaboration, { isChangeOrigin } from '@tiptap/extension-collaboration'
   import CollaborationCursor from '@tiptap/extension-collaboration-cursor'
@@ -202,6 +203,9 @@
     insertSeparatorLine: () => {
       editor?.commands.setHorizontalRule()
     },
+    insertEmoji: () => {
+      editor?.commands.setHorizontalRule()
+    },
     focus: () => {
       focus()
     }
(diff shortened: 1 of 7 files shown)


packages/text-editor/src/components/CollaborativeTextEditor.svelte
```diff
@@ -202,6 +203,9 @@
     insertSeparatorLine: () => {
       editor?.commands.setHorizontalRule()
     },
+    insertEmoji: () => {
+      editor?.commands.setHorizontalRule()
```
aonnikov, line 207: Why do you call `setHorizontalRule` here?

namanvats: Basically, it was built to showcase that the functionality can be extended to add emoji support so adding it as place holder for review purposes if required, it can be easily extended to support emoji popups. The screenshot shows that emoji command can be added to this document and popups too can be called in place of setHorizontalRule which is a placeholder command for now

aonnikov: Even though we implemented it in this way, I don't think that's the right way of extending this menu.
1. Currently this menu is supposed to insert only block elements. An emoji is an inline element and I don't see much value in inserting an emoji in the beginning of the line.
2. The way we use to extend the menu requires a lot of boilerplate. I believe we need to have a single list of commands that will be used in both bubble menu, inline menu and this side menu.

namanvats: On # 1 : Agree with you but many users are using it to start tech docs or quick start guides that's why added it in inline commands as well.

On # 2 : I followed the existing implementation flow as it was done and tried to extend this support but based upon your comment looks like there might be a need to implement it differently let me know if I can try that way.

kristina-fefelova:
Closing as outdated.

PR #5912, opened 2024-06-24
EQMS-1033 Filter out readonly projects when creating new controlled document

aonnikov:
<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2Njc5OTUxYjNlOGFhOTFjZTY2YWViOGYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.Y1Eptu1WV1hShu7PONmgRohbIrVmiVuYnGKRNK8PjBo">Huly&reg;: <b>UBERF-7392</b></a></sub>

### DIFF
--- a/plugins/controlled-documents-resources/src/utils.ts
+++ b/plugins/controlled-documents-resources/src/utils.ts
@@ -15,6 +15,7 @@ import core, {
   type AttachedData,
   type Class,
   type Doc,
+  type DocumentQuery,
   type Hierarchy,
   type Ref,
   type Tx,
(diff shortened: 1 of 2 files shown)


plugins/controlled-documents-resources/src/utils.ts
```diff
@@ -714,7 +721,7 @@ export async function getLatestProjectId (spaceRef: Ref<DocumentSpace>): Promise
 
   if (space !== undefined) {
     if (space.$lookup?.type?.projects ?? false) {
-      const project = await getLatestProject(spaceRef)
+      const project = await getLatestProject(spaceRef, editableOnly)
```
lexiv0re: Should condition be inverted?

aonnikov: no, that's just wrong name :)

BykhovDenis (approved):

plugins/controlled-documents-resources/src/utils.ts
lexiv0re: Ok, but why do we include readonly in document space presenter then? I thought we wanted to only show not readonly versions there? Or it's in another component?

lexiv0re: discussed

lexiv0re (approved):

Issue #5854, opened 2024-06-18
Typo in apply.js


l2dy:
### Description of the issue
https://github.com/hcengineering/platform/blob/1e9fea356ea31c0095ffe7e07996ab638f628a16/templates/apply.js#L129-L131

### Possible solutions
`.forEach()` discards the return value anyway, so I assume `[]` should be replaced with `{}`.

PR #5952, opened 2024-06-28
Fixed a bracket typo in an IF statement - templates/apply.js:(129-131)

Nikoraii:
### Fixed a bracket typo in apply.js
**[ISSUE: 5854](https://github.com/hcengineering/platform/issues/5854)**

In an `if` statement `[]` was used instead of `{}`, which was inside a `.forEach()` loop. As `.forEach()` discards the return value and always returns `undefined` there is no need for `[]`.


Gnarus-G:
@Nikoraii Friendly tip: you need to sign off your commit. The [DCO check](https://github.com/hcengineering/platform/pull/5952/checks?check_run_id=26827531861) failed. Here is the documentation on signing commits, https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Nikoraii:
Moved to new a branch so I could properly set up signing.

Thank you @Gnarus-G !

PR #5953, opened 2024-06-28
feat: make inline commands searchable

akevinge:
Similar behavior can be found in Notion and Obsidian. Just a quality of life bump :)

image


packages/text-editor/src/components/InlineCommandsList.svelte
```diff
@@ -33,6 +35,15 @@
   let dummyPopup: PopupResult
   let menuPopup: SelectPopup
 
+  let fuse: Fuse<any>
+  let sortedItems: any[] = []
+  $: fuse = new Fuse(items, { shouldSort: true, keys: ['id'] })
```
aonnikov, line 40: can we get rid of dependency on fuse.js and just check that item name contains or starts with query string?

haiodo:
Closing since outdated.

kristina-fefelova:
Closing as outdated.

PR #5985, opened 2024-07-02
Fix bad link formatting

kujaomega:
This PR removes link formatting from tiptap fixing bad link formatting:
Screenshot from 2024-07-02 21-14-52
The changes adds links to URLs only.

### DIFF
--- a/packages/presentation/src/components/markup/NodeContent.svelte
+++ b/packages/presentation/src/components/markup/NodeContent.svelte
@@ -40,6 +40,57 @@
   function toNumber (value: string | number | undefined): number | undefined {
     return value !== undefined ? (typeof value === 'string' ? parseInt(value) : value) : undefined
   }
+
+
+  function getIndicesOf(searchStr: string, str: string): [string] {
+    const searchStrLen = searchStr.length
+    if (searchStrLen == 0) {
+      return []
+    }
+    let startIndex = 0, index, indices = []
+    while ((index = str.indexOf(searchStr, startIndex)) > -1) {
+      const endIndex = index + searchStrLen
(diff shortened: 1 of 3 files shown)


packages/text/src/kits/default-kit.ts
```diff
@@ -54,11 +54,7 @@ export const DefaultKit = Extension.create<DefaultKitOptions>({
       Highlight.configure({
         multicolor: false
       }),
-      Typography.configure({}),
-      Link.configure({
```
aonnikov, line 58: You cannot just remove this extension:
1. Link button in toolbar will be broken
2. Existing content that contains links will be broken

packages/presentation/src/components/markup/NodeContent.svelte
```diff
@@ -49,7 +100,9 @@
   {#if node.type === MarkupNodeType.doc}
     <MarkupNodes {nodes} {preview} />
   {:else if node.type === MarkupNodeType.text}
-    {node.text}
+    <p>
```
aonnikov, line 103: adding additional paragraphs may affect layout

packages/presentation/src/components/markup/NodeContent.svelte
```diff
@@ -49,7 +100,9 @@
...
+      {@html urlify(node.text)}
```
aonnikov, line 104: `@html` is potentially dangerous , we are trying not to use it

aonnikov (changes requested):

packages/presentation/src/components/markup/NodeContent.svelte
kujaomega: urlify function, encodes strings to html encoded characters except added links

PR #6097, opened 2024-07-18
Add login and signup by code

kristina-fefelova:
<https://github.com/user-attachments/assets/cc5a7263-c21b-4cb3-9b16-021697b71e64>

Email example:
Screenshot 2024-07-25 at 21 13 45

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2Njk5NzYxMzEwNmQ1ZmVkZWI1ZWMwYWMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.ConSMjmU_1tnKCX9vIHSDr96nzjgWRIH8eN6xNQiDww">Huly&reg;: <b>UBERF-7625</b></a></sub>


BykhovDenis:
I don't think that we need it for login, only for sign up

kristina-fefelova:
> I don't think that we need it for login, only for sign up

Hm, maybe I misunderstood something. I thought the idea was to login without a password

lexiv0re (dismissed):

server/account/src/operations.ts
```diff
@@ -387,6 +432,117 @@ export async function login (
...
+      .limit(1)
+      .toArray()
+  )[0]
+  const isValid = otpData !== undefined && otpData.expires > now
+
+  if (isValid) {
+    return { sent: true, expires: otpData.expires }
+  }
+
+  const timeToLive = 30 * 1000
```
aonnikov: 30 sec is too small interval for otp code, I recommend to increase it to 1-5 minutes. Even better to make it configurable via env variable.

server/account/src/operations.ts
```diff
@@ -387,6 +432,117 @@ export async function login (
...
+  if (!isValid) {
+    throw new PlatformError(new Status(Severity.ERROR, platform.status.InvalidOtp, {}))
+  }
+
+  try {
+    const info = toAccountInfo(account)
+    const result = {
+      endpoint: '',
+      email,
+      confirmed: info.confirmed ?? true,
```
aonnikov: why do we need confirmation info here? since we use email channel for otp codes, hence if user entered a valid code, then the email can be considered as confirmed.

server/account/src/operations.ts
kristina-fefelova: Moved to config, by default 1 minute

server/account/src/operations.ts
kristina-fefelova: Marked email as confirmed on otp login

server/account/src/operations.ts
kristina-fefelova: Also extracted retry_delay to config

aonnikov:
> > I don't think that we need it for login, only for sign up
> 
> Hm, maybe I misunderstood something. I thought the idea was to login without a password

Actually both 😄

BykhovDenis (approved):

PR #6185, opened 2024-07-30
UBERF-5395: Have loading option for menu and show only intersecting statuses there

ThetaDR:
Show loading indicator for menu. Show only intersecting statuses instead of empty component.


haiodo:
Closing as outdated.

PR #6264, opened 2024-08-05
Fix Indexer use of $ne

haiodo:
Fix Indexer use of $ne

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NmIxM2I0OWU1NTRjZjllNTY4ZmEzZDMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.5xX5DqUTtI7PiNicXch0qOR6bfsOZCLeRkclBpbjhCw">Huly&reg;: <b>UBERF-7815</b></a></sub>

### DIFF
--- a/server/core/src/fulltext.ts
+++ b/server/core/src/fulltext.ts
@@ -14,9 +14,11 @@
 // limitations under the License.
 //
 
+import { Analytics } from '@hcengineering/analytics'
 import core, {
   type AttachedDoc,
   type Class,
+  type Collection,
   type Doc,
   type DocIndexState,
   type DocumentQuery,
@@ -47,7 +49,6 @@ import { createStateDoc } from './indexer/utils'
 import { getScoringConfig, mapSearchResultDoc } from './mapper'
 import { type StorageAdapter } from './storage'
 import type { FullTextAdapter, IndexedDoc, ServerStorage, WithFind } from './types'
-import { Analytics } from '@hcengineering/analytics'
 
 /**
  * @public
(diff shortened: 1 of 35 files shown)


server/core/src/fulltext.ts
```diff
@@ -143,7 +145,7 @@ export class FullTextIndex implements WithFind {
 
     const ids: Set<Ref<Doc>> = new Set<Ref<Doc>>()
     const baseClass = this.hierarchy.getBaseClass(_class)
-    let classes = this.hierarchy.getDescendants(baseClass)
+    let classes = this.hierarchy.getDescendants(baseClass).filter((it) => !this.hierarchy.isMixin(it))
```
ThetaDR, line 148: Why filter for mixins is needed here?

haiodo: We could not have _class in DocIndexState to be one of mixins so they should not go into elastic, so we should skip them on search.

haiodo: Probable we need to add getBaseClass there, is a good point

ThetaDR: My concern is that it may filter out recruit.mixin.Candidate and skip it in the search. I might be wrong, though.

haiodo: Nope, it work fine with candidate, because of this.hierarchy.getBaseClass(_class)

ThetaDR (approved):

BykhovDenis (approved):

PR #6340, opened 2024-08-15
UBERF-7690: Trigger improvements

haiodo:
UBERF-7690: Trigger improvements

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NmJkYzk5ZDEwMDhmZDk4YzJkZTJmZjUiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIiwicHJvZHVjdElkIjoiIn0.m6wn0cPIXsIZjXyX3dk-a7L9f4DIavySQj2vpmsvMJI">Huly&reg;: <b>UBERF-7891</b></a></sub>


server-plugins/chunter-resources/src/index.ts
```diff
@@ -286,15 +269,37 @@ async function OnThreadMessageDeleted (tx: Tx, control: TriggerControl): Promise
...
+  ) {
+    res.push(
+      ...(await control.ctx.with(
+        'OnThreadMessageCreated',
+        {},
+        async (ctx) => await OnThreadMessageCreated(tx, control)
+      ))
+    )
+  }
+  if (control.hierarchy.isDerived(actualTx.objectClass, chunter.class.ThreadMessage)) {
```
kristina-fefelova: should add `actualTx._class === core.class.TxRemoveDoc` ?

server-plugins/notification-resources/src/index.ts
```diff
@@ -801,24 +810,24 @@ async function updateContextsTimestamp (
...
   if (contexts.length === 0) return
   if (unsubscribe.length === 0) return
 
-  const unsubscribeAccounts = await control.modelDb.findAll(contact.class.PersonAccount, {
-    _id: { $in: unsubscribe }
-  })
   const res: Tx[] = []
 
   for (const context of contexts) {
-    const account = unsubscribeAccounts.find(({ _id }) => _id === context.user)
```
kristina-fefelova, line 823: We should add something like to remove only correct contexts: 
```
if(!unsubscribe.includes(context.user)) {
continue;
}
```

server-plugins/notification-resources/src/index.ts
```diff
@@ -862,23 +871,31 @@ export async function createCollabDocInfo (
...
     }
 
     return res
   }
 
-  const notifyContexts = await control.findAllCtx(ctx, notification.class.DocNotifyContext, { objectId: object._id })
-  await removeContexts(notifyContexts, unsubscribe, control)
-  await updateContextsTimestamp(notifyContexts, originTx.modifiedOn, control, originTx.modifiedBy)
+  let notifyContexts: DocNotifyContext[] = []
+  if (!(object._id === tx.objectId && tx._class === core.class.TxCreateDoc)) {
```
kristina-fefelova, line 886: Why we should skip this case?

server-plugins/love-resources/src/index.ts
```diff
@@ -313,7 +321,8 @@ export async function OnInvite (tx: Tx, control: TriggerControl): Promise<Tx[]>
               name: formatName(from.name, control.branding?.lastNameFirst)
             })
             : await translate(love.string.InivitingLabel, {})
-        await createPushNotification(control, userAcc._id, title, body, invite._id, from, path)
+        // TODO: Select a proper user
+        await createPushNotification(control, userAcc[0]._id, title, body, invite._id, from, path)
```
kristina-fefelova, line 325: In the current implementation of accounts, shouldn't we send notification to all user accounts?

server-plugins/notification-resources/src/utils.ts
```diff
@@ -474,33 +472,50 @@ export async function getUsersInfo (
...
-      async () => await control.findAll(contact.class.Person, { _id: { $in: personIds } })
-    )
+  const persons: IdMap<WithLookup<Person>> = new Map()
+  const spaces = new Map<Ref<Person>, PersonSpace>()
+
+  const p = await ctx.with('find-persons', {}, async (ctx) =>
+    (
+      await control.queryFind(
+        ctx,
+        contact.mixin.Employee,
```
kristina-fefelova: If we want to find here only `Employes`, so we should find info about notification sender separately, because sender might be not `Employee` (systemAccount, github user)

Now we find info about all users (sender + receivers) in this fn

server-plugins/notification-resources/src/index.ts
haiodo: Because document is not yet created, it could not have DocNotifyContext yet

packages/core/src/classes.ts
```diff
@@ -440,6 +440,8 @@ export interface Permission extends Doc {
 export interface Account extends Doc {
   email: string
   role: AccountRole
+
+  person?: Ref<Doc>
```
BykhovDenis, line 444: In this case we can throw away `PersonAccount`

server-plugins/notification-resources/src/index.ts
kristina-fefelova: got it!

plugins/contact-resources/src/index.ts
```diff
@@ -282,9 +282,11 @@ async function kickEmployee (doc: Person): Promise<void> {
       (res?: boolean) => {
         if (res === true) {
           // eslint-disable-next-line @typescript-eslint/no-floating-promises
-          getResource(login.function.LeaveWorkspace).then(async (f) => {
-            await f(email.email)
-          })
+          for (const i of accounts) {
+            void getResource(login.function.LeaveWorkspace).then(async (f) => {
```
BykhovDenis: I think better `getResource` once

BykhovDenis (dismissed):

server-plugins/chunter-resources/src/index.ts
haiodo: Fixed.

server-plugins/notification-resources/src/index.ts
haiodo: Fixed.

server-plugins/love-resources/src/index.ts
haiodo: Let's let it be this way as current implementation, since we will rework it in near future.

packages/core/src/classes.ts
haiodo: Yep, on next step.

BykhovDenis (dismissed):

kristina-fefelova (dismissed):

BykhovDenis (dismissed):

kristina-fefelova (approved):

BykhovDenis (approved):

PR #6345, opened 2024-08-16
add new channel page tests

jkrusty:
fix naming old test 'check if user can pin message' -> 'check if user can reply message'
add new tests:
'check if user can pin message'
'check if user can unpin message on pinned messages page'
'check if user can unpin message messages page'

added some necessary methods to models


haiodo:
Closing as outdated.

kristina-fefelova:
Closing as outdated.

PR #6390, opened 2024-08-26
Add tests for 'is typing...' status in channels

kakRostropovich:
**Add tests for [this PR](https://github.com/hcengineering/platform/pull/6373)**. 

I've added this PR to comply with this project structure and maybe work with us if you're interested.

About PR:
- Create a test suite inside the channel tests file. It contains some tests for the "[username] is typing..." feature
- Work with the current user and 2 temporary users. I've named them Alice and Bob, it's just for fun, I can rename them to user1 and user2 if necessary.
- Use existing methods for now. I think it's possible to create more page object's methods in the future to make my code and the code in this file more compact. I will be happy to do this if we continue to connect.
- I decided to group tests in this case, it doesn't look atomic enough, but I need to check, maybe I will be better to remove nested beforeEach and redesign it.

I hope this PR will be helpful for the team.
I'm going to keeping it in Draft, want to check test running in CI and maybe fix problems.

### DIFF
--- a/tests/sanity/tests/chat/chat.spec.ts
+++ b/tests/sanity/tests/chat/chat.spec.ts
@@ -1,4 +1,4 @@
-import { expect, test } from '@playwright/test'
+import { expect, Page, test } from '@playwright/test'
 import { ApiEndpoint } from '../API/Api'
 import { ChannelPage } from '../model/channel-page'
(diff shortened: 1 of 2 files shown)


kakRostropovich:
Request will be reopened without a fork

PR #6391, opened 2024-08-26
fix: parallel blob processing in tools an migration

aonnikov:


dev/tool/src/storage.ts
```diff
@@ -46,29 +53,37 @@ export async function moveFiles (
...
-          5,
-          async () => {
-            await moveFile(ctx, exAdapter, workspaceId, blob)
-          },
-          50
-        )
-      } catch (err) {
-        console.error('failed to process blob', name, data._id, err)
-      }
+      await rateLimiter.add(async () => {
```
haiodo: better to use exec, it will fill next if some previous will complete

haiodo (approved):

PR #6460, opened 2024-09-02
UBERF-7684: Workspace service

lexiv0re:
* Account service only creates workspace records
* Workspace service actually creates and upgrades workspaces
* It is configurable per: region, operation
* Allows for multiple account services
* Allows for multiple workspace services
* Upgrade region operation starts when an upgrade workspace services sends a handshake to account

Related to: https://front.hc.engineering/workbench/platform/tracker/UBERF-7673

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NmQ1OWY3NmFjY2U3NzU1MjYyZmNiZWIiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIn0.GKGBQWX0d5RmooEHmZd-sL5b64p6ykAWdzsdajYlx1Y">Huly&reg;: <b>UBERF-8014</b></a></sub>


plugins/workbench-resources/src/connect.ts
```diff
@@ -83,13 +83,13 @@ export async function connect (title: string): Promise<Client | undefined> {
 
   setMetadata(presentation.metadata.Token, token)
 
-  if (workspaceLoginInfo?.creating === true) {
+  if (['pending-creation', 'creating'].includes(workspaceLoginInfo?.mode ?? '')) {
```
haiodo, line 86: Could we have some predicate instead? It used 3 times.

haiodo (approved):

plugins/workbench-resources/src/connect.ts
lexiv0re: Sure, will extract it into an utility in the next PR.

PR #6492, opened 2024-09-06
UBERF-8044: staging model version

lexiv0re:
* Take tagged commit date as model version for staging builds
* Fix attempts query in account service to work properly with old workspaces

Related to: https://front.hc.engineering/workbench/platform/tracker/UBERF-8044

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NmRhZTg4Y2UxNzVjYmM2NmQzYTk3ODMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIn0.BGqQOwKOnjRpunzh04ga1Ia5Jaw76oJhvbMxjArO8No">Huly&reg;: <b>UBERF-8045</b></a></sub>

### DIFF
--- a/common/scripts/show_version.js
+++ b/common/scripts/show_version.js
@@ -15,12 +15,31 @@
 
 const fs = require('fs')
 const path = require('path')
+const exec = require('child_process').exec
 
-try {
-  const versionFilePath = path.resolve(__dirname, 'version.txt')
-  const version = fs.readFileSync(versionFilePath, 'utf8').trim()
-  
-  console.log(version)
-} catch (error) {
-  console.log('0.6.0')
-}
+exec('git describe --tags --abbrev=0', (err, stdout) => {
+  if (err !== null) {
+    console.log('"0.7.0"')
+    return
+  }
+
+  const tag = stdout.trim()
+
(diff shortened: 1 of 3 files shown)


common/scripts/show_version.js
```diff
@@ -15,12 +15,31 @@
...
-}
+exec('git describe --tags --abbrev=0', (err, stdout) => {
+  if (err !== null) {
+    console.log('"0.7.0"')
+    return
+  }
+
+  const tag = stdout.trim()
+
+  if (tag.startsWith('s')) {
```
haiodo: i'm think we need it to be passed from CI as mode, like `--mode=date` or we should use two different scripts, so if some env is passed we use this method.

lexiv0re: Parametrised by an env var and added this var definition to the CI. Defaults to the old behavior (take version from the file).

haiodo (approved):

PR #6505, opened 2024-09-09
feat: codeblock syntax highlight

aonnikov:
add highligher to code blocks


haiodo (approved):

haiodo:
@aonnikov will github markdown also work fine?

aonnikov:
> @aonnikov will github markdown also work fine?

lowlight supports markdown, so yes, it will be highlighted

PR #6507, opened 2024-09-10
Update QMS docker-compose.yaml so it can view and download PDF files

JasminMus:
* Requires `127.0.0.1 host.docker.internal` in `/etc/hosts`


README.md
```diff
@@ -127,12 +127,17 @@ Alternatively, you can just execute:
...
 ```
 
-Accessing the URL http://localhost:8087 will lead you to the app in production mode.
+Add the following line to your /etc/hosts file
+
+```
+127.0.0.1 host.docker.internal
+```
+
+Accessing the URL http://host.docker.internal:8087 will lead you to the app in production mode.
```
haiodo: May be in development mode? Let's not confuse users.

dev/docker-compose.yaml
```diff
@@ -141,22 +141,22 @@ services:
...
-      - ACCOUNTS_URL=http://localhost:3000
+      - ACCOUNTS_URL=http://host.docker.internal:3000
       - UPLOAD_URL=/files
       - ELASTIC_URL=http://host.docker.internal:9200
-      - GMAIL_URL=http://localhost:8088
-      - CALENDAR_URL=http://localhost:8095
-      - TELEGRAM_URL=http://localhost:8086
-      - REKONI_URL=http://localhost:4004
-      - COLLABORATOR_URL=ws://localhost:3078
+      - GMAIL_URL=http://host.docker.internal:8088
```
haiodo, line 147: May be we name it huly.localhost? If any name is possible?

dev/docker-compose.yaml
```diff
@@ -92,10 +92,10 @@ services:
       # - WS_OPERATION=create
       - SERVER_SECRET=secret
       - MONGO_URL=mongodb://mongodb:27017?compressors=snappy
-      - TRANSACTOR_URL=ws://transactor:3333;ws://localhost:3333
+      - TRANSACTOR_URL=ws://transactor:3333;ws://host.docker.internal:3333
```
haiodo, line 95: Let's use two variants, it is better since it will be more clear internal or external variant are used.

dev/docker-compose.yaml
lexiv0re: It's possible but I think it's more common to use `host.docker.internal` as it's a well-known domain resolvable within default network created in Docker w/o any additional configurations of the containers. Some docker applications even automatically append host.docker.internal 127.0.0.1 to /etc/hosts on your host machine when you install them.

haiodo: ok.

README.md
lexiv0re: Thanks, updated

haiodo (approved):

dev/docker-compose.yaml
lexiv0re: Discussed

PR #6550, opened 2024-09-13
fix: codeblock various fixes

aonnikov:
1. Render language selector on init
2. Highlight code blocks in activity
3. Revert inline code highlight

### DIFF
--- a/plugins/diffview-resources/src/components/Highlight.svelte
+++ b/plugins/diffview-resources/src/components/Highlight.svelte
@@ -1,5 +1,5 @@
-//
-// Copyright © 2023 Hardcore Engineering Inc.
+<!--
+// Copyright © 2024 Hardcore Engineering Inc.
 //
 // Licensed under the Eclipse Public License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License. You may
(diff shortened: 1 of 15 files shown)


plugins/diffview-resources/src/components/Highlight.svelte
```diff
@@ -0,0 +1,24 @@
...
+<script lang="ts">
+  import { highlightText } from '../highlight'
+
+  export let value: string
+  export let language: string | undefined = undefined
+
+  $: highlighted = highlightText(value, { language })
+</script>
+
+{@html highlighted}
```
lexiv0re: Is it safe to use @html with a value produced from a user input?

haiodo: Nope, it is not safe at all.

aonnikov: it is safe here, because we use @html on the input processed by highlight.js

they mention that html produced by the highlighter **is safe**:
https://github.com/highlightjs/highlight.js/issues/3229

We can post-process with something like DOMPurify, but I do not see much sense.

lexiv0re: Looks good but should we add a couple of tests to verify that the input is sanitized so we have some additional layer of security. E.g. in case some defaults change in `highlight` or something.

aonnikov: Added dompurify 😄

haiodo (approved):

PR #6599, opened 2024-09-17
Visual check for PDF

JasminMus:

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NmU5MzExMTU2YTcxNjQ4OTUzYTZmNmYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIn0.W1psXFaHY5dKv7uX1B4BDL6T4_3nBBKF1e_C2joe5AQ">Huly&reg;: <b>UBERF-8147</b></a></sub>


lexiv0re:
@JasminMus hey, could you fix this PR so we can review and merge it?

qms-tests/sanity/tests/documents/ES-20.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+      await documentContentPage.openApprovals()
+      const documentApprovalsPage = new DocumentApprovalsPage(page)
+      await documentApprovalsPage.checkSuccessApproval(documentDetails.owner)
+      await attachScreenshot('TESTS-277_approve_document.png', page)
+    })
+    await test.step('6. Show Full screen preview', async () => {
+      await documentContentPage.clickDocumentThreeDots()
+      const pdfPages = new PdfPages(page)
+      await pdfPages.printToPdfClick()
+      await page.waitForTimeout(2000)
```
lexiv0re, line 89: Is it safe to wait 2s for pdf generation? Is there a more reliable way to check that the generation is completed?

qms-tests/sanity/tests/documents/ES-40.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    await allure.tms('TESTS-391', 'https://tracex.hc.engineering/workbench/platform/tracker/TESTS-391')
+    await test.step('2. check if non member can see space', async () => {
+      const folderName = faker.word.words(1)
+      const documentContentPage = new DocumentContentPage(page)
+      await documentContentPage.clickAddFolderButton()
+      await documentContentPage.fillDocumentSpaceFormManager(folderName)
+      await documentContentPage.changeDocumentSpaceMembers(folderName)
+      await documentContentPage.checkIfTheSpaceIsVisible(folderName, false)
+      await documentContentPage.clickDocumentsSpace()
+      await page.waitForTimeout(1000)
```
lexiv0re: Why do we need this timeout?

lexiv0re (dismissed):

qms-tests/sanity/tests/documents/ES-40.spec.ts
JasminMus: unfortunately we need, I noticed sometimes the test is flaky without it

qms-tests/sanity/tests/documents/ES-20.spec.ts
JasminMus: I tested it and it was working fine with 2s, the problem is that the pdf is inside the iframe and we can't detect it. I can maybe create a function to check if the content of an iframe is generated

qms-tests/sanity/tests/documents/ES-40.spec.ts
lexiv0re: The next thing executed after the timeout is:
`const teamspaceOrArrow = await this.page.isVisible('.w-full > button:nth-child(2)')`
Can we instead try tp increase the timeout of the function which waits for this thing to become visible?

qms-tests/sanity/tests/documents/ES-20.spec.ts
lexiv0re: I'm not sure you can check the content of the PDF viewer inside the iframe but you can probably wait for some iframe events like`domcontentloaded`/`networkidle0` or something which could help to improve the stability of the test and be less dependent from some random factors.

qms-tests/sanity/tests/documents/ES-40.spec.ts
JasminMus: we can increase the max timeout only, but it still sees this element right away so if we increase  the functionality of 

`const teamspaceOrArrow = await this.page.isVisible('.w-full > button:nth-child(2)')
`

it will just increase the max timeout to wait.

qms-tests/sanity/tests/documents/ES-20.spec.ts
JasminMus: that can be a bit risky since in some cases networkidle get stuck, but I can add in case it get stuck than it would be something like a promise race where it will be if the networkidle is done first proceed, if not in certain period then do that timeout


```
export async function waitForNetworIdle (page: Page, timeout = 2000): Promise<void> {
  await Promise.race([page.waitForLoadState('networkidle'), new Promise((resolve) => setTimeout(resolve, timeout))])
}
```

qms-tests/sanity/tests/documents/ES-40.spec.ts
lexiv0re: I assumed that it was failing there w/o the explicit timeout. Where it was failing then?

qms-tests/sanity/tests/documents/ES-20.spec.ts
lexiv0re: It was just an example and if it doesn't work in this case we should not use it. But if it's supposed to go "networkidle" normally I'd use this approach and with the bigger race timeout. Also, we can wait for a combination of events if it helps to achieve more stability for the test.
IMO The main goal would be not to save 2 or even 15 seconds every run but to prevent flakiness of the test as a flaky test causes much more time being wasted later.

qms-tests/sanity/tests/documents/ES-40.spec.ts
JasminMus: it just skipped the step, like it clicked before the page was rendered. Let me try something

qms-tests/sanity/tests/documents/ES-20.spec.ts
JasminMus: totally agreed on this one, in my opinion this is currently more than enough. I will try with race condition to see how it behaves since well this is a bit tricky. The tests were passing without an issue with this hardcoded timeouts. I will try to update it to not use them

lexiv0re (approved):

PR #6724, opened 2024-09-25
feat: restore lost markup tool

aonnikov:


lexiv0re (dismissed):

dev/tool/src/markup.ts
```diff
@@ -212,3 +220,123 @@ async function processMigrateMarkupFor (
...
+            } else if (innerTx._class === core.class.TxUpdateDoc) {
+              const updateTex = innerTx as TxUpdateDoc<Doc>
+              markup = (updateTex.operations as any)[attr.name]
+            } else {
+              continue
+            }
+
+            if (markup === undefined || !markup.startsWith('{')) continue
+
+            console.log(doc._class, doc._id, attr.name, markup)
```
kristina-fefelova, line 298: Do we need this and below logs?

aonnikov: Yeah, this is to understand what we are going to changes. So the idea is to run one time with ‘show’ parameter and see what will be changed, then you run once again with ‘restore’ and see what changed.

dev/tool/src/markup.ts
```diff
@@ -212,3 +220,123 @@ async function processMigrateMarkupFor (
...
+
+          // try to restore by txes
+          // we need last tx that modified the attribute
+
+          const txes = await connection.findAll(core.class.Tx, query, { sort: { modifiedOn: SortingOrder.Descending } })
+          for (const tx of txes) {
+            const innerTx = TxProcessor.extractTx(tx)
+
+            let markup: string | undefined
+            if (innerTx._class === core.class.TxMixin) {
```
kristina-fefelova, line 282: Do  you need this `if`?
you didnt request  `core.class.TxMixin` : `class: { $in: [core.class.TxCreateDoc, core.class.TxUpdateDoc] }`

aonnikov: It is a bug, we need to request `core.class.TxMixin` as well

dev/tool/src/markup.ts
kristina-fefelova: ok, if it's planned
I just thought that is too many logs for each tx

dev/tool/src/markup.ts
aonnikov: Fixed

kristina-fefelova (approved):

PR #6802, opened 2024-10-03
fiix: disable failing planner test

aonnikov:


haiodo (dismissed):

tests/sanity/tests/planning/plan.spec.ts
```diff
@@ -377,6 +377,7 @@ test.describe('Planning ToDo tests', () => {
     })
 
     await test.step('Check time changes', async () => {
+      test.skip()
```
haiodo: I think you skipped not a test but a step. Should we skip test instead?

haiodo (approved):

tests/sanity/tests/planning/plan.spec.ts
aonnikov: :) Fixed

PR #6819, opened 2024-10-05
Ignore diff in blank lines when posting issue back to github

akhismat:
Ignore diff in blank lines when posting issue back to github

### DIFF
--- a/packages/text/src/markdown/compare.ts
+++ b/packages/text/src/markdown/compare.ts
@@ -34,13 +34,20 @@ export function calcSørensenDiceCoefficient (a: string, b: string): number {
  * Perform markdown diff/comparison to understand do we have a major differences.
  */
 export function isMarkdownsEquals (source1: string, source2: string): boolean {
-  const lines1 = source1
-    .split('\n')
-    .map((it) => it.trimEnd())
-    .join('\n')
-  const lines2 = source2
-    .split('\n')
-    .map((it) => it.trimEnd())
-    .join('\n')
+  const normalizeLineEndings = (str: string) =>
+    str.replace(/\r?\n/g, '\n');
+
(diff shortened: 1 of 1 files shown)


packages/text/src/markdown/compare.ts
```diff
@@ -34,13 +34,20 @@ export function calcSørensenDiceCoefficient (a: string, b: string): number {
...
 export function isMarkdownsEquals (source1: string, source2: string): boolean {
-  const lines1 = source1
-    .split('\n')
-    .map((it) => it.trimEnd())
-    .join('\n')
-  const lines2 = source2
-    .split('\n')
-    .map((it) => it.trimEnd())
-    .join('\n')
+  const normalizeLineEndings = (str: string) =>
```
haiodo, line 37: If comments will not be supported we will potentially break description on github, let's support html/comments first and then add a trim of lines >2 into 1 into comparison, but not allow without line and with line be the same.

haiodo (changes requested):
I'm not ready to merge it before comments will be supported.

packages/text/src/markdown/compare.ts
akhismat: Just fyi: even with this change description with comment and without comment will not be equal.

akhismat:
@haiodo 
There are two independent problems:
1. Extra empty lines.
2. Comments, which are not yet supported. Huly just removes them from a description and never put the issue back to GitHub because the description was changed. 

This PR has fix only for №1: 
Allow put issue back if changes are in empty lines only.

№2 is in progress 
But even with this change Huly will never re-write the description in case the original issue has comments in the description.

Are you OK with this fix for №1 only?

PR #6836, opened 2024-10-08
refactor: server/client: getAccount refactored to use early returns

SomeoneIsWorking:
ClientSession::getAccount changed to use less indentation with early returns.

### DIFF
--- a/server/server/src/client.ts
+++ b/server/server/src/client.ts
@@ -104,32 +104,35 @@ export class ClientSession implements Session {
 
   async getAccount (ctx: ClientSessionCtx): Promise<void> {
     const account = this._pipeline.context.modelDb.getAccountByEmail(this.token.email)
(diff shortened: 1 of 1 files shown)


server/server/src/client.ts
```diff
@@ -104,32 +104,35 @@ export class ClientSession implements Session {
...
-      }
+
+    if (account !== undefined || this.token.extra?.admin !== 'true') {
+      await ctx.sendResponse(account);
+      return;
+    }
+    
+    const systemAccount = this._pipeline.context.modelDb.findObject(this.token.email as Ref<Account>);
+    if (systemAccount === undefined) {
+      await ctx.sendResponse(systemAccount);
```
SomeoneIsWorking, line 115: This just responds with `undefined`. Is this correct? I haven't changed the logic.

PR #6858, opened 2024-10-10
UBERF-7911: Branding worker

lexiv0re:
* Adds a supplementary branding worker to be later used in other workers

Related to: https://front.hc.engineering/workbench/platform/tracker/UBERF-7911

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzA3Nzk0ZTg0YTQ0MTNjOGY4MWI1OTYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIn0.f5R15qPQQRPGC50PKXH_Cd5G_6o6ZvGsdL_ETJCKZ7w">Huly&reg;: <b>UBERF-8406</b></a></sub>


workers/branding/wrangler.toml
```diff
@@ -0,0 +1,115 @@
...
+[build]
+command = "npm run build"
+watch_dir = "./src"
+
+[[kv_namespaces]]
+binding = "KV"
+id = "c020474a-1223-402b-be0c-cd122bab9599"
+preview_id = "c320577a-5563-457b-be0c-cd152bab5599"
+
+# [[r2_buckets]]
```
aonnikov: we don't need this commented configuration, let's remove it

workers/branding/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const router = Router<Request>({
+  before: [preflight],
+  catch: error,
+  finally: [(r: any) => r ?? error(404), corsify]
+})
+
+async function getBrandings(env: Env): Promise<BrandingMap> {
+  // Note: up to 100 keys can be listed at once. If more, use cursor.
+  const keys = (await env.KV.list()).keys
```
aonnikov: what do you think about using prefix for all branding keys? so you can do `env.KV.list({ prefix: 'branding' })`

aonnikov (dismissed):

workers/branding/src/index.ts
lexiv0re: I was thinking we'll have a separate KV namespace dedicated to branding only. Do you think it would be better to share it with something else?

workers/branding/wrangler.toml
lexiv0re: Done

workers/branding/src/index.ts
aonnikov: not share, but just in case you have to put something else into this KV but related to brandings

aonnikov (dismissed):

workers/branding/src/index.ts
lexiv0re: Thanks, updated

aonnikov (approved):

PR #6937, opened 2024-10-15
UBERF-8433: Support for archived workspaces

haiodo:
UBERF-8433: Support for archived workspaces


packages/core/src/classes.ts
```diff
@@ -680,6 +680,10 @@ export interface BackupStatus {
 export interface BaseWorkspaceInfo {
   workspace: string // An uniq workspace name, Database names
   disabled?: boolean
+
+  archived?: boolean
```
lexiv0re: As an idea: extend `WorkspaceMode` with `archived` and use `lastProcessingTime` as an archived date when the mode is archived. Otherwise, we need to explicitly alter workspace table in postgres via a new migration to support the new fields. 
What do you think?

haiodo: Yes, we could use mode, will update tomorrow

aonnikov (dismissed):

BykhovDenis (dismissed):

lexiv0re (dismissed):

aonnikov (dismissed):

lexiv0re (approved):

aonnikov (approved):

PR #6952, opened 2024-10-16
feat: datalake worker initial version

aonnikov:


packages/presentation/src/file.ts
```diff
@@ -61,6 +119,42 @@ export function getUploadUrl (): string {
...
+
+  return { method: 'form-data', url: config['form-data'].url }
+}
+
+/**
+ * @public
+ */
+export function getFileUploadParams (blobId: string, blob: Blob): FileUploadParams {
+  const workspaceId = encodeURI(getCurrentWorkspaceId())
+  const fileId = encodeURI(blobId)
```
lexiv0re: Minor: is the use of `encodeURI` instead of `encodeURIComponent` intended?

packages/presentation/src/file.ts
```diff
@@ -110,24 +232,54 @@ export async function uploadFile (file: File): Promise<Ref<Blob>> {
...
-export async function deleteFile (id: string): Promise<void> {
-  const fileUrl = getFileUrl(id)
-
-  const resp = await fetch(fileUrl, {
-    method: 'DELETE',
+async function uploadFileWithSignedUrl (file: File, uuid: string, uploadUrl: string): Promise<void> {
+  const response = await fetch(uploadUrl, {
+    method: 'POST',
     headers: {
       Authorization: 'Bearer ' + (getMetadata(plugin.metadata.Token) as string)
```
lexiv0re, line 241: Do we need auth when uploading with signed url?

lexiv0re: Nvm, it's actually a step of obtaining one :)

aonnikov: Yep, we need auth to obtain url and do NOT need to upload with signed URL. An attempt to use auth with signed URL will result in 401 or 403 error.

packages/presentation/src/file.ts
aonnikov: nope, that's a bug

aonnikov: resolved

server/datalake/src/client.ts
```diff
@@ -158,30 +117,35 @@ export class Client {
...
-      throw new Error('HTTP error ' + response.status)
+  async putObject (
+    ctx: MeasureContext,
+    workspace: WorkspaceId,
+    objectName: string,
+    stream: Readable | Buffer | string,
+    metadata: ObjectMetadata,
+    size?: number
+  ): Promise<void> {
+    if (size === undefined || size < 64 * 1024 * 1024) {
```
lexiv0re, line 131: Should the size limit be taken from the config now?

lexiv0re (approved):

server/datalake/src/client.ts
aonnikov: Yeah, maybe, will change in another PR

PR #6978, opened 2024-10-18
UBERF-8511: Configurable account db NS

lexiv0re:
* Adds configuration for account db namespace as `DB_NS` account service env var

Related to: https://front.hc.engineering/workbench/platform/tracker/UBERF-8511


server/account-service/src/index.ts
```diff
@@ -95,7 +95,8 @@ export function serveAccount (measureCtx: MeasureContext, brandings: BrandingMap
   const hasSignUp = process.env.DISABLE_SIGNUP !== 'true'
   const methods = getMethods(hasSignUp)
 
-  const accountsDb = getAccountDB(dbUrl)
+  const dbNs = process.env.DB_NS
```
haiodo, line 98: Please add default, since self-host users will forget to set value

haiodo (approved):

PR #6987, opened 2024-10-18
Speedup move tool and fix backup error

haiodo:
Speedup move tool and fix backup error

### DIFF
--- a/dev/tool/src/index.ts
+++ b/dev/tool/src/index.ts
@@ -15,7 +15,6 @@
 //
 
 import accountPlugin, {
-  type AccountDB,
   assignWorkspace,
   confirmEmail,
   createAcc,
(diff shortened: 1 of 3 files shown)


dev/tool/src/index.ts
```diff
@@ -1100,7 +1185,7 @@ export function devTool (
     .command('move-files')
     .option('-w, --workspace <workspace>', 'Selected workspace only', '')
     .option('-m, --move <move>', 'When set to true, the files will be moved, otherwise copied', 'false')
-    .option('-bl, --blobLimit <blobLimit>', 'A blob size limit in megabytes (default 50mb)', '50')
+    .option('-bl, --blobLimit <blobLimit>', 'A blob size limit in megabytes (default 50mb)', '999999')
```
aonnikov, line 1188: `default 50mb` should be changed as well

aonnikov (approved):

dev/tool/src/index.ts
haiodo: Will do on next PR ;-)

PR #6989, opened 2024-10-18
UBERF-8508: Get rid of Mongo in storage adapter

haiodo:
1. UBERF-8508: Get rid of Mongo in storage adapter
2. Will add migration to remove blobs collection on next PR.


server/server-storage/src/fallback.ts
```diff
@@ -0,0 +1,268 @@
...
+    this.code = 'NoSuchKey'
+  }
+}
+
+/**
+ * Perform operations on storage adapter and map required information into BinaryDocument into provided DbAdapter storage.
+ */
+export class FallbackStorageAdapter implements StorageAdapter, StorageAdapterEx {
+  // Adapters should be in reverse order, first one is target one, and next ones are for fallback
+  constructor (readonly adapters: [string, StorageAdapter][]) {}
```
aonnikov: can we use an object instead of [string, StorageAdapter] array? It is difficult to read code like `exAdapter.adapters[0][1]` :)

aonnikov (dismissed):

server/server-storage/src/fallback.ts
haiodo: Done.

aonnikov (approved):

Issue #6925, opened 2024-10-15
[Feature Request] Emoji selection when typing : (colon) symbol in the chat box


reynaldichernando:
### Description of the issue
I would like to request a feature for selecting emoji when typing colon in the chat box, for example typing :joy will show a selection of emoji to choose.

### Your environment
* Version of huly: [huly.app](huly.app)
* Browser (and version): Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.0 Safari/605.1.15
* Your operating system (and version): macOS Sequoia 15.0

### Steps to reproduce
Type something like :joy in the text input in chat.

### Expected behaviour
A selection of emojis would show up. (screenshot from Slack)
<img width="513" alt="image" src="https://github.com/user-attachments/assets/3512ee7f-d9b3-470c-9243-c7cf93be2792">

### Actual behaviour
Nothing happens on Huly.
<img width="345" alt="image" src="https://github.com/user-attachments/assets/ae163a0b-d403-457a-8dcd-5fb7dbb4f955">

### Possible solutions
Since we already have user selection when typing with @ (at) symbol, I think a similar solution can be implemented when typing the : (colon) symbol, but for searching emojis instead.

PR #6991, opened 2024-10-18
feat: #6925 - Emoji selection when typing : (colon) symbol in the chat box

ElSalvo96:
## About this PR

This PR adds a new feature for selecting emoji when typing a colon in the chat box; for example, typing :joy will show a selection of emoji to choose from.

This PR should close this issue: https://github.com/hcengineering/platform/issues/6925

The list of emoji is generated from `/platform/plugins/text-editor-resources/src/components/extension/emojiIdMapGenerator.ts`, a revisited file from https://github.com/ikatyang/emoji-cheat-sheet/blob/master/scripts/fetch.ts

### Before

image

### After

image
image

Note: Since this is my first experience using Svelte, I would appreciate any input.


plugins/text-editor-resources/src/components/EmojiList.svelte
```diff
@@ -0,0 +1,48 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+-->
+<script lang="ts">
+  import EmojiPopup from './EmojiPopup.svelte'
+  import Popup from './Popup.svelte'
+
+  export let query: string = ''
```
kasir-barati:
```suggestion
  export let query = ''
```

Redundant type annotation.

plugins/text-editor-resources/src/Completion.ts
```diff
@@ -177,3 +177,76 @@ export const Completion = Node.create<CompletionOptions>({
...
+                {
+                  type: 'text',
+                  text: ' '
+                }
+              ])
+              .run()
+          }
+        },
+        allow: ({ editor, range }) => {
+          if (range.from > editor.state.doc.content.size) return false
```
kasir-barati:
```suggestion
          if (range.from > editor.state.doc.content.size) {
            return false
          }
```

Just my taste you could say but I prefer to add those curlies even if I have only one line of code. Kinda put my mind at ease, and also help me to read the code more readily.

plugins/text-editor-resources/src/Completion.ts
kasir-barati, line 1: [Non-blocking comment]: In this file I have seen 3 `eslint-disable-next-line`. I do not wanna say do not use it, but rather I prefer to stay safe as much as possible. So for example if the type is an issue I try first to fix that and not use `eslint-disable-next-line`.

plugins/text-editor-resources/src/components/EmojiList.svelte
```diff
@@ -0,0 +1,48 @@
...
+  export let clientRect: () => ClientRect
+  export let command: (props: any) => void
+  export let close: () => void
+
+  let searchPopup: EmojiPopup
+
+  function dispatchItem(item: { id: string; objectclass: string }): void {
```
kasir-barati:
```suggestion
  function dispatchItem(item?: { id: string; objectclass: string }): void {
```

Just me trying to keep my function signature as self-explanatory as possible. So that the `if  else` makes more sense.

plugins/text-editor-resources/src/components/EmojiPopup.svelte
```diff
@@ -0,0 +1,134 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+-->
+<script lang="ts">
+  import presentation, { reduceCalls } from '@hcengineering/presentation'
+  import { Label, ListView, resizeObserver } from '@hcengineering/ui'
+  import { createEventDispatcher } from 'svelte'
+  import emojiReplaceDict from './extension/emojiIdMap.json'
+
+  export let query: string = ''
```
kasir-barati: same: redundant type annotation

plugins/text-editor-resources/src/components/EmojiPopup.svelte
```diff
@@ -0,0 +1,134 @@
...
+              >
+                <div class="flex-row-center">
+                  <div class="flex-center p-1 content-dark-color flex-no-shrink">
+                    {item[1]}
+                  </div>
+                  <span class="ml-1 max-w-120 overflow-label searchResult">
+                    <span class="name">{item[0]}</span>
+                  </span>
+                </div>
+              </div>
```
kasir-barati, line 103: Semantic html please!

I cannot help you with that here since I am not fully aware of the context. But the least we could do for accessibility is to add `aria-*` or other html attributes.

plugins/text-editor-resources/src/components/EmojiPopup.svelte
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  .searchResult {
+    display: flex;
+    flex-direction: row;
+
+    .name {
+      display: flex;
+      flex: 1;
+    }
```
kasir-barati, line 132: Love that you're using "CSS Nesting" :heart:.

plugins/text-editor-resources/src/components/Popup.svelte
```diff
@@ -0,0 +1,122 @@
...
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+-->
+<script lang="ts">
+  import { deviceOptionsStore as deviceInfo, PopupResult, resizeObserver, showPopup } from '@hcengineering/ui'
+  import { onDestroy, onMount } from 'svelte'
+  import DummyPopup from './DummyPopup.svelte'
+
+  export let query: string = ''
```
kasir-barati: same

plugins/text-editor-resources/src/components/Popup.svelte
```diff
@@ -0,0 +1,122 @@
...
+  })
+
+  export function done(): void {}
+
+  function updateStyle(): void {
+    const rect = clientRect()
+    const wDoc = $deviceInfo.docWidth
+    const hDoc = $deviceInfo.docHeight
+    let tempStyle = ''
+    if (rect.top < hDoc - rect.bottom) {
```
kasir-barati, line 54: Just out of curiosity, cannot we turn this into [decompose conditional](https://refactoring.guru/decompose-conditional)?

Same for the other `if else` blocks.

plugins/text-editor-resources/src/components/ReferenceInput.svelte
```diff
@@ -84,7 +84,7 @@
     insertTemplate: (name, markup) => {
       editor?.insertMarkup(markup)
     },
-    insertTable (options: { rows?: number, cols?: number, withHeaderRow?: boolean }) {
+    insertTable(options: { rows?: number; cols?: number; withHeaderRow?: boolean }) {
```
kasir-barati, line 87: To me it looks like someones IDE is not configured to format the code properly. Do we need some kind of pipeline to reject unformatted code? Or maybe husky to format the code before push?

plugins/text-editor-resources/src/components/extension/emojiIdMapGenerator.ts
kasir-barati, line 1: Could not we just remove this file?

You know what I mean, it was added in this PR and if it is not needed then should not be here as well.

kasir-barati (changes requested):
Hi,

Thanks for the PR. I jot down some pointer I could put my finger on. But since I do not know svelte I could not help you with that.

ElSalvo96:
Hello @kasir-barati, thanks for the review.

When I did this feature, I just copied from other components, avoiding editing too much of the logic itself. 
In that moment, I didn't care about the type inference or some other stuff like the eslint rules. I was more focused on shipping the feature faster without logical error. General speaking, I think that there is a lot of redundant code that can be optimized, but knowledge about the framework is still limited. 

In any case, I feel that you were right into a lot of the comment.

plugins/text-editor-resources/src/components/EmojiPopup.svelte
ElSalvo96: I did how they did. 
https://github.com/ElSalvo96/platform/blob/4e7673cb5b1af10e66cb84f05fba3790494b05d4/packages/presentation/src/components/SearchResult.svelte#L27-L46

What kind of improvement do you expect with the `aria-*` tag? I think there you don't need any of these

plugins/text-editor-resources/src/components/Popup.svelte
ElSalvo96: Yes I think so, this was an older component `MentionList.svelte`. I slitted in two to share the logic between this and `EmojiList.svelte` component

plugins/text-editor-resources/src/components/ReferenceInput.svelte
ElSalvo96: There is no husky in this project. 
The file is saved with the prettier-eslint extension. So I think that my new one is right.

plugins/text-editor-resources/src/components/extension/emojiIdMapGenerator.ts
ElSalvo96: Yes I could. I was put there because no one know if in the future is still need to generate the file. With that instead of going into the thought process, you only need to run this script. 

In any case, do you believe that deleting it is necessary? It's different approach it depends on how do they usually work. 
Let me know

plugins/text-editor-resources/src/components/ReferenceInput.svelte
kasir-barati: Yeah, I meant that it should be added otherwise these kind of unnecessary changes will be pushed and someone has to code review these kind of changes while we can prevent it simply by adding something like husky. I do not know how ususally thing are done here but if it were up to me I would have created a task called add husky or at least do some R&D around it to find out whether it worth the effort or not.

plugins/text-editor-resources/src/components/extension/emojiIdMapGenerator.ts
kasir-barati: Hmm, cannot we just doc it instead if it is really something useful somewhere else? Dunno what kind of documentation generating tool we use here but it could go there if you ask me.

kristina-fefelova:
Closing as outdated.

PR #7000, opened 2024-10-21
UBERF-8518: Optimize client model

haiodo:
UBERF-8518: Optimize client model

Github client ModelDb reduced from 1.7Mb to 700kb

### DIFF
--- a/plugins/client-resources/src/index.ts
+++ b/plugins/client-resources/src/index.ts
@@ -24,7 +24,16 @@ import core, {
   TxWorkspaceEvent,
   WorkspaceEvent,
   concatLink,
-  createClient
+  createClient,
+  fillConfiguration,
+  pluginFilterTx,
+  type Class,
+  type ClientConnection,
+  type Doc,
+  type ModelFilter,
+  type PluginConfiguration,
+  type Ref,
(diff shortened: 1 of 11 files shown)


plugins/client-resources/src/index.ts
```diff
@@ -70,68 +79,123 @@ export default async () => {
...
+  )
+  return pluginFilterTx(excludedPlugins, configs, txes)
+}
+
+function returnClientTxes (txes: Tx[]): Tx[] {
+  const configs = new Map<Ref<PluginConfiguration>, PluginConfiguration>()
+  fillConfiguration(txes, configs)
+  const excludedPlugins = Array.from(configs.values()).filter((it) => !it.enabled || it.pluginId.startsWith('server-'))
+
+  const toExclude = new Set([
```
aonnikov, line 167: if we add new class for UI specific tx, we will need to add new class here? Can we solve this with some attribute in tx instead?

haiodo: I've just filtered out most significant ones, so no need to update it often.

aonnikov (approved):

BykhovDenis (approved):

PR #7036, opened 2024-10-25
Fix online/offline status when person has multiple accounts

kristina-fefelova:

### DIFF
--- a/plugins/contact-resources/src/components/UserInfo.svelte
+++ b/plugins/contact-resources/src/components/UserInfo.svelte
@@ -15,11 +15,10 @@
 <script lang="ts">
   import Avatar from './Avatar.svelte'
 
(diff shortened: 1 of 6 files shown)


akhismat (dismissed):

plugins/contact-resources/src/components/UserInfo.svelte
```diff
@@ -30,17 +29,14 @@
 
   const client = getClient()
   const hierarchy = client.getHierarchy()
```
haiodo: We should not hold hierarchy in component, since it will be not updated on reconnect-refresh.

kristina-fefelova: Done

haiodo (approved):

PR #7039, opened 2024-10-25
UBERF-8557 Save collaborative content as JSON

aonnikov:


models/controlled-documents/src/types.ts
```diff
@@ -259,7 +259,7 @@ export class TDocument extends TDoc implements Document {
     state!: DocumentState
 
   @Prop(TypeCollaborativeDoc(), documents.string.CollaborativeDocument)
-    content!: CollaborativeDoc
+    content!: Ref<Blob> | null
```
lexiv0re: Minor: as an idea we may have kept a separate type for collaborative content which would simply be an alias for `Ref<Blob>`. It could improve readability and make it simpler for further refactoring if needed. WDYT?

models/core/src/migration.ts
```diff
@@ -258,6 +253,149 @@ async function processMigrateContentFor (
...
+    })
+
+    if (attributes.length === 0) continue
+    if (hierarchy.isMixin(_class) && attributes.every((p) => p.attributeOf !== _class)) continue
+
+    const query = hierarchy.isMixin(_class) ? { [_class]: { $exists: true } } : { _class }
+
+    const iterator = await client.traverse(domain, query)
+    try {
+      console.log('processing', _class)
```
lexiv0re: Minor: use `ctx.info` instead if you wanna keep these logs? There are more console.logs across the file.

packages/core/src/collaboration.ts
```diff
@@ -13,175 +13,41 @@
...
 
 /** @public */
-export function makeCollaborativeDoc (
-  objectId: Ref<Doc>,
-  objectAttr?: string | undefined,
-  versionId?: string | undefined
+export function makeCollabId<T extends Doc, U extends keyof T> (
+  objectClass: Ref<Class<T>>,
+  objectId: Ref<T>,
+  objectAttr: Extract<U, string> | string
```
lexiv0re, line 29: Just curious, isn't `| string` removing the check for the attribute to be a valid key completely?

packages/core/src/collaboration.ts
```diff
@@ -13,175 +13,41 @@
...
- */
-export function collaborativeDocFromCollaborativeDoc (
-  collaborativeDoc: CollaborativeDoc,
-  sourceCollaborativeDoc: CollaborativeDoc
-): CollaborativeDoc {
-  return collaborativeDocWithSource(collaborativeDoc, collaborativeDocFromLastVersion(sourceCollaborativeDoc))
+export function makeCollabJsonId (doc: CollaborativeDoc): Ref<Blob> {
+  const timestamp = Date.now()
+  const { objectId, objectAttr } = doc
+  return [objectId, objectAttr, timestamp].join('-') as Ref<Blob>
```
lexiv0re: Minor: are you parsing it back anywhere? If yes, `-` seems a very popular character that can easily be a part of `objectId` somewhere or sometime later.

server/client/src/blob.ts
```diff
@@ -35,6 +35,7 @@ export class BlobClient {
       url = url.slice(0, url.length - 1)
     }
 
+    // TODO ???
```
lexiv0re: ? :)

models/controlled-documents/src/types.ts
aonnikov: yeah will do, can you suggest any name? :)

packages/core/src/collaboration.ts
aonnikov: Yep, but in editor it still works better than just string. I failed to do this properly because of many places where we process document in generic way when we don't have object type - in this case this utility function becomes unusable.

packages/core/src/collaboration.ts
aonnikov: nope, it is for file name, it will never be parsed back. We could use any other character like % or |, but in file names `-` is much better because it does not require URL sanitization and visually looks better than other characters.

lexiv0re (dismissed):
LGTM

models/controlled-documents/src/types.ts
lexiv0re: I'd go with something generic and vague like `CollaborativeContent` which will be clarified in the type definition itself.

packages/core/src/collaboration.ts
lexiv0re: Ok, let's keep it this way.

models/core/src/migration.ts
aonnikov: Done

models/controlled-documents/src/types.ts
aonnikov: Done

server/client/src/blob.ts
aonnikov: Removed :)

lexiv0re (dismissed):

akhismat:
All imports work fine with in this branch (tested on commit 29c3a0fc445941fcaeb8b3c0c3c108143de99b7d)

lexiv0re (dismissed):

lexiv0re (dismissed):

lexiv0re (approved):

PR #7054, opened 2024-10-28
UBERF-8553: Stats as separate service

haiodo:
UBERF-8553: Stats as separate service


pods/stats/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN apt-get update
+RUN apt-get install libjemalloc2
+RUN apt-get clean
+
+ENV LD_PRELOAD=libjemalloc.so.2
+ENV MALLOC_CONF=dirty_decay_ms:1000,narenas:2,background_thread:true
+
+COPY bundle/bundle.js ./
+
+EXPOSE 3000
```
aonnikov: 4900?

aonnikov (dismissed):

pods/stats/Dockerfile
haiodo: Good catch.

PR #7058, opened 2024-10-28
UBERF-8547: Inbox cleanup  and other

kristina-fefelova:
- Select next card in inbox on delete
- Show selected card in left panel when show unreads
- Try bulk query updates
- Remove unused scroll view component
- Make inner pushes transient and dont save in memdb
- Group inner pushes by doc
- Do not show inner pushes if target doc opened in the middle or in sidebar


packages/query/src/index.ts
```diff
@@ -1294,6 +1316,12 @@ export class LiveQuery implements WithTx, Client {
       }
       result.push(await this._tx(tx, docCache))
     }
+
+    if (this.queriesToUpdate.size > 0) {
```
kristina-fefelova, line 1307: I want the query to be updated once if we receive a bunch of updates in notify so I tried to bulk them
I thought we could mark queries we need to notify and invoke callbacks at the end

please take a look @haiodo

haiodo: I think we could do it always collect updated queries and call callbacks after all txes are processed. We even could collect updated queries like in 10-20ms and update them on timer, it will allow to bulk more updates into one.

haiodo: Also you do not need to wait for callback to complete, also callback call is not asynchronous so we could remove asynchronous from it :-)

haiodo (approved):

PR #7081, opened 2024-11-01
UBERF-8657 API Client package

aonnikov:


packages/api-client/src/account.ts
```diff
@@ -0,0 +1,66 @@
...
+      params: [user, password, workspace]
+    })
+  })
+
+  const result = await response.json()
+  return result.result?.token
+}
+
+/** @public */
+export async function selectWorkspace (
```
haiodo, line 46: let's not expose close this method.

packages/api-client/src/client.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  const { endpoint, token } = await getWorkspaceToken(url, options, config)
+  return await createClient(endpoint, token, options)
+}
+
+/** @public */
+export async function createClient (endpoint: string, token: string, options: ConnectSocketOptions): Promise<APIClient> {
+  addLocation(clientId, () => import(/* webpackChunkName: "client" */ '@hcengineering/client-resources'))
+
+  setMetadata(client.metadata.ClientSocketFactory, options.socketFactory)
```
haiodo: let's move all this 3 as parameters to client factory.

packages/api-client/src/socket/node.ts
```diff
@@ -0,0 +1,103 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import { type ClientSocket, type ClientSocketFactory } from '@hcengineering/client'
+import WebSocket from 'ws'
```
haiodo: please move to factory as require('ws') and make dependency optional.

packages/api-client/src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  workspace: string
+}
+
+/** @public */
+export type AuthOptions = PasswordAuthOptions | TokenAuthOptions
+
+/** @public */
+export interface ConnectSocketOptions {
+  socketFactory?: ClientSocketFactory
+  useBinaryProtocol?: boolean
```
haiodo: this should not be exposed

aonnikov: Only `useBinaryProtocol` or both `useXXX`?

packages/api-client/src/account.ts
aonnikov: this `account` module is internal and not supposed to be exposed to users. Is that ok or you want to do something else?

packages/api-client/src/client.ts
aonnikov: done

packages/api-client/src/socket/node.ts
aonnikov: done

packages/api-client/src/types.ts
haiodo: This is internal so, I think we do not need to expose them both.

packages/api-client/src/account.ts
haiodo: internal is fine.

packages/api-client/src/types.ts
aonnikov: removed both

haiodo (approved):

PR #7087, opened 2024-11-02
Fix pg application name

BykhovDenis:

### DIFF
--- a/server/postgres/src/utils.ts
+++ b/server/postgres/src/utils.ts
@@ -182,8 +182,9 @@ export function getDBClient (connectionString: string, database?: string): Postg
 
   if (existing === undefined) {
     const sql = postgres(connectionString, {
-      connectionString,
-      application_name: 'transactor',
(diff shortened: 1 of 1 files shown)


haiodo (approved):

server/postgres/src/utils.ts
```diff
@@ -182,8 +182,9 @@ export function getDBClient (connectionString: string, database?: string): Postg
 
   if (existing === undefined) {
     const sql = postgres(connectionString, {
-      connectionString,
-      application_name: 'transactor',
+      connection: {
+        application_name: 'transactor'
```
haiodo, line 186: we need it customizable

haiodo (approved):

server/postgres/src/utils.ts
BykhovDenis: We can pass any POSTGRES_OPTIONS

PR #7092, opened 2024-11-03
DEFAULT_LOGIN_METHOD env for when mail otp isn't available (self-hosted)

gptlang:
When self hosting, SMTP isn't supported yet. The default email based login doesn't work. Login by email is just a click away but it would be easier if that could just be made default.

This change makes it such that we can simply add

```bash
LOGIN_PASSWORD_ONLY=true
```

to the environment to switch the defaults around.

### DIFF
--- a/dev/prod/src/platform.ts
+++ b/dev/prod/src/platform.ts
@@ -294,6 +294,7 @@ export async function configurePlatform() {
 
   setMetadata(login.metadata.AccountsUrl, config.ACCOUNTS_URL)
   setMetadata(login.metadata.DisableSignUp, config.DISABLE_SIGNUP === 'true')
+  setMetadata(login.metadata.DefaultLoginMethod, config.DEFAULT_LOGIN_METHOD)
   setMetadata(presentation.metadata.FilesURL, config.FILES_URL)
   setMetadata(presentation.metadata.UploadURL, config.UPLOAD_URL)
   setMetadata(presentation.metadata.CollaboratorUrl, config.COLLABORATOR_URL)
--- a/dev/prod/src/platform.ts
+++ b/dev/prod/src/platform.ts
@@ -152,6 +152,7 @@ export interface Config {
   TELEGRAM_BOT_URL?: string
   AI_URL?:string
   DISABLE_SIGNUP?: string
+  DEFAULT_LOGIN_METHOD?: 'otp'|'password'
   // Could be defined for dev environment
   FRONT_URL?: string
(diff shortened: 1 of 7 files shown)


dev/prod/src/platform.ts
```diff
@@ -294,6 +294,7 @@ export async function configurePlatform() {
 
   setMetadata(login.metadata.AccountsUrl, config.ACCOUNTS_URL)
   setMetadata(login.metadata.DisableSignUp, config.DISABLE_SIGNUP === 'true')
+  setMetadata(login.metadata.LoginPasswordOnly, config.LOGIN_PASSWORD_ONLY === 'true')
```
lexiv0re: Please also apply for `desktop` client.

plugins/login-resources/src/components/LoginForm.svelte
```diff
@@ -22,9 +22,14 @@
 
   export let navigateUrl: string | undefined = undefined
   export let signUpDisabled = false
+  export let loginPasswordOnly = false
 
   let method: LoginMethods = LoginMethods.Otp
```
lexiv0re, line 28: Minor: better extract this as a param from this component instead?

server/front/src/starter.ts
```diff
@@ -119,6 +119,8 @@ export function startFront (ctx: MeasureContext, extraConfig?: Record<string, st
 
   const disableSignUp = process.env.DISABLE_SIGNUP
 
+  const loginPasswordOnly = process.env.LOGIN_PASSWORD_ONLY
```
lexiv0re: Minor: I don't really like the name.  IMO `DEFAULT_LOGIN_METHOD` would be more readable and extensible in case we'd want to add more than `otp`/`password` later.

lexiv0re:
Hello @gptlang , thanks for your contribution!

lexiv0re (changes requested):

aonnikov:
Just an idea, can we use SES url to determine if we should use OTP login method? If we don't have email service configured, then we should not have SES url and OTP is not possible.

lexiv0re:
@aonnikov yeah, that's a good idea. I thought about this too and it appears we don't currently have SES url in our front service config file. So we'd have to add SES_URL to the front only to check whether OTP is supported or not. Why I probably would't want to expose it right now is that with the upcoming changes in the account we may immediately have other OTP delivery methods connected to various social logins other than email (e.g. telegram) so it might become a little bit tricky then.

PR #7106, opened 2024-11-05
Fix desktop file menu on Linux

akhismat:
Change settings shortcut for Linux

### DIFF
--- a/desktop/src/main/menu.ts
+++ b/desktop/src/main/menu.ts
@@ -16,6 +16,7 @@
 import { Menu, MenuItemConstructorOptions, type BrowserWindow } from 'electron'
 
 const isMac = process.platform === 'darwin'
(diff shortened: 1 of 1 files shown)


haiodo (approved):

desktop/src/main/menu.ts
```diff
@@ -24,7 +25,7 @@ export const addMenus = (getWindow: () => BrowserWindow, sendCommand: (cmd: stri
       submenu: [
         {
           label: 'Settings',
-          accelerator: 'Meta+,',
+          accelerator: isLinux ? 'Ctrl+,' : 'Meta+,',
```
haiodo, line 28: should it be same for windows?

akhismat: I will test and update if needed.

haiodo (approved):

PR #7121, opened 2024-11-07
fix: adjust bump script

aonnikov:
1. fix repository link
2. remove repository from assets packages
3. add fix option to bump script

### DIFF
--- a/packages/query/package.json
+++ b/packages/query/package.json
@@ -42,5 +42,9 @@
     "@hcengineering/core": "^0.6.32",
     "@hcengineering/analytics": "^0.6.0",
     "fast-equals": "^5.0.1"
+  },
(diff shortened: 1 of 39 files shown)


BykhovDenis (dismissed):

packages/query/package.json
```diff
@@ -42,9 +42,5 @@
     "@hcengineering/core": "^0.6.32",
     "@hcengineering/analytics": "^0.6.0",
     "fast-equals": "^5.0.1"
-  },
```
haiodo, line 45: we need to keep this publishable

aonnikov: removed because now it is not published, will add proper config

haiodo (approved):

PR #7154, opened 2024-11-11
UBERF-8520: Test management

ArtyomSavchenko:
Draft version of test management extension with the following functionality:
1. New navigation component for parent objects
Screenshot 2024-11-12 at 17 02 34
Screenshot 2024-11-12 at 17 02 23
2. Basic operations for test projects
Screenshot 2024-11-12 at 17 45 14
3. Basic operations for test suites
Screenshot 2024-11-12 at 17 46 16

4. Basic operations for test cases
Screenshot 2024-11-12 at 17 16 35


models/test-management/src/migration.ts
aonnikov, line 1: empty migration not needed

models/test-management/src/types.ts
```diff
@@ -0,0 +1,221 @@
...
+  @Prop(TypeString(), testManagement.string.TestRunName)
+  @Index(IndexKind.FullText)
+    name!: string
+
+  @Prop(TypeCollaborativeDoc(), testManagement.string.FullDescription)
+  @Index(IndexKind.FullText)
+    description!: CollaborativeDoc
+
+  @Prop(TypeDate(DateRangeMode.DATETIME), testManagement.string.DueDate)
+  declare dueDate?: Timestamp
```
aonnikov: it should not be `declare`, should it?

plugins/test-management/src/types.ts
```diff
@@ -0,0 +1,126 @@
...
+  fullDescription?: Markup
+}
+
+/** @public */
+export interface TestSuite extends Doc {
+  space: Ref<TestProject>
+  name: string
+  description?: string
+  parent?: Ref<TestSuite>
+}
```
aonnikov, line 87: not sure if we need this, but sometimes test suites and test cases have order of execution, that can be adjusted, maybe we should add `rank` so users can change order of execution manually. But I'm not sure how it is going to be used with test case priority together.

plugins/test-management/src/types.ts
```diff
@@ -0,0 +1,126 @@
...
+  description: CollaborativeDoc
+  type: TestCaseType
+  priority: TestCasePriority
+  status: TestCaseStatus
+  // Estimated time in minutes
+  estimatedTime: number
+  preconditions?: string
+  steps?: string
+  suite: Ref<TestSuite>
+  assignee: Ref<Employee>
```
aonnikov, line 96: What's the purpose of assignee? Is that the one who should write test case description or it is going to be default tester?

plugins/test-management/src/types.ts
```diff
@@ -0,0 +1,126 @@
...
+
+/** @public */
+export interface TestCase extends AttachedDoc {
+  name: string
+  description: CollaborativeDoc
+  type: TestCaseType
+  priority: TestCasePriority
+  status: TestCaseStatus
+  // Estimated time in minutes
+  estimatedTime: number
```
aonnikov: I think we don't need to add attributes like this, it is easy to add them via custom attributes.

models/test-management/src/types.ts
```diff
@@ -0,0 +1,221 @@
...
+  @Prop(TypeRef(testManagement.class.TestSuite), testManagement.string.TestSuite)
+    parent?: Ref<TestSuite>
+
+  declare space: Ref<TestProject>
+}
+
+/**
+ * @public
+ */
+@Model(testManagement.class.TestCase, core.class.Doc, DOMAIN_TEST_MANAGEMENT)
```
aonnikov: You should use `core.class.AttachedDoc` as parent class for classes that extend `AttachedDoc`

models/test-management/src/types.ts
```diff
@@ -0,0 +1,221 @@
...
+  @Prop(TypeString(), testManagement.string.SuiteName)
+  @Index(IndexKind.FullText)
+    name!: string
+
+  @Prop(TypeMarkup(), testManagement.string.SuiteDescription)
+  @Index(IndexKind.FullText)
+    description?: string
+
+  @Prop(TypeRef(testManagement.class.TestSuite), testManagement.string.TestSuite)
+    parent?: Ref<TestSuite>
```
aonnikov: I suggest to make it required and use well-known value `testManagement.ids.NoParent` instead of null/undefined.

plugins/test-management/src/types.ts
```diff
@@ -0,0 +1,126 @@
...
+  preconditions?: string
+  steps?: string
+  suite: Ref<TestSuite>
```
aonnikov: `TestCase` is and `AttachedDoc`, what it is attached to? If it is attached to test suite, then you don't need `suite` attribute, use `attachedTo` instead.

plugins/test-management/src/types.ts
```diff
@@ -0,0 +1,126 @@
...
+  attachments?: CollectionSize<Attachment>
+  comments?: number
+}
+
+/** @public */
+export interface TestRun extends Doc {
+  name: string
+  description: CollaborativeDoc
+  dueDate?: Timestamp
+  items?: CollectionSize<TestRunItem>
```
aonnikov, line 106: You use collection here but `TestRunItem` is not an `AttachedDoc`

plugins/test-management-resources/src/components/test-case/CreateTestCase.svelte
```diff
@@ -0,0 +1,192 @@
...
+  const hierarchy = client.getHierarchy()
+
+  let description = EmptyMarkup
+
+  fillDefaults(hierarchy, object, testManagement.class.TestCase)
+
+  async function createTestCase (): Promise<void> {
+    const op = client.apply()
+    await updateMarkup(object.description, { description })
+    await op.createDoc(testManagement.class.TestCase, _space, object, id)
```
aonnikov: If TestCase is an AttachedDoc, then you should use `addCollection` method.

aonnikov (changes requested):

plugins/test-management/src/types.ts
ArtyomSavchenko: I think for test case assignee is person who write test case description or who should review the test case.
Default tester should be assigned for test run

plugins/test-management-resources/src/components/test-case/CreateTestCase.svelte
ArtyomSavchenko: Thank you, fixed

models/test-management/src/types.ts
ArtyomSavchenko: Thank you, fixed

plugins/test-management/src/types.ts
ArtyomSavchenko: Thank you, fixed

models/test-management/src/types.ts
ArtyomSavchenko: Thank you, fixed

models/test-management/src/types.ts
ArtyomSavchenko: Thank you, fixed

plugins/test-management/src/types.ts
ArtyomSavchenko: Thank you, fixed

plugins/test-management/src/types.ts
ArtyomSavchenko: Thank you, fixed

plugins/test-management/src/types.ts
ArtyomSavchenko: Thank you, I think we can add it and test case id in a separate MR

plugins/test-management-resources/src/components/test-case/AssigneeEditor.svelte
aonnikov, line 1: please remove this in separate pull request, we should already have a generic component for this

haiodo (approved):

plugins/test-management-resources/src/components/test-case/AssigneeEditor.svelte
haiodo: Let's work on improvements on next PR's

aonnikov (approved):

PR #7162, opened 2024-11-12
Add draft survey app

Chunosov:
Adds an ability to create survey forms and conduct surveys attached to an entity (currently only available for the Application object in the recruit app)

image


dev/prod/package.json
```diff
@@ -233,6 +233,9 @@
     "@hcengineering/products-resources": "^0.1.0",
     "@hcengineering/ai-bot": "^0.6.0",
     "@hcengineering/ai-bot-resources": "^0.6.0",
+    "@hcengineering/survey": "^0.0.1",
```
lexiv0re: Please use `0.6.0` as the first version for consistency :)

dev/prod/src/platform.ts
```diff
@@ -231,6 +233,7 @@ function configureI18n(): void {
    addStringsLoader(loveId, async (lang: string) => await import(`@hcengineering/love-assets/lang/${lang}.json`))
    addStringsLoader(printId, async (lang: string) => await import(`@hcengineering/print-assets/lang/${lang}.json`))
    addStringsLoader(analyticsCollectorId, async (lang: string) => await import(`@hcengineering/analytics-collector-assets/lang/${lang}.json`))
+   addStringsLoader(surveyId, async (lang: string) => await import(`@hcengineering/survey-assets/lang/${lang}.json`))
```
lexiv0re, line 239: Please apply similar changes to `platform.ts` in `desktop`

dev/prod/webpack.config.js
```diff
@@ -330,6 +330,7 @@ module.exports = [
       disableDotRule: true
     },
     host: '0.0.0.0',
+    allowedHosts: 'all',
```
lexiv0re, line 333: Just curious why this is needed?

models/recruit/src/types.ts
```diff
@@ -159,6 +160,9 @@ export class TApplicant extends TTask implements Applicant {
   @Prop(TypeRef(core.class.Status), task.string.TaskState, { _id: recruit.attribute.State })
   @Index(IndexKind.Indexed)
   declare status: Ref<Status>
+
+  @Prop(Collection(survey.class.Poll), survey.string.Polls)
+    polls?: number
```
lexiv0re: 1. Please add this field to `Applicant` interface as well.
2. Minor: please use `Collection<Poll>` type to increase readability.

models/survey/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@hcengineering/model-survey",
+  "version": "0.0.1",
```
lexiv0re: Please use `0.6.0` for this and the other new packages for consistency.

models/survey/src/migration.ts
```diff
@@ -0,0 +1,42 @@
...
+export const surveyOperation: MigrateOperation = {
+  async migrate (client: MigrationClient, logger: ModelLogger): Promise<void> {
+    await tryMigrate(client, surveyId, [])
+  },
+  async upgrade (state: Map<string, Set<string>>, client: () => Promise<MigrationUpgradeClient>): Promise<void> {
+    await tryUpgrade(state, client, surveyId, [
+      {
+        state: 'createSpace',
+        func: async (client) => {
+          await createDefaultSpace(client, survey.space.Survey, { name: 'Surveys', description: 'Surveys' })
```
lexiv0re, line 37: Don't you need `autoJoin: true`? Did you test if new users in your workspace can also see your surveys?

models/survey/src/types.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import { IndexKind, type Domain, type Ref } from '@hcengineering/core'
+import { ArrOf, Index, Model, Prop, TypeRecord, TypeRef, TypeString, UX } from '@hcengineering/model'
+import core, { TAttachedDoc, TDoc } from '@hcengineering/model-core'
+import { getEmbeddedLabel } from '@hcengineering/platform'
+import { type Poll, type Question, type Survey } from '@hcengineering/survey'
+import survey from './plugin'
+
+export const DOMAIN_SURVEY = 'survey' as Domain
+export const DOMAIN_POLL = 'survey-poll' as Domain
```
lexiv0re: I think one (survey) domain should be enough.

models/survey/src/types.ts
```diff
@@ -0,0 +1,54 @@
...
+  @Prop(TypeRef(survey.class.Survey), survey.string.Survey)
+    survey!: Ref<Survey>
+
+  @Prop(TypeString(), survey.string.Name)
+    name!: string
+
+  @Prop(TypeString(), survey.string.Prompt)
+    prompt!: string
+
+  @Prop(ArrOf(TypeRecord()), getEmbeddedLabel('Answers'))
```
lexiv0re, line 51: Minor: if it's not a hidden field embedded label might not be a good idea as it's not translatable.

plugins/survey-assets/lang/ru.json
```diff
@@ -0,0 +1,38 @@
+{
```
lexiv0re, line 1: Minor: it would be good to provide translations for the other languages in the next PR.

plugins/survey-assets/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@hcengineering/survey-assets",
+  "version": "0.0.1",
+  "main": "src/index.ts",
+  "author": "Anticrm Platform Contributors",
```
lexiv0re: Please use `Hardcore Engineering Inc` for new modules.

plugins/survey-resources/src/components/EditQuestion.svelte
```diff
@@ -0,0 +1,462 @@
...
+      : question.kind === QuestionKind.OPTIONS
+        ? survey.icon.QuestionKindOptions
+        : question.kind === QuestionKind.OPTION
+          ? survey.icon.QuestionKindOption
+          : survey.icon.QuestionKindString
+
+  let newOption = ''
+  let newQuestion = ''
+
+  async function updateParent (): Promise<void> {
```
lexiv0re, line 45: Minor: with parent being updated on every change it feels it'd be more correct to have questions as a collection of separate docs attached to a survey. Then you will be CRUD-ing individual docs (questions).
Can be updated later if needed.

dev/prod/package.json
Chunosov: Changed

dev/prod/src/platform.ts
Chunosov: Added

dev/prod/webpack.config.js
Chunosov: I run the server on a VM and the client on the host. Without this option, when I havigate to `host.docker.internal:8080` the server answer with the plain text content "Invalid Host header" with code 200 :)

models/recruit/src/types.ts
Chunosov: 1. Added
2. There is already imported functon `Collection` used for property typing. So I had to import type `Collection as Array`. (It'd be better to name the function `TypeCollection` to be consistent with other `Type*` functions)

models/survey/package.json
Chunosov: Changed

models/survey/src/migration.ts
Chunosov: Yes, a new user can see old surveys and create a new ones. Both users see the same surveys table.

models/survey/src/types.ts
Chunosov: It's hidden on the viewlet table level. Added `@Hidden` here

plugins/survey-assets/package.json
Chunosov: Changed

plugins/survey-assets/lang/ru.json
Chunosov: Sure, I just was not sure about the final set of strings since it's a draft app

models/survey/src/types.ts
Chunosov: Removed the second domain. I considered them as names for tables having different structure, but it is not the approach here :)

plugins/survey-resources/src/components/EditQuestion.svelte
Chunosov: I think surveys are not supposed to have a lot of questions. So it was easier to store all them in one place. We can extract them into separate objects if see the performance suffers.

models/recruit/src/types.ts
lexiv0re: Yes, there's such problem. But `Array` is really confusing as it interferes with the standard `Array` type.
I usually rename the property type one instead to something like: `Collection as CollectionType`

haiodo (dismissed):

haiodo (approved):

PR #7180, opened 2024-11-15
fix: better handle markup in api client

aonnikov:
simplify usage of collaborative markup when creating/updating documents

### DIFF
--- a/packages/api-client/src/markup/types.ts
+++ b/packages/api-client/src/markup/types.ts
@@ -21,6 +21,24 @@ export type MarkupRef = CollaborativeDoc
 /** @public */
 export type MarkupFormat = 'markup' | 'html' | 'markdown'
 
+/** @public */
+export class MarkupContent {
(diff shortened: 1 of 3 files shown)


packages/api-client/src/markup/types.ts
```diff
@@ -21,6 +21,24 @@ export type MarkupRef = CollaborativeDoc
...
+/** @public */
+export class MarkupContent {
+  constructor (
+    readonly content: string,
+    readonly kind: MarkupFormat
+  ) {}
+}
+
+/** @public */
+export function html (content: string): MarkupContent {
```
haiodo, line 33: Do we really need html?

aonnikov: Yes, why not. HTML is still very popular and native for some editors, including tiptap. It is easier to convert from/to HTML than markdown.

haiodo (approved):

PR #7189, opened 2024-11-18
UBERF-8592: Fix live query performance

haiodo:
UBERF-8592: Fix live query performance


packages/query/src/__tests__/query.test.ts
```diff
@@ -717,6 +718,7 @@ describe('query', () => {
     await secondPromise
   })
 
+  jest.setTimeout(5000000)
```
BykhovDenis: Extra

packages/query/src/__tests__/query.test.ts
```diff
@@ -757,6 +759,10 @@ describe('query', () => {
       )
     })
 
+    await new Promise((resolve) => {
```
BykhovDenis, line 761: Why?

plugins/view-resources/src/actions.ts
```diff
@@ -68,9 +74,11 @@ export async function getActions (
   derived: Ref<Class<Doc>> = core.class.Doc,
   mode: ViewContextType = 'context'
 ): Promise<Action[]> {
-  const actions: Action[] = await client.findAll(view.class.Action, {
-    'context.mode': mode
-  })
+  let actions: Action[] | undefined = allActions.get(mode)
+  if (actions === undefined) {
+    actions = client.getModel().findAllSync(view.class.Action, { 'context.mode': mode })
+    allActions.set(mode, actions)
```
BykhovDenis, line 80: For what? Get values from model should be instant, i don't think that it's really improve perfomance

plugins/view-resources/src/components/filter/ObjectFilter.svelte
```diff
@@ -57,28 +66,50 @@
...
-        { projection: { _id: 1, archived: 1, _class: 1 } }
+    let baseObjects: WithLookup<Doc>[] = await client.findAll(filter.key._class, space !== undefined ? { space } : {}, {
+      projection: { [filter.key.key]: 1, space: 1 },
+      lookup: {
+        space: core.class.Space
+      },
+      limit: 1000
+    })
+    if (baseObjects.length === 1000) {
+      // We have more soe let's fetch all
```
BykhovDenis: Typo 'soe'

plugins/view-resources/src/components/filter/ObjectFilter.svelte
```diff
@@ -57,28 +66,50 @@
...
       await objectsPromise
     }
     targets.clear()
 
-    const spaces = (
-      await client.findAll(
-        core.class.Space,
-        { archived: { $ne: true } },
-        { projection: { _id: 1, archived: 1, _class: 1 } }
+    let baseObjects: WithLookup<Doc>[] = await client.findAll(filter.key._class, space !== undefined ? { space } : {}, {
```
BykhovDenis: It's really strange func... Maybe we should implement honest distinct?...

plugins/view-resources/src/components/filter/ValueFilter.svelte
```diff
@@ -75,47 +84,39 @@
...
+            space: core.class.Space
+          }
         }
       )
       if (limit !== undefined) {
         objectsPromise = p
       }
-      const res = await p
+      let res: WithLookup<Doc>[] = await p
+      // We need to filter archived in case it is required
```
BykhovDenis, line 116: Maybe better use db for it, and don't lookup space if we don't need it)

packages/query/src/index.ts
```diff
@@ -415,27 +378,30 @@ export class LiveQuery implements WithTx, Client {
...
 
   private remove (): void {
-    const q = this.queue.shift()
-    if (q === undefined) return
-    const queries = this.queries.get(q._class)
-    const pos = queries?.indexOf(q) ?? -1
-    if (pos >= 0 && queries !== undefined) {
-      queries.splice(pos, 1)
+    const used = Array.from(this.queue.values()).sort((a, b) => a.lastUsed - b.lastUsed)
+    for (let i = 0; i < CACHE_SIZE / 10; i++) {
```
BykhovDenis, line 389: Why /10?

plugins/view-resources/src/actions.ts
haiodo: It clone 250 actions every time for 1500 messages :-) in our meeting minutes

packages/query/src/__tests__/query.test.ts
haiodo: Since query call for first time after callback 1ms. Probable we need to remove it, and after optimization it become not called expected by test two times.

packages/query/src/index.ts
haiodo: To not remove every time, but from time to time

plugins/view-resources/src/components/filter/ValueFilter.svelte
haiodo: yes, will do

BykhovDenis (approved):

PR #7209, opened 2024-11-20
Fixes no-priority issue ordering (#7147)

vshashenko:
#7147
When ordering issues by priority, the issues with no priority are now in the end.

### DIFF
--- a/plugins/view-resources/src/components/list/ListCategory.svelte
+++ b/plugins/view-resources/src/components/list/ListCategory.svelte
@@ -32,6 +32,7 @@
   import { DocWithRank, makeRank } from '@hcengineering/task'
   import { AnyComponent, AnySvelteComponent, ExpandCollapse, mouseAttractor } from '@hcengineering/ui'
   import { AttributeModel, BuildModelKey, ViewOptionModel, ViewOptions, Viewlet } from '@hcengineering/view'
+  import { DocWithPriority, sortDocsByPriority } from '@hcengineering/tracker'
   import { createEventDispatcher } from 'svelte'
   import { fade } from 'svelte/transition'
   import { showMenu } from '../../actions'
@@ -175,6 +176,13 @@
   }
 
   $: limited = limitGroup(items, limit)
+
+  $: {
+    if (viewOptions.orderBy[0] === 'priority') {
+      // When there's sorting by priority, there's 'priority' property.      
+      sortDocsByPriority(limited as unknown as DocWithPriority[])
+    }
+  }
 
   $: selectedObjectIdsSet = new Set<Ref<Doc>>(selectedObjectIds.map((it) => it._id))
 
--- a/packages/kanban/package.json
+++ b/packages/kanban/package.json
@@ -45,6 +45,7 @@
     "@hcengineering/core": "^0.6.32",
     "@hcengineering/presentation": "^0.6.3",
     "@hcengineering/notification": "^0.6.23",
-    "@hcengineering/rank": "^0.6.4"
+    "@hcengineering/rank": "^0.6.4",
+    "@hcengineering/tracker": "^0.6.24"
   }
 }
--- a/packages/kanban/src/components/KanbanRow.svelte
+++ b/packages/kanban/src/components/KanbanRow.svelte
@@ -30,6 +30,7 @@
   import { slide } from 'svelte/transition'
   import { CardDragEvent, DocWithRank, Item } from '../types'
   import { createQuery } from '@hcengineering/presentation'
+  import { DocWithPriority, sortDocsByPriority } from '@hcengineering/tracker'
 
   export let stateObjects: Item[]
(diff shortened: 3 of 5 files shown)


plugins/view-resources/src/components/list/ListCategory.svelte
```diff
@@ -32,6 +32,7 @@
   import { DocWithRank, makeRank } from '@hcengineering/task'
   import { AnyComponent, AnySvelteComponent, ExpandCollapse, mouseAttractor } from '@hcengineering/ui'
   import { AttributeModel, BuildModelKey, ViewOptionModel, ViewOptions, Viewlet } from '@hcengineering/view'
+  import { DocWithPriority, sortDocsByPriority } from '@hcengineering/tracker'
```
aonnikov, line 35: view-resources cannot depend on tracker

aonnikov (commented):
@vshashenko thanks for your contribution!

While the idea of this pull request might be right, I do not agree with the way it is implemented.

1. "No priority" first/last should be configurable by user. If I prioritize all issues, I may want to see issues without priority in the top, because I may want to prioritize them first - there might be more severe issues than already prioritized.
2. The changes should be applicable to any orderable data types with null value, I can add, say, Severity, field and want to have the same behavior.
3. Proper sorting should be done on the database level, otherwise it won't work for large lists

vsh1111:
Agree, especially on point 3. I forgot about pagination. I was looking at the code in the database level but couldn't wrap my head around it since it's my first time seeing it. Can you point at specific files and functions to look at?
Another thing with sorting by the database though. Priorities are defined in an enum as:
0-NoPriority, 1-Urgent, 2-High, 3-Medium, 4-Low. 
And as I understand these values go into the database as is, without mapping.
I guess we need this behavior:
Asc: 0-NoPriority, 4-Low, 3-Medium, 2-High, 1-Urgent.
Desc: 1-Urgent, 2-High, 3-Medium, 4-Low, 0-NoPriority.
I suspect the database can only do asc/desc sort, and to get the desired behavior we need it to treat NoPriority as some MAX_INT instead of 0. Migration the NoPriority value from 0 to MAX_INT for the whole database and updating client code at the same time can be problematic I think. Besides, can be pointless if in some cases we'd still need the sorting behavior as if NoPriority=0, although NoPriority=MAX_INT indeed looks like more common case.
What do you think?

aonnikov:
> Agree, especially on point 3. I forgot about pagination. I was looking at the code in the database level but couldn't wrap my head around it since it's my first time seeing it. Can you point at specific files and functions to look at? Another thing with sorting by the database though. Priorities are defined in an enum as: 0-NoPriority, 1-Urgent, 2-High, 3-Medium, 4-Low. And as I understand these values go into the database as is, without mapping. I guess we need this behavior: Asc: 0-NoPriority, 4-Low, 3-Medium, 2-High, 1-Urgent. Desc: 1-Urgent, 2-High, 3-Medium, 4-Low, 0-NoPriority. I suspect the database can only do asc/desc sort, and to get the desired behavior we need it to treat NoPriority as some MAX_INT instead of 0. Migration the NoPriority value from 0 to MAX_INT for the whole database and updating client code at the same time can be problematic I think. Besides, can be pointless if in some cases we'd still need the sorting behavior as if NoPriority=0, although NoPriority=MAX_INT indeed looks like more common case. What do you think?

Actually you are right, we use 0 as a value for NoPriority. I was wrong regarding null values. So, we cannot have good solution for showing NoPriority in the bottom without migration. As an alternative, users can create their own column with the same name and configure ordering as they want. We don't support icons for custom enums, though.

vsh1111:
Regarding migration. The challenge here is that the whole database needs to be migrated, plus client code be updated to use new values, plus maybe client local storage be migrated (not sure if you use one). All this has to be done in sync and such that the running clients with non-migrated code (users that keep Huly page open during database migration) don't see broken behavior. This probably would require a downtime for Huly.

I've seen some functions with "migration" in their names. Do you actually have a migration mechanism for cases such as this?

aonnikov:
> Regarding migration. The challenge here is that the whole database needs to be migrated, plus client code be updated to use new values, plus maybe client local storage be migrated (not sure if you use one). All this has to be done in sync and such that the running clients with non-migrated code (users that keep Huly page open during database migration) don't see broken behavior. This probably would require a downtime for Huly.
> 
> I've seen some functions with "migration" in their names. Do you actually have a migration mechanism for cases such as this?

Yes, we have mechanism of migrations, but it is not automatic, usually we have to write some code to migrate data. But this does not mean that we are going to do migration for issue priorities, current order is by design 😄

vsh1111:
I see. It looks like we've come to the conclusion of this issue and I'm closing this PR. Thank you.

vsh1111:
Looks like Mongo does have some custom sort capabilities. Something to look into.

PR #7216, opened 2024-11-21
Drawing overlay for image preview dialog

Chunosov:
Adds a draft feature for drawing over attached images; currently available only in the image preview dialog

Core drawing module is domain and (almost) ui-framework agnostic.

Only the Pen and Eraser tools are available for now, predefind color set, no pen width selector


models/attachment/src/index.ts
```diff
@@ -80,8 +82,19 @@ export class TSavedAttachments extends TPreference implements SavedAttachments {
...
 }
 
+@Model(attachment.class.Drawing, core.class.Doc, DOMAIN_ATTACHMENT)
+export class TDrawing extends TDoc implements Drawing {
+  @Prop(TypeRef(core.class.Doc), getEmbeddedLabel('Parent'))
+  @Index(IndexKind.Indexed)
+  @Hidden()
+    parent!: Ref<Doc>
+
+  @Prop(TypeString(), getEmbeddedLabel('Content'))
```
haiodo, line 97: If content is file please use Ref\<Blob> instead

packages/presentation/src/components/DrawingBoard.svelte
```diff
@@ -0,0 +1,147 @@
...
+  export let active = false
+  export let readonly = false
+  export let imageWidth: number | undefined
+  export let imageHeight: number | undefined
+  export let drawingData: DrawingData
+  export let saveDrawing: (data: any) => Promise<void>
+
+  let drawingTool: DrawingTool = 'pen'
+  let penColor = 'blue'
+  const penColors = ['red', 'green', 'blue', 'white', 'black']
```
haiodo, line 29: We have colors.ts with a lot of color constants, better to use colors from it.

models/attachment/src/index.ts
```diff
@@ -80,8 +82,19 @@ export class TSavedAttachments extends TPreference implements SavedAttachments {
   declare attachedTo: Ref<Attachment>
 }
 
+@Model(attachment.class.Drawing, core.class.Doc, DOMAIN_ATTACHMENT)
+export class TDrawing extends TDoc implements Drawing {
```
aonnikov, line 86: I think that eventually we will need separate plugin for drawings.

models/attachment/src/index.ts
```diff
@@ -80,8 +82,19 @@ export class TSavedAttachments extends TPreference implements SavedAttachments {
...
+  @Prop(TypeRef(core.class.Doc), getEmbeddedLabel('Parent'))
+  @Index(IndexKind.Indexed)
+  @Hidden()
+    parent!: Ref<Doc>
```
aonnikov, line 90: If you are planning to add drawing not only for Attachment, then you need to add parent class attribute.

packages/presentation/src/drawing.ts
```diff
@@ -0,0 +1,398 @@
...
+        draw.scale = {
+          x: canvas.width / entry.contentRect.width,
+          y: canvas.height / entry.contentRect.height
+        }
+      }
+    }
+  })
+  resizeObserver.observe(canvas)
+
+  canvas.onpointerdown = (e) => {
```
aonnikov, line 218: do we need to throttle this so we don't produce too many points for small changes?

packages/presentation/src/components/DrawingBoard.svelte
aonnikov: we don't have colors like wite or black in our palette, only some weird colors :)

packages/presentation/src/drawing.ts
Chunosov: Yes, we have that, see `minLineLength`. 
This also helps to make lines smoother. A bigger value makes lines even more smoother, but in the same time gives an impression of  like 'slow' lines. So this an empirical value based on my taste :) could'be audjusted

packages/presentation/src/components/DrawingBoard.svelte
Chunosov: This is just random start color set, they fit for any theme. I suppose, eventually there should be a color selector dialog and a dynamic color set of the most recently used colors.

models/attachment/src/index.ts
Chunosov: I think we decided on not to use files for drawings because they will only clutter the space. So this is not a reference to a file, but serialized drawing content. An object the drawing is attached to is in the `parent` field

models/attachment/src/index.ts
Chunosov: Yes, probably, when it grows. But semantically they are "attachments" because they don't make sence without a parent, so I decided to put them in the "attachment" plugin for now.

models/attachment/src/index.ts
Chunosov: Sure, added this as `parentClass` field

models/attachment/src/index.ts
haiodo: ok.

haiodo (approved):

packages/presentation/src/drawing.ts
aonnikov: Maybe you already have this, but to reduce this "slow" lines we can use the current cursor position as the last point (no need to save it until `minLineLength` or mouse up).

PR #7246, opened 2024-11-30
Fix hideAdd does not prevent adding by keyboard on TagsPopup

andr3h3nriqu3s11:
While looking to the code of `TagsPopup`, to make it work with keyboard only, I noticed that the `hideAdd` prop does not prevent the creating of a new tag via the keyboard.


aonnikov (approved):

aonnikov:
@haiodo any idea why we don't have "Approve and Run" button to trigger CI for this pull request?

PR #7247, opened 2024-12-01
Add next test button

ArtyomSavchenko:
Screenshot 2024-12-02 at 16 39 59


plugins/test-management-resources/src/components/test-result/NextButton.svelte
```diff
@@ -0,0 +1,84 @@
...
+
+  export let object: WithLookup<TestResult> | undefined
+  let isLoading = true
+  let hasNext = false
+  let navigateToNext = false
+
+  const client = getClient()
+  const hierarchy = client.getHierarchy()
+
+  testIteratorStore.subscribe(() => {
```
aonnikov: I feel like this may produce leak of subscribers. The `subscribe` returns unsubscribe function that should be called to unsubscribe.

haiodo (dismissed):

plugins/test-management-resources/src/components/test-result/NextButton.svelte
ArtyomSavchenko: Thank you, fixed

haiodo (approved):

PR #7253, opened 2024-12-02
Fix queries for $in: [null, ...etc] in pg

haiodo:
Fix queries for $in: [null, ...etc] in pg

### DIFF
--- a/server/postgres/src/storage.ts
+++ b/server/postgres/src/storage.ts
@@ -1137,7 +1137,16 @@ abstract class PostgresAdapterBase implements DbAdapter {
           case '$in':
             switch (type) {
               case 'common':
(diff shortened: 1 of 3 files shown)


server/postgres/src/storage.ts
```diff
@@ -1148,6 +1157,14 @@ abstract class PostgresAdapterBase implements DbAdapter {
...
           case '$nin':
+            if (Array.isArray(val) && val.includes(null)) {
+              res.push(
+                `(${tkey} NOT IN (${val
+                  .filter((it) => it != null)
+                  .map((v: any) => `'${v}'`)
+                  .join(', ')}) AND ${tkey} IS NOT NULL)`
+              )
+            }
             if (val.length > 0) {
```
aonnikov: should it be in `else` block?

haiodo: Yep, it shoudl be in else block. Thanks.

aonnikov (approved):

PR #7262, opened 2024-12-05
tool: copy files from R2 to datalake

aonnikov:
- add tool for copying files to datalake
- add multipart upload option

### DIFF
--- a/packages/presentation/src/file.ts
+++ b/packages/presentation/src/file.ts
@@ -258,8 +258,8 @@ async function uploadFileWithSignedUrl (file: File, uuid: string, uploadUrl: str
       method: 'PUT',
       headers: {
         'Content-Type': file.type,
-        'Content-Length': file.size.toString(),
-        'x-amz-meta-last-modified': file.lastModified.toString()
+        'Content-Length': file.size.toString()
+        // 'x-amz-meta-last-modified': file.lastModified.toString()
       }
     })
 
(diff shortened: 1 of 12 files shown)


packages/presentation/src/file.ts
```diff
@@ -258,8 +258,8 @@ async function uploadFileWithSignedUrl (file: File, uuid: string, uploadUrl: str
       method: 'PUT',
       headers: {
         'Content-Type': file.type,
-        'Content-Length': file.size.toString(),
-        'x-amz-meta-last-modified': file.lastModified.toString()
+        'Content-Length': file.size.toString()
+        // 'x-amz-meta-last-modified': file.lastModified.toString()
```
haiodo, line 262: Do we need this change?

server/datalake/src/client.ts
```diff
@@ -236,8 +274,8 @@ export class Client {
         method: 'PUT',
         headers: {
           'Content-Type': metadata.type,
-          'Content-Length': metadata.size?.toString() ?? '0',
-          'x-amz-meta-last-modified': metadata.lastModified.toString()
+          'Content-Length': metadata.size?.toString() ?? '0'
+          // 'x-amz-meta-last-modified': metadata.lastModified.toString()
```
haiodo, line 278: same?

haiodo (approved):

packages/presentation/src/file.ts
aonnikov: yes, this temporary fixes problem with presigned URL upload

server/datalake/src/client.ts
aonnikov: same

PR #7272, opened 2024-12-06
UBERF-8614: Mermaidjs integration

mr1name:
UBERF-8614: Mermaidjs integration


packages/theme/styles/prose.scss
```diff
@@ -381,19 +387,94 @@ pre.proseCodeBlock {
...
+  code {
+    cursor: text;
+  }
+
+  p div {
+    cursor: auto;
+  }
+}
+
+.proseMemraidDiagram {
```
aonnikov: typo

plugins/text-editor-resources/src/components/CollaborativeTextEditor.svelte
```diff
@@ -270,7 +271,8 @@
           { id: 'code-block', label: textEditor.string.CodeBlock, icon: view.icon.CodeBlock },
           { id: 'separator-line', label: textEditor.string.SeparatorLine, icon: view.icon.SeparatorLine },
           { id: 'todo-list', label: textEditor.string.TodoList, icon: view.icon.TodoList },
-          { id: 'drawing-board', label: textEditor.string.DrawingBoard, icon: IconScribble as any }
+          { id: 'drawing-board', label: textEditor.string.DrawingBoard, icon: IconScribble as any },
+          { id: 'mermaid', label: textEditor.string.MermaidDiargram, icon: view.icon.Model }
```
aonnikov, line 284: we need some icon

plugins/text-editor-resources/src/components/extension/mermaid.ts
```diff
@@ -0,0 +1,680 @@
...
+  doc.descendants((node, pos, parent, index) => {
+    if (node.type.name !== MermaidExtension.name) {
+      return
+    }
+
+    // See https://github.com/yjs/y-prosemirror/issues/49
+    // Ideally, one would hope that decorations would be recovered through the use of
+    // transaction mapping, but with Yjs this is not possible in many cases.
+    // So there's a need for dirty hacks to keep the state of the node intact. (See below).
+    const yjsdoc = options.ydoc?.getXmlFragment('content')
```
aonnikov: Is that still true and reproducible? We use decorations in several places but never noticed behavior like this, but this is probably because we recreate all decorations in this case instead of mapping them.

In either case, we will need some generic solution for this to use decorations properly, and ideally I would like to avoid use of ydoc here.

aonnikov: You need to pass field name along with ydoc, the field is not constant :(

mr1name: Yes, it's almost impossible to avoid. (The simplest scenario is to perform undo). Without it, the plugin will not work properly. But the dependency on YDoc is optional, if it is not there, the behavior will not change.

plugins/text-editor-resources/src/components/CollaborativeTextEditor.svelte
mr1name: This appears to be the most suitable from the ones lying around. I doubt a new icon will be available anytime soon.

packages/theme/styles/prose.scss
mr1name: Fixed.

plugins/text-editor-resources/src/components/extension/mermaid.ts
mr1name: Fixed.

aonnikov (dismissed):

aonnikov (approved):

Issue #6914, opened 2024-10-14
Bug: Opening document chat in sidebar causes freeze


AbigailDawson:
**Summary**

When you click the icon to open the chat for a document in the sidebar, the page freezes. Noticed on both front & app.

**Steps to reproduce:**

1. Login to front or app, navigate to chat
2. Select a document from the left sidebar
3. Click the icon to open the chat in the right sidebar

**Actual Result**

Page freezes

**Expected Result**

Chat should open in the right sidebar

<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzBkMDk3YjU2NTUyNjAyOTFmYWE5NGIiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIn0.V4w-eQzJ609zLrg87l5eRtEcWLVRkJmWY2VIIAhrfyk">Huly&reg;: <b>UBERF-8463</b></a></sub>

PR #7294, opened 2024-12-07
fix: fallback to defaultIcon when value is undefined

dakshesh14:
**Fix:**

- Fallback to default icon when value is undefined
- This change can avoid errors like #6914

Before:

[before-document-sidebar.webm](https://github.com/user-attachments/assets/d813442e-b670-4c07-97da-b4a6f9e3dc89)

After:

[after-document-sidebar.webm](https://github.com/user-attachments/assets/2dc07ed2-ba94-4aef-aff2-487fbe8c6783)

### DIFF
--- a/plugins/document-resources/src/components/DocumentIcon.svelte
+++ b/plugins/document-resources/src/components/DocumentIcon.svelte
@@ -20,7 +20,7 @@
 
   import { ComponentType } from 'svelte'
 
-  export let value: IconProps
+  export let value: IconProps | undefined
   export let size: IconSize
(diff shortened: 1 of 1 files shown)


dakshesh14:
The crash was caused by the `plugins/document-resources/src/components/DocumentIcon.svelte` component. After investigating its usage, I found that the `value` prop can potentially be undefined. To address this, I added a null check to handle cases where `value` is undefined, preventing the crash.

haiodo (approved):

haiodo:
Thanks @dakshesh14

PR #7424, opened 2024-12-10
fix: prevent content shift on select workspace page during loading

dakshesh14:
**Fix:**

- Added a loading state to the select workspace and username on the select workspace page.
- This change prevents content shift during data loading, improving user experience.

Before(delay added for demonstration):

[before-content-shift.webm](https://github.com/user-attachments/assets/6c527231-98a8-42cd-89fc-6635f612e50d)

After(delay added for demonstration):

[after-content-shift.webm](https://github.com/user-attachments/assets/2014c5c2-668d-4e18-80be-0d4fd96e3dc1)


haiodo (approved):

haiodo:
Thanks @dakshesh14

PR #7437, opened 2024-12-12
UBERF-8884 Revert datalake to Hyperdrive use

aonnikov:
As discussed, reverting to Hyperdrive use in datalake


haiodo (approved):

BykhovDenis:
maybe we can move it to env? Or `return postgres(env.DB_URL ?? env.HYPERDRIVE.connectionString,`

aonnikov:
Yeah, next time I need to change it I will do like this

PR #7473, opened 2024-12-15
Inline comments plugin

mr1name:
inline comments plugin


plugins/text-editor-resources/src/components/extension/inlineComment.ts
```diff
@@ -0,0 +1,984 @@
...
+function getMeta (tx?: Transaction): MetaPatch | undefined {
+  return tx?.getMeta('inline-comment')
+}
+
+function setMeta (tx: Transaction, meta: MetaPatch): Transaction {
+  return tx.setMeta('inline-comment', meta).setMeta('addToHistory', false)
+}
+
+function getYDocCommentMap (options: InlineCommentExtensionOptions): YMap<InlineComment> {
+  return options.ydoc.getMap(options.ydocExtensionField ?? 'inline-comments')
```
aonnikov, line 76: this won't work :( we currently store only one field from ydoc. we will need to do a bit more changes to allow to store additional information

plugins/text-editor-resources/src/components/extension/inlineComment.ts
```diff
@@ -0,0 +1,984 @@
...
+        this.destructors.push(() => this.container?.removeEventListener(listener[0], listener[1]))
+      }
+
+      this.destructors.push(
+        observeSizeChanges(this.container, () => {
+          forceUpdateDecorations(view)
+        })
+      )
+
+      this.tippynode = tippy(view.dom, {
```
aonnikov, line 634: can we do this without tippy on our tooltips only?

mr1name: It would be easier to write a new tooltip system from scratch. Unfortunately the current system can't help here in any meaningful way, whereas popperjs and tippy are shipped with tiptap and already used for the editor menus.

plugins/text-editor-resources/src/components/extension/inlineComment.ts
aonnikov: as discussed on the call, keep this as is for now

plugins/text-editor-resources/src/components/extension/inlineComment.ts
```diff
@@ -0,0 +1,1004 @@
...
+    }
+  ]
+
+  const fetchComponent = async (): Promise<void> => {
+    // In the current package structure, it is practically impossible to import anything from here
+    // without introducing cyclic dependencies, since text-editor-resources is in the dependency
+    // hierarchy of almost all other packages.
+    // In the future, it might be worth considering putting text editor plugins in the text-editor-plugins
+    // package so that one can create dependencies on packages such as activity and chunter.
+    const component = await getResource('chunter:component:InlineCommentThread' as AnyComponent)
```
aonnikov: There should be no problem depending on `chunter` plugin as it does not depend on `text-editor-resources`

mr1name: The comment was more about being able to create a Svelte component for comments within the comments plugin itself. So far I've had to put it in chunter-resources and grab it via dynamic import. I may add an import of the chunter itself for ref typing separately (do we want chunter dependency here?), but I'm not sure if that would be highly relevant to the original problem.

aonnikov (approved):

PR #7490, opened 2024-12-17
Add initial transactor rpc

Chunosov:
<sub><a href="https://front.hc.engineering/guest/platform?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzYxM2M3ZGMyOWE3OGFmZmY2NmQ1YWEiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InBsYXRmb3JtIn0.jQsa8zL3pTEjDNndRQsqZWVc64HZZcmDFRGVWZcleBQ">Huly&reg;: <b>UBERF-8958</b></a></sub>


workers/transactor/src/index.ts
```diff
@@ -48,3 +49,26 @@ export default {
     return await router.fetch(request).catch(error)
   }
 } satisfies ExportedHandler<Env>
+
+export class TransactorRPC extends WorkerEntrypoint<Env> {
+  async rpcFindAll (
+    token: string,
```
haiodo: Please add workspaceId here, so could think about security at next step

aonnikov: I don't like that we have to pass token and workspaceId fro every request. I propose to use stubs to hide this.
In the transactor there will be only one method, named `connect` (or any other name) that will return stub with two methods: find and tx.

https://developers.cloudflare.com/workers/runtime-apis/rpc/
```
using client = env.TRANSACTOR.connect(token, workspaceId)
await client.tx(...)
```

workers/transactor-api/src/index.ts
```diff
@@ -0,0 +1,138 @@
...
+  const model = JSON.parse(jsonString) as Tx[]
+  return model
+}
+
+export class TransactorRpcClient implements TransactorClient {
+  private transactorRpcStub: TransactorRawApi | undefined
+  private model: ModelDb | undefined
+
+  constructor (
+    private readonly ctx: MeasureContext,
```
aonnikov: for end users ctx should be not require in the constructor, either create it when it needed or make it optional

workers/transactor-api/src/index.ts
```diff
@@ -0,0 +1,138 @@
...
+
+interface TransactorRawApi {
+  findAll: (_class: Ref<Class<Doc>>, query?: DocumentQuery<Doc>, options?: FindOptions<Doc>) => Promise<FindResult<Doc>>
+
+  tx: (tx: Tx) => Promise<TxResult>
+
+  getModel: () => Promise<Buffer>
+}
+
+export interface TransactorClient {
```
aonnikov: It should be the same as core client interface

workers/transactor-api-demo/src/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    const router = Router()
+    router
+
+      .get('/demo-find-raw/:email/:password/:workspace', async ({ params }) => {
+        console.log('REQ', params)
+        const client = await transactorService.openRpc(await getToken(params), params.workspace)
+        try {
+          const result = await client.findAll('contact:class:Person' as Ref<Class<Doc>>)
```
aonnikov: how about `tx`? 😄 we should be able to create a doc, and for this we will need to use `TxOperations`

aonnikov (dismissed):

aonnikov (dismissed):

haiodo (approved):

PR #7585, opened 2025-01-04
create pl.json

igorsynak:
polish translate. I don't know the structure of the project but I hope that when I send the translated .json file, you will add d;


haiodo:
Hi @igorsynak It is a good start, it is more a more json.files they are all in lang folders. 
Also could you please try to add a flag into `packages/ui/src/components/internal/LangSelector.svelte` Without flag Huly will not be able to understand your translation files.
Thanks.

igorsynak:
okay ,
    { id: 'pl', label: ui.string.Polish, logo: '&#x1F1F5;&#x1F1F1;' }

igorsynak:
I don't know how to edit the code when I already have a pull request, but it's 1 line, so I send it to d;

PR #7624, opened 2025-01-09
Text & Table cell color styling features

mr1name:


plugins/text-editor-resources/src/components/extension/colors.ts
```diff
@@ -0,0 +1,101 @@
...
+    colorSpec('green'),
+    colorSpec('blue'),
+    colorSpec('purple'),
+    colorSpec('pink'),
+    colorSpec('red')
+  ]
+}
+
+export async function openBackgroundColorOptions (editor: Editor, event: MouseEvent): Promise<void> {
+  console.log(palette)
```
haiodo:
```suggestion
```

plugins/text-editor-resources/src/components/extension/colors.ts
```diff
@@ -0,0 +1,101 @@
...
+
+function colorSpec (tag: string, prefix = 'text'): ColorSpec {
+  const color = colorVar(tag, prefix)
+  return { color, preview: colorVar(tag) }
+}
+
+const palette = {
+  background: [
+    { color: 'transparent' },
+    colorSpec('gray', 'bg'),
```
haiodo, line 42: May be we could use our palette? colors.ts?

mr1name: There are a few problems with this. Both palettes (in ui/src/colors.ts) are oriented to the same constraints (the same colors are used there for both themes), so black text will not be readable, besides that they are just too vibrant to be used in tables even for a black theme. We would also need a different rendering mechanism for these extensions to work with data represented in this way. Right now, the colors that are provided there are linked to CSS variable names, so we can update the palettes if needed in the future when they are adjusted for the needs of these contexts.

haiodo (approved):

PR #7629, opened 2025-01-10
UBERF-9099: Rate limits

haiodo:
UBERF-9099: Rate limits

+ Fix Github comment with Link appear on restart

### DIFF
--- a/packages/core/src/__tests__/limits.test.ts
+++ b/packages/core/src/__tests__/limits.test.ts
@@ -1,0 +1,118 @@
+//
+// Copyright © 2023 Hardcore Engineering Inc.
+//
+// Licensed under the Eclipse Public License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


server/server/src/client.ts
```diff
@@ -154,7 +158,9 @@ export class ClientSession implements Session {
     query: DocumentQuery<T>,
     options?: FindOptions<T>
   ): Promise<void> {
-    await ctx.sendResponse(ctx.requestId, await this.findAllRaw(ctx.ctx, ctx.pipeline, _class, query, options))
+    await this.findRateLimit.exec(async () => {
```
BykhovDenis: Why we need it here? It's really bad....

haiodo: Why bad? To prevent ddos by individual connection.

BykhovDenis (approved):

PR #7631, opened 2025-01-10
EZQMS-1317: authors signature on review and approval request & block review bypass

mr1name:
Note: review bypass is blocked only if the current non-effective version has been reviewed at least once already

### DIFF
--- a/plugins/controlled-documents-resources/src/stores/editors/document/canSendForApproval.ts
+++ b/plugins/controlled-documents-resources/src/stores/editors/document/canSendForApproval.ts
@@ -29,7 +29,7 @@ export const $canSendForApproval = combine(
   (document, isLatestVersion, state, comments, training, reviewHistory) => {
     let haveBeenReviewedOnce = false
     if (document !== null) {
-      const reviews = (reviewHistory ?? []).filter(review => review.attachedTo === document._id)
+      const reviews = (reviewHistory ?? []).filter((review) => review.attachedTo === document._id)
       if (reviews.length > 0) haveBeenReviewedOnce = true
     }
     return (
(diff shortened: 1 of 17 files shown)


plugins/controlled-documents-resources/src/stores/editors/document/canSendForApproval.ts
```diff
@@ -17,17 +17,24 @@ import { ControlledDocumentState, DocumentState } from '@hcengineering/controlle
...
   $isLatestVersion,
   $documentState,
   $documentComments,
   $training,
-  (isLatestVersion, state, comments, training) => {
+  $reviewRequestHistory,
+  (document, isLatestVersion, state, comments, training, reviewHistory) => {
+    let haveBeenReviewedOnce = false
+    if (document !== null) {
+      const reviews = (reviewHistory ?? []).filter((review) => review.attachedTo === document._id)
```
lexiv0re, line 32: Aren't you fetching only review requests attached to this document? Is additional filtering needed?

mr1name: Yes, the effect doesn't always trigger and update the state - this also applies to other requests that are already there. I haven't investigated the cause yet, but it should probably be considered as a separate bug.

lexiv0re:
@mr1name there's an additional requirement to provide this new information on `Review and Approvals` right tab of a document.

mr1name:
> @mr1name there's an additional requirement to provide this new information on `Review and Approvals` right tab of a document.

Can you clarify what that would mean in this context? Since there are no new information really whatsoever

lexiv0re:
It's about showing it in this panel. We can probably discuss the UI later and do it in a separate PR.
image

lexiv0re:
The simplest thing would probably be just to show an additional "row" but with "Author" in place of "Review/Approval"

lexiv0re (approved):

mr1name:
> We can probably discuss the UI later and do it in a separate PR

Sure, let's discuss it next week.

PR #7662, opened 2025-01-14
UBERF-9107: Add backup list support

haiodo:
UBERF-9107: Add backup list support

Allow to expose backup to owners

### DIFF
--- a/plugins/setting-resources/src/components/Backup.svelte
+++ b/plugins/setting-resources/src/components/Backup.svelte
@@ -1,0 +1,355 @@
+<!--
+// Copyright © 2025 Hardcore Engineering Inc.
+//
+// Licensed under the Eclipse Public License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License. You may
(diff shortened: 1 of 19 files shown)


plugins/setting-resources/src/components/Backup.svelte
```diff
@@ -0,0 +1,355 @@
...
+  }
+  | undefined
+
+  $: fileSizes = new Map(backupInfo?.files?.map((it) => [it.name, it.size ?? 0]))
+
+  function formatDate (timestamp: number): string {
+    return new Date(timestamp).toLocaleString()
+  }
+
+  function getSize (size: number): string {
```
aonnikov, line 47: We have an utility for this somewhere, used in Drive

aonnikov (approved):

plugins/setting-resources/src/components/Backup.svelte
haiodo: Let it be for now.

PR #7671, opened 2025-01-15
fix: add mermaid plugin to server kit

aonnikov:


packages/presentation/src/components/markup/NodeContent.svelte
```diff
@@ -186,6 +186,8 @@
         {/each}
       {/if}
     </th>
+  {:else if node.type === MarkupNodeType.mermaid}
+    <!-- TODO -->
```
aonnikov, line 190: will be done in separate task

haiodo (approved):

PR #7672, opened 2025-01-15
CopyTextToClipboard fixed for mobile browsers

SasLord:


packages/ui/src/components/internal/Root.svelte
```diff
@@ -141,6 +142,22 @@
     $deviceInfo.isMobile &&
     (($deviceInfo.isPortrait && $deviceInfo.docWidth <= 480) ||
       (!$deviceInfo.isPortrait && $deviceInfo.docHeight <= 480))
+
+  // Copying text to the clipboard (alternative method for mobile devices)
+  $: if ($clipboardText !== '') {
```
haiodo: Let's move this code just in regular function in utility's, do not put it here, also you should use it only if regular copy is not working check.

haiodo:
да пусть поболтается

haiodo (approved):

PR #7691, opened 2025-01-16
EZQMS-1234: means for transferring controlled documents between spaces

mr1name:
EZQMS-1234: means for transferring controlled documents between spaces


plugins/controlled-documents/src/utils.ts
```diff
@@ -129,6 +138,363 @@ export async function deleteProjectDrafts (client: ApplyOperations, source: Ref<
...
+      })
+    }
+
+    for (const m of bundle.ProjectMeta) update(m, { project })
+    for (const m of bundle.ProjectDocument) update(m, { project })
+  }
+
+  if (mode === 'check') return true
+
+  const promises = Array.from(updates).map((u) => client.update(u[0], u[1]))
```
BykhovDenis: I think better use `client.apply`

plugins/controlled-documents-assets/lang/ru.json
```diff
@@ -295,7 +295,12 @@
     "DeleteDocumentCategoryPermission": "Удалять категорию",
     "DeleteDocumentCategoryDescription": "Предоставляет пользователям разрешение удалять категорию",
     "ConfigLabel": "Управляемые Документы",
-    "ConfigDescription": "Расширение для управления управляемыми документами"
+    "ConfigDescription": "Расширение для управления управляемыми документами",
```
lexiv0re, line 298: What about other languages?

plugins/controlled-documents-assets/lang/en.json
```diff
@@ -295,7 +295,12 @@
     "DeleteDocumentCategoryPermission": "Delete document category",
     "DeleteDocumentCategoryDescription": "Grants users ability to delete a document category",
     "ConfigLabel": "Controlled Documents",
-    "ConfigDescription": "Extension to manage controlled documents"
+    "ConfigDescription": "Extension to manage controlled documents",
+
+    "Transfer": "Transfer",
+    "TransferWarning": "You and your team members may not be able to edit this document after this action.",
```
lexiv0re: We should be checking that the guy who moves a document is a member and can create new documents in the target space so this warning it not valid. Should be something like:
"Some of the document's team members may not be able to view or edit this document after this action."

plugins/controlled-documents/src/utils.ts
```diff
@@ -129,6 +138,364 @@ export async function deleteProjectDrafts (client: ApplyOperations, source: Ref<
...
+  ControlledDocument: ControlledDocument[]
+  ChangeControl: ChangeControl[]
+  DocumentRequest: DocumentRequest[]
+  DocumentSnapshot: DocumentSnapshot[]
+  ChatMessage: ChatMessage[]
+  TagReference: TagReference[]
+  Attachment: Attachment[]
+}
+
+function emptyBundle (): DocumentBundle {
```
lexiv0re, line 204: Do all these objects live in a documents space??

mr1name: Yes

mr1name: Some of them in different domain, but they are still attached to the document space

plugins/controlled-documents-assets/lang/ru.json
mr1name: What do you mean? There are eng/rus/fr atm.

lexiv0re: Despite that cloud TraceX shows only these three languages let's keep all the languages translations up-to-date. There might be self-hosted installations of TraceX that use different languages.

mr1name: np, fixed

plugins/controlled-documents-assets/lang/en.json
mr1name: updated

lexiv0re (approved):

PR #7739, opened 2025-01-21
EQMS-1414 Fix pod print issues

aonnikov:
1. Fixes regression in print container caused by base image
2. More proper error handling, show error instead of loading UI in preview
3. Report to analytics

### DIFF
--- a/dev/base-image/print.Dockerfile
+++ b/dev/base-image/print.Dockerfile
@@ -1,5 +1,9 @@
 FROM hardcoreeng/base
 
+# Chromium hangs when usging LD_PRELOAD and MALLOC_CONF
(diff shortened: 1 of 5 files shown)


dev/base-image/print.Dockerfile
```diff
@@ -1,5 +1,9 @@
 FROM hardcoreeng/base
 
+# Chromium hangs when usging LD_PRELOAD and MALLOC_CONF
```
haiodo, line 3: please override in place where it is used, not here.

haiodo (changes requested):

haiodo (approved):

PR #7760, opened 2025-01-22
UBERF-9230: Fix ses webpush

haiodo:
UBERF-9230: Fix ses webpush

### DIFF
--- a/services/ses/pod-ses/src/main.ts
+++ b/services/ses/pod-ses/src/main.ts
@@ -13,6 +13,7 @@
 // limitations under the License.
 //
 
+import type { Ref } from '@hcengineering/core'
 import { PushSubscription, type PushData } from '@hcengineering/notification'
 import type { Request, Response } from 'express'
 import webpush, { WebPushError } from 'web-push'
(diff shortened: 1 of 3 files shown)


kristina-fefelova (dismissed):

services/ses/pod-ses/src/main.ts
```diff
@@ -22,25 +23,39 @@ import { SES } from './ses'
...
-      if (errorMessages.some((p) => JSON.stringify((err as WebPushError).body).includes(p))) {
-        return 'clear-push'
+async function sendPushToSubscription (
+  subscriptions: PushSubscription[],
+  data: PushData
+): Promise<Ref<PushSubscription>[]> {
+  const result: Ref<PushSubscription>[] = []
+  for (const subscription of subscriptions) {
+    try {
+      await webpush.sendNotification(subscription, JSON.stringify(data))
```
aonnikov, line 33: can we send them in parallel?

aonnikov (approved):

services/ses/pod-ses/src/main.ts
haiodo: It is for one user, I think no need, we send in parallel for different users.

Issue #7749, opened 2025-01-21
Set first day of the week


AbigailDawson:
Request from Huly users to be able to set Sunday as the first day of the week (instead of Monday) for the whole workspace.

Link to community message: <https://hulycommunity.slack.com/archives/C06BJD10YBC/p1736689787929139>

PR #7770, opened 2025-01-23
UBERF-9220: Set first day of the week

SasLord:
screen-eng
screen-2
screen-1
screen-3
screen-4

Fixed: UBERF-9220 (#7749)

### DIFF
--- a/packages/ui/src/components/internal/Root.svelte
+++ b/packages/ui/src/components/internal/Root.svelte
@@ -141,6 +141,13 @@
     $deviceInfo.isMobile &&
     (($deviceInfo.isPortrait && $deviceInfo.docWidth <= 480) ||
       (!$deviceInfo.isPortrait && $deviceInfo.docHeight <= 480))
+
+  $deviceInfo.mondayStart = localStorage.getItem('mondayStart') === 'true' ?? true
+  let oldmondayStart: boolean = $deviceInfo.mondayStart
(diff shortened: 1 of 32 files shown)


packages/ui/src/components/internal/Root.svelte
```diff
@@ -141,6 +141,13 @@
     $deviceInfo.isMobile &&
     (($deviceInfo.isPortrait && $deviceInfo.docWidth <= 480) ||
       (!$deviceInfo.isPortrait && $deviceInfo.docHeight <= 480))
+
+  $deviceInfo.mondayStart = localStorage.getItem('mondayStart') === 'true' ?? true
```
kristina-fefelova: Maybe we should save `mondayStart` in personal `Preferences` instead of  `localStorage`?

aonnikov:
Can you group settings  - delete workspace button should be put near to workspace name, not after calendar.

packages/ui/src/components/internal/Root.svelte
aonnikov: Do we need this setting at all? Is that possible to figure out by locale?

kristina-fefelova: It seems that the best option is to get default by locale + setting if someone's preferences are different

haiodo (dismissed):

haiodo (approved):

PR #7789, opened 2025-01-24
Fixed validation of class inheritance in queries

mr1name:
The original source of the problem is [here](https://github.com/hcengineering/platform/blob/0958c241cdbc43a2972b3f3c84dac7ab3fb4c6fb/server/mongo/src/storage.ts#L930), but we should probably do the check in the matchQuery itself

### DIFF
--- a/packages/core/src/query.ts
+++ b/packages/core/src/query.ts
@@ -156,7 +156,8 @@ export function matchQuery<T extends Doc> (
   hierarchy: Hierarchy,
   skipLookup: boolean = false
 ): Doc[] {
-  let result = docs
+  const baseClass = hierarchy.getBaseClass(clazz)
+  let result = docs.filter((r) => hierarchy.isDerived(r._class, baseClass))
   for (const key in query) {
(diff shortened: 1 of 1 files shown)


packages/core/src/query.ts
```diff
@@ -156,7 +156,8 @@ export function matchQuery<T extends Doc> (
   hierarchy: Hierarchy,
   skipLookup: boolean = false
 ): Doc[] {
-  let result = docs
+  const baseClass = hierarchy.getBaseClass(clazz)
+  let result = docs.filter((r) => hierarchy.isDerived(r._class, baseClass))
```
BykhovDenis, line 160:
```suggestion
  let result = docs.filter((r) => hierarchy.isDerived(r._class, baseClass))
  if (baseClass !== clazz) {
    result = docs.filter((r) => hierarchy.hasMixin(r, clazz))
  }
```

BykhovDenis: Let's also check mixin

BykhovDenis (approved):

PR #7799, opened 2025-01-27
UBERF-9264: Fix set/unset parent issue

haiodo:
UBERF-9264: Fix set/unset parent issue

image

image


plugins/tracker-resources/src/components/issues/edit/EditIssue.svelte
```diff
@@ -276,8 +283,22 @@
     </svelte:fragment>
 
     {#if hasParentIssue}
-      <div class="mb-6">
+      <div class="mb-6 flex-row-center">
         <SubIssueSelector {issue} />
+        {#if hasParentIssue}
```
BykhovDenis: Extra check, we have the same in 285

BykhovDenis: Maybe we should hide or disable button if readonly?

BykhovDenis (dismissed):

plugins/tracker-resources/src/components/issues/edit/EditIssue.svelte
haiodo: Thanks, Fixed.

BykhovDenis (approved):

PR #7807, opened 2025-01-28
UBERF-9262: hide document activity & inline comments for guests & PDFs

mr1name:

### DIFF
--- a/plugins/text-editor-resources/src/components/CollaborativeTextEditor.svelte
+++ b/plugins/text-editor-resources/src/components/CollaborativeTextEditor.svelte
@@ -17,23 +17,25 @@
 <script lang="ts">
   import { Analytics } from '@hcengineering/analytics'
   import {
+    AccountRole,
     type Blob,
     Class,
     type CollaborativeDoc,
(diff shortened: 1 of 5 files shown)


plugins/text-editor-resources/src/components/CollaborativeTextEditor.svelte
```diff
@@ -433,7 +438,7 @@
   onMount(async () => {
     await ph
 
-    if (enableInlineComments) {
+    if (enableInlineComments && !isGuest) {
```
lexiv0re, line 441: Is it safe to not add this extension at all rather than not to display anything if it's a guest? Won't there be any exceptions for unknown marks in the doc?

mr1name: Recently, the comments plugin has been split into UI and Core parts. Core part is always loaded. And this plugin is only for decoration and integration with collaborative features.
https://github.com/hcengineering/platform/pull/7699

lexiv0re (approved):

PR #7834, opened 2025-01-29
EZQMS-1268: Doc ID changes editable phases

mr1name:
Turns out this one was already implemented a long time ago by @lexiv0re

### DIFF
--- a/plugins/controlled-documents-resources/src/components/document/popups/ChangeCategoryPopup.svelte
+++ b/plugins/controlled-documents-resources/src/components/document/popups/ChangeCategoryPopup.svelte
@@ -1,0 +1,48 @@
+<!--
+// Copyright © 2025 Hardcore Engineering Inc.
+//
+// Licensed under the Eclipse Public License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License. You may
+// obtain a copy of the License at https://www.eclipse.org/legal/epl-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+-->
+
+<script lang="ts">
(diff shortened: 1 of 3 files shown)


plugins/controlled-documents-resources/src/components/document/right-panel/DocumentInfoTab.svelte
```diff
@@ -59,7 +59,7 @@
   }
 
   $: isDocCodeEditable =
-    $isEditable && $controlledDocument != null && $controlledDocument.major === 0 && $controlledDocument.minor === 0
+    $isEditable && $controlledDocument != null && $controlledDocument.major === 0 && $controlledDocument.minor === 1
```
lexiv0re: 1. Actually, it's not that simple. Based on the selector in the Release tab the first draft might be either v0.1 or v1.0. But, v1.0 might also be not the first version of the document. E.g. v0.1 -> v0.2 -> v1.0.
So we need another condition to check if it's the very first draft. E.g. check document's meta documents collection field for the number of versions.
2. What about changing category for a template, is it also implemented?

mr1name: fixed

plugins/controlled-documents-resources/src/components/document/popups/ChangeCategoryPopup.svelte
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+-->
+
+<script lang="ts">
+  import documents, { DocumentCategory, DocumentTemplate } from '@hcengineering/controlled-documents'
+  import { createQuery, getClient } from '@hcengineering/presentation'
+  import { createEventDispatcher } from 'svelte'
+
+  import { Ref } from '@hcengineering/core'
+  import DropdownLabelsPopup from '@hcengineering/ui/src/components/DropdownLabelsPopup.svelte'
```
lexiv0re: Bad import

mr1name: fixed

lexiv0re (approved):

PR #7854, opened 2025-01-31
Additional tweaks for EQMS-1317

mr1name:


lexiv0re:
@mr1name could you add screenshots of how it looks for various cases:
1. only review
2. only approval
3. review + approval
4. review -> draft -> review
5. review -> draft -> review -> approval

mr1name:
It's identical in all these cases just that each review or approval block now has a row with the author and the date (“signature”) of initiation.

Screenshot 2025-01-31 at 15 08 01

lexiv0re:
@mr1name I think what we wanted is a new standalone block so it'd look like

_________
**Approval**
John
_________
**Review**
Jack
_________
**Author**
Jane
_________

mr1name:
@lexiv0re Why? These are foldable blocks that refer to a whole review or approval request. What's the point of making it a separate block? This would simply imply duplicating all blocks with the same meaning in the general case.

lexiv0re:
@mr1name the point is to match the expectations where one would like to see three independent signatures for a document from:
1. Author
2. Reviewer
3. Approver

The idea is that the author signs the document itself, not a review/approval request. Theoretically, the author's signature could live on it's own w/o any reviews/approvals:
1. Author drafts a document and signs it. -> The signature is there but there's no review/approval yet.
2. He gives it to a reviewer/approver. -> Only at this point the review/approval process is started.

It's just in the system we combine the actions of signing by the author and sending to review/approval.

lexiv0re:
At least that was my understanding. We can check with Charles to clarify this before proceeding in any direction.

charles-rollet:
Hey @lexiv0re  & @mr1name to clarify, what we want is to be sure that the signature orders are respected on the current effective version (called "current version" on the right hand side panel).  
- If there is no reviewer, the order should be: author > approver > on the approved and then effective doc
- if there is a reviewer, the order should be: author > reviewer > approver > on the approved and then effective doc

So, If I take an example (in prod) of a doc that was reviewed and then re-sent for review and then approval, and describe what should be expected regarding the discussion point: 
Capture d’écran 2025-02-02 à 12 33 29

So, 
i) merging the "draft revision 1" signature records and "current version" signatures records within the same block 
ii)  in line with the block name e.g. "current version", so removing the names "approval", "review" as then, do not make sense anymore
iii) So, then putting the role of the signers (author, reviewer, approver) in line with there user names (line in the @mr1name screenshot above)
iv) for "Draft revision 1" - having listed: author + reviewer signatures
v) for the "current  revision" - having listed:  author + reviewer + approver signatures

and so, if: 
i) there are multiple revisions >>  to show them with multiple blocks
ii) there are multiple reviewers, approvers >> to show them
iii) if there is no reviewer >> to no show them


AND, the author, reviewer, approver timestamp in the PDF are from the "current version"

Hope this clarify - if not we can discuss it together ;-) 

(Also, if this implementation is too complexe with current architecture, we can of course discuss it ;-)  )

mr1name:
Redid the UI based on @charles-rollet comment and rewrote the signature formation logic - it's now the same for both the PDF and the Validation Workflow tab. 

Screenshot 2025-02-06 at 20 01 04

lexiv0re (dismissed):

lexiv0re (approved):

PR #7938, opened 2025-02-05
Fixes for document hierarchy presentation in QMS

mr1name:


plugins/controlled-documents-resources/src/utils.ts
```diff
@@ -867,3 +870,53 @@ export async function moveDocumentAfter (doc: ProjectMeta, after: ProjectMeta):
 
   await client.update(doc, { parent, path, rank })
 }
+
+export class DocumentHiearchyQuery {
+  queries = {
+    ProjectMeta: createQuery(),
```
lexiv0re: Minor: start with lower case?

mr1name: idk, it's mapping to DocumentBundle fields, which maps fields to class names. If that's the only change, I probably wouldn't waste CI on it 😄

lexiv0re: So it matches the class or what's the point? :)

lexiv0re: Still checking the rest

lexiv0re:
Margins look bad. Please fix them as well.
Current:
image

Supposed to be:
image

lexiv0re:
The rest LGTM

lexiv0re (approved):

PR #7947, opened 2025-02-06
Track drawing board movements via presence service

Chunosov:
Related PR in Presence Service: https://github.com/hcengineering/huly-cloud/pull/39

### DIFF
--- a/plugins/presence/package.json
+++ b/plugins/presence/package.json
@@ -38,6 +38,7 @@
   },
   "dependencies": {
     "@hcengineering/platform": "^0.6.11",
+    "@hcengineering/contact": "^0.6.24",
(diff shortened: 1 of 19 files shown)


plugins/presence/package.json
```diff
@@ -38,6 +38,7 @@
   },
   "dependencies": {
     "@hcengineering/platform": "^0.6.11",
+    "@hcengineering/contact": "^0.6.24",
```
haiodo, line 41: why we need it? @lexiv0re did we have some interface in class to use instead?

lexiv0re: Nope, if an internal `Person` is needed there's no other interface for that.

aonnikov (approved):

PR #7949, opened 2025-02-06
Add communication api on server and add separate client

kristina-fefelova:


packages/presentation/src/communication.ts
```diff
@@ -0,0 +1,267 @@
...
+  async findNotifications (params: FindNotificationsParams, queryId?: number): Promise<Notification[]> {
+    return await this.connection.findNotifications(params, queryId)
+  }
+
+  async unsubscribeQuery (id: number): Promise<void> {
+    await this.connection.unsubscribeQuery(id)
+  }
+
+  close (): void {
+    // do nothing
```
ArtyomSavchenko, line 214: Shouldn't the connection be closed here?
`this.connection.close()`

ArtyomSavchenko: Looks like communication client uses connection from other client, so it shouldn't be closed here

aonnikov (approved):

haiodo (approved):

PR #7952, opened 2025-02-06
UBERF-9390: Fix infinite loop

akhismat:


BykhovDenis (changes requested):
Why? I mean no class should extends itself or loop, it should be impossible

BykhovDenis (approved):

PR #7969, opened 2025-02-10
UBERF-9455: Fix change of configurations and proper notifyTx

haiodo:
UBERF-9455: Fix change of configurations and proper notifyTx

### DIFF
--- a/packages/presentation/src/configuration.ts
+++ b/packages/presentation/src/configuration.ts
@@ -13,7 +13,7 @@
 // limitations under the License.
 //
 
-import core, { SortingOrder, type PluginConfiguration, type TxUpdateDoc } from '@hcengineering/core'
+import core, { SortingOrder, type PluginConfiguration, type Tx, type TxUpdateDoc } from '@hcengineering/core'
 import { getResourcePlugin, type Plugin, type Resource } from '@hcengineering/platform'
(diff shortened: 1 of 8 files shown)


server/middleware/src/model.ts
```diff
@@ -74,8 +67,7 @@ export class ModelMiddleware extends BaseMiddleware implements Middleware {
 
   @withContext('get-model')
   async getUserTx (ctx: MeasureContext, txAdapter: TxAdapter): Promise<Tx[]> {
-    const allUserTxes = await ctx.with('fetch-model', {}, (ctx) => txAdapter.getModel(ctx))
-    return allUserTxes.filter((it) => isUserTx(it))
```
lexiv0re: Won't it throw somewhere w/o filtering? `contact:class:PersonAccount` entries are still in the model but hierarchy doesn't know about them anymore.

haiodo: Hmps, let me return, may be

haiodo: Nope, it will not, it was used before to filter out system transactions, but it doesn't required now.

BykhovDenis (dismissed):

aonnikov (dismissed):

kristina-fefelova (approved):

PR #7976, opened 2025-02-10
UBERF-9447: Move accounts to CR tool

lexiv0re:
Related to: https://front.hc.engineering/workbench/platform/tracker/UBERF-9447

### DIFF
--- a/dev/tool/src/db.ts
+++ b/dev/tool/src/db.ts
@@ -1,5 +1,12 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import { type AccountDB, type Workspace, getAccount, getWorkspaceById } from '@hcengineering/account'
+import {
+  type AccountDB,
+  type MongoAccountDB,
+  type Workspace,
+  getAccount,
+  getWorkspaceById,
+  getWorkspaces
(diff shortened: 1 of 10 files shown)


dev/tool/src/db.ts
```diff
@@ -174,191 +181,193 @@ export async function moveAccountDbFromMongoToPG (
...
-export async function updateDataWorkspaceIdToUuid (
-  ctx: MeasureMetricsContext,
-  accountDb: AccountDB,
-  dbUrl: string | undefined,
-  dryRun = false
-): Promise<void> {
-  if (dbUrl === undefined) {
-    throw new Error('dbUrl is required')
+        if (person.lastName == null) {
+          person.lastName = 'n/a'
```
haiodo, line 201: may be we should let it be empty?

haiodo (approved):

dev/tool/src/db.ts
lexiv0re: We could, I wasn't sure if keeping it empty is a good thing or we should highlight the issue with `n/a`. I thought with both first and last name empty there might be some weird issues that might be hard to see.

PR #8006, opened 2025-02-13
Automatic link to ref conversion

mr1name:


haiodo:
@mr1name Let's add few UI tests.

plugins/text-editor-resources/src/components/extension/link.ts
```diff
@@ -39,5 +45,156 @@ export const LinkShortcutsExtension = Extension.create<any>({
...
+        return {
+          references: prev.references,
+          queue: new Set([...prev.queue, ...urls])
+        }
+      }
+    }
+  })
+}
+
+async function getReferenceFromUrl (text: string): Promise<ReferenceProps | undefined> {
```
aonnikov, line 147: Probably not needed to fix right now, but I would move this logic somewhere to an utility function or reused existing (if any) to ensure we keep consistent logic. For example, we already have references extension that do similar things.

mr1name: The thing is, there's no good place to put it at the moment, and there's no existing equivalent alternative. The only thing that knows how to parse links is the Workbench svelte component. It should probably be moved somewhere else, but in the Workbench refactoring context.

mr1name:
@haiodo added one simple test, not sure if something more sophisticated is necessary.

haiodo (approved):

PR #8026, opened 2025-02-17
QFIX: Fix missing SES_AUTH_TOKEN

haiodo:
QFIX: Fix missing SES_AUTH_TOKEN


server/account/src/utils.ts
```diff
@@ -390,7 +391,8 @@ export async function sendOtpEmail (
   await fetch(concatLink(sesURL, '/send'), {
```
lexiv0re, line 391: There are 2 more methods in `operations.ts` where emails are being sent. Please adjust them as well.

haiodo: Fixed.

lexiv0re (approved):

PR #8032, opened 2025-02-17
EQMS-1435: Enable watermark in obsolete documents

mr1name:


lexiv0re:
@mr1name the fix for PDF is good but the original task was about adding a watermark over an obsolete document in the platform itself (not only for PDF export)

mr1name:
@lexiv0re done

plugins/controlled-documents-resources/src/components/document/EditDocContent.svelte
```diff
@@ -262,6 +263,13 @@
             {$controlledDocument.title}
           {/if}
         </DocumentTitle>
+        {#if $controlledDocument.state === DocumentState.Obsolete}
+          <div class="watermark-container">
+            {#each { length: 24 } as _, i}
```
lexiv0re, line 268: What is this magic? :)

mr1name: What specifically?

lexiv0re: Like what's 24 here and what does it do at all?

lexiv0re (approved):

PR #8051, opened 2025-02-18
UBERF-9144: Stay in same view after delete sub-issue

mr1name:


packages/presentation/package.json
```diff
@@ -55,7 +55,8 @@
     "@hcengineering/collaborator-client": "^0.6.4",
     "fast-equals": "^5.2.2",
     "png-chunks-extract": "^1.0.0",
-    "uuid": "^8.3.2"
+    "uuid": "^8.3.2",
+    "@hcengineering/workbench": "^0.6.16"
```
haiodo: Nope!!!!

mr1name: why?

mr1name: The workbench is almost empty. I can't do the reverse due to the fact that presentation has svelte components.

haiodo: core -> presentation -> workbench, we could have any other starting apps, workbench is just one of them.

mr1name: hm... do you have any suggestions on where to shove this one function then?

haiodo: core should depend only on core
packages should depend only on packages and core, if we have some of deps they need to be eliminated. 
And of cause we should not add more.

mr1name: there are literally no other place to put it then atm 😄
i can only suggest leaving the fn in text-editor-resources, which is kinda strange, but...

mr1name: the problem is that text-editor-resources is the de-facto core for UI. Almost all packages that have Svelte components depend on it, so if something is needed there, it usually stays there, regardless of where it would be more appropriate to put it, because otherwise there will be a cyclic dependency

mr1name: though probably the idea was that we should avoid dependencies between *-resource packages, and the fact that there are any at all is something that needs to be addressed.

plugins/tracker-resources/src/index.ts
```diff
@@ -170,11 +170,12 @@ import ProjectSpacePresenter from './components/projects/ProjectSpacePresenter.s
...
 
 import { settingId } from '@hcengineering/setting'
+import type { TaskType } from '@hcengineering/task'
 import { getAllStates } from '@hcengineering/task-resources'
+import view, { type Filter } from '@hcengineering/view'
 import EstimationValueEditor from './components/issues/timereport/EstimationValueEditor.svelte'
 import TimePresenter from './components/issues/timereport/TimePresenter.svelte'
-import type { TaskType } from '@hcengineering/task'
-import view, { type Filter } from '@hcengineering/view'
+import { getTargetObjectFromUrl } from '@hcengineering/text-editor-resources/src/components/extension/reference'
```
haiodo: let's have a normal import please.

mr1name: done

haiodo (approved):

PR #8053, opened 2025-02-19
UBERF-9297: Display markdown message

ArtyomSavchenko:


packages/presentation/src/components/MessageViewer.svelte
```diff
@@ -28,6 +31,12 @@
...
-</div>
+{#if !isMarkdown}
+  <div class="text-markup-view">
+    <Node {node} {preview} />
+  </div>
+{:else}
+  <div class="markdown-view">
+    <Html value={marked(message, { async: false })} />
+  </div>
+{/if}
```
denis-tingaikin, line 42: It looks like `node` and `empty` are not going to be used for the `isMarkdown` branch, moreover, the `MessageViewer.svelte` looks  simple. 

Just curious, is it possible to use this code _inline_ in`MailThread.svelte` and do not modify  `ChatMessagePresenter.svelte`?

ArtyomSavchenko: In this case all other parts of `ChatMessagePresenter` will not be reused: author avatar, date, styles, etc.

packages/presentation/package.json
```diff
@@ -54,6 +54,7 @@
     "@hcengineering/client": "^0.6.18",
     "@hcengineering/collaborator-client": "^0.6.4",
     "fast-equals": "^5.2.2",
+    "marked": "^15.0.7",
```
haiodo, line 57: We need to use our markdown converter I think, so you only need to convert markdown used in github service or others and put it into Node as node.

ArtyomSavchenko: Yes, this option was a quick fix for demo, but now it's not actual.
I will create a new PR with markdown converter

PR #8072, opened 2025-02-21
UBERF-9504: Add role to employee mixin

lexiv0re:
Only differentiates between "user" and "guest" for now.

Related to: https://front.hc.engineering/workbench/platform/tracker/UBERF-9504

### DIFF
--- a/qms-tests/sanity/tests/documents/ES-40.spec.ts
+++ b/qms-tests/sanity/tests/documents/ES-40.spec.ts
@@ -61,6 +61,10 @@ test.describe('ISO 13485, 4.2.4 Control of documents ensure that documents of ex
       await page.waitForURL((url) => {
         return url.pathname.startsWith(`/workbench/${PlatformWs}/`)
       })
+      // We should wait for the user's client to finish creating social ids and ensuring employee
+      // Otherwise, when done in parallel with kicking the workspace the employee creation
+      // finishes after the kick. Better later change to some more meaningful wait but it's not obvious now.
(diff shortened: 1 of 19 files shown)


qms-tests/sanity/tests/documents/ES-40.spec.ts
```diff
@@ -61,6 +61,10 @@ test.describe('ISO 13485, 4.2.4 Control of documents ensure that documents of ex
       await page.waitForURL((url) => {
         return url.pathname.startsWith(`/workbench/${PlatformWs}/`)
       })
+      // We should wait for the user's client to finish creating social ids and ensuring employee
+      // Otherwise, when done in parallel with kicking the workspace the employee creation
+      // finishes after the kick. Better later change to some more meaningful wait but it's not obvious now.
+      await page.waitForTimeout(2000)
```
haiodo, line 67: Why?

haiodo (dismissed):

qms-tests/sanity/tests/documents/ES-40.spec.ts
lexiv0re: Because we have two windows with two different users. One creates and activates an employee and the other one kicks this employee from the workspace. Sometimes the first window enables the new employee after it's been kicked from another window and the test fails. And it's not very clear how to now wait for all the pieces of the employee to be created before starting kicking because we don't have all this info displayed in the UI. So it's a temporary thing to stabilize the test.

aonnikov (approved):

PR #8074, opened 2025-02-22
Rewrite calendar service

BykhovDenis:


BykhovDenis:
Moved to draft, 'isExternal' will not work, we have to handle source of tx, not event. Will fix

haiodo (approved):

PR #8085, opened 2025-02-24
Added migration to remove QMS internal codes in DocumentMeta

mr1name:


lexiv0re:
@mr1name do we need to reindex them as well?

mr1name:
@lexiv0re I'm not yet familiar with how we do these things. Do we need to bump something in the migration itself?

lexiv0re (approved):

PR #8088, opened 2025-02-25
UBERF-9516: Use cards for mail threads

ArtyomSavchenko:


models/mail/src/index.ts
```diff
@@ -74,3 +72,24 @@ export function createModel (builder: Builder): void {
...
+  builder.createDoc<Viewlet>(
+    view.class.Viewlet,
+    core.space.Model,
+    {
+      attachTo: mail.class.MailThread,
+      descriptor: view.viewlet.Table,
+      config: [
+        { key: 'createdBy', label: mail.string.From, displayProps: { fixed: 'left', key: 'app' } },
+        '',
+        { key: 'modifiedOn', label: mail.string.Date, displayProps: { key: 'modified', fixed: 'right' } }
```
BykhovDenis: Why?

BykhovDenis: If it only for label better override it using `declare` in model

BykhovDenis (dismissed):

models/mail/src/index.ts
ArtyomSavchenko: I've removed custom labels

BykhovDenis (approved):

PR #8093, opened 2025-02-25
EQMS-1463 hot fix

BykhovDenis:


plugins/contact-resources/src/components/CreateEmployee.svelte
```diff
@@ -72,7 +72,7 @@
 
       await client.createDoc(contact.class.Person, contact.space.Contacts, person, id)
       await client.createMixin(id, contact.class.Person, contact.space.Contacts, contact.mixin.Employee, {
-        active: false
```
haiodo, line 75: Why we need to make it active?

BykhovDenis: Because it's easier then fix condition for all components) We don't need it, but we definitely should have difference between kicked and inactive user, but I think we should do it on global account. In fact it's just revert previous PR and allows owners to edit other employees

haiodo (approved):

PR #8094, opened 2025-02-25
Fixes for updated emojis

SasLord:
1. Delayed emoji rendering in EmojiPopup for interface responsiveness
2. In the mobile version, emojis are grouped into groups, there is no general list.
3. Updated IconPicker, Reactions layout

screen-1 screen-2
screen-3
screen-4


plugins/view/src/types.ts
```diff
@@ -818,7 +818,7 @@ export interface ViewOptionsModel {
  */
 export interface IconProps {
   icon?: Asset
-  color?: number
+  color?: number | number[]
```
haiodo, line 821: Why?

haiodo (approved):

PR #8103, opened 2025-02-26
Merge 9522 9513

haiodo:
Merge 9522 9513


packages/api-client/src/client.ts
```diff
@@ -288,11 +288,18 @@ class PlatformClientImpl implements PlatformClient {
   }
 }
 
-async function getWorkspaceToken (
+export interface WorkspaceToken {
+  endpoint: string
+  token: string
+  workspaceId: string
```
lexiv0re: Please add correct type of the id

plugins/login-resources/src/utils.ts
```diff
@@ -390,6 +390,38 @@ export async function fetchWorkspace (): Promise<[Status, WorkspaceInfoWithStatu
...
+    return false
+  }
+
+  const request = {
+    method: 'performWorkspaceOperation',
+    params: [workspaceId, 'unarchive']
+  }
+
+  try {
+    const response = await fetch(accountsUrl, {
```
lexiv0re: Better use account client?

lexiv0re (dismissed):

plugins/login-resources/src/utils.ts
haiodo: Done.

packages/api-client/src/client.ts
haiodo: Done

lexiv0re (approved):

PR #8107, opened 2025-02-27
Add ESM output for packages used as external dependencies in workers and pages

Chunosov:


packages/api-client/src/socket/node.ts
```diff
@@ -14,12 +14,12 @@
 //
 
 import { type ClientSocket, type ClientSocketFactory } from '@hcengineering/client'
+import WebSocket from 'ws'
```
haiodo, line 17: Nope, no way, since package is intended to be also used from browser, and 'ws' is for nodejs websocket.

packages/account-client/tsconfig.esm.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig.json",
+
+  "compilerOptions": {
+    "module": "ESNext",
+    "outDir": "esm",
+    "tsBuildInfoFile": ".build/build.esm.tsbuildinfo"
+  }
+}
```
ArtyomSavchenko, line 9: What do you think about adding `tsconfig.esm.json` to the platform-rig base profiles and reusing it?

PR #8150, opened 2025-03-06
Improve rate limit on sendInvite

haiodo:


server/account/src/operations.ts
```diff
@@ -2558,6 +2562,20 @@ export async function sendInvite (
...
 
+async function checkSendRateLimit (currentAccount: Account, workspace: string, db: AccountDB): Promise<void> {
+  let sendOps = (currentAccount.sendOperations ?? 0) + 1
+
+  const timePassed = Date.now() - (currentAccount.lastWorkspace ?? 0)
+
+  if (timePassed < 60 * 1000 && sendOps > 5) {
+    throw new PlatformError(new Status(Severity.ERROR, platform.status.WorkspaceRateLimit, { workspace }))
+  }
+  if (sendOps > 5 && timePassed < sendOps * 60 * 1000) {
```
aonnikov: I think there is a mistake in this condition . If I have sendOps = 6 and timePassed = 1 month, to we do not reset sendOps.

haiodo: we will reset after 5 minutes

haiodo: My mistake

aonnikov (approved):

PR #8152, opened 2025-03-06
UBERF-9569: Fix hanging transactor connections

ArtyomSavchenko:
Before the fix http connections were hanging when there were errors while executing requests:
Screenshot 2025-03-07 at 11 48 02
Now error response should be returned in this case:
Screenshot 2025-03-07 at 12 47 42


aonnikov (approved):

server/server/src/sessionManager.ts
```diff
@@ -1188,6 +1188,7 @@ export class TSessionManager implements SessionManager {
             service.binaryMode,
             service.useCompression
           )
+          throw err
```
haiodo, line 1191: why we need it? as we already send response here?

ArtyomSavchenko: Here response is sending through websocket, but REST response is not sent.
Then in the function `withSession` the response is not sent as well and the connection remains until the timeout:
Screenshot 2025-03-07 at 16 27 51

haiodo: ok

PR #8163, opened 2025-03-07
UBERF-9577: Fix using default from address in emails

ArtyomSavchenko:


Chunosov (dismissed):

services/mail/pod-mail/src/main.ts
```diff
@@ -55,19 +58,36 @@ export const main = async (): Promise<void> => {
...
     res.status(400).send({ err: "'to' is missing" })
     return
   }
+  if (fromAddress === undefined) {
+    res.status(400).send({ err: "'from' is missing" })
+    return
+  }
+  const message: SendMailOptions = {
+    from: fromAddress,
+    to,
```
haiodo, line 79: to in 18 builds was converted to array, is it required?

ArtyomSavchenko: No, nodemailer accepts both string and array:
Screenshot 2025-03-07 at 12 14 52

services/mail/pod-mail/src/main.ts
```diff
@@ -21,6 +21,9 @@ import { createServer, listen } from './server'
 import { MailClient } from './mail'
 import { Endpoint } from './types'
 
+// Import Mail using require since esModuleInterop is not enabled
+import Mail = require('nodemailer/lib/mailer')
```
haiodo: why not enable it?

haiodo (commented):
please describe why 22 builds is not working in front

ArtyomSavchenko:
> please describe why 22 builds is not working in front

22 build is not working, because `SOURCE` env variable is not using when `from` address is not provided

services/mail/pod-mail/src/main.ts
ArtyomSavchenko: Fixed

haiodo (approved):

PR #8166, opened 2025-03-07
Support export to JSON, CSV

akhismat:
Support exporting:

1. Issues
2. Documents
3. Test Cases
4. Test Plans
5. Test Runs


plugins/setting-resources/src/components/Export.svelte
```diff
@@ -0,0 +1,133 @@
...
+-->
+<script lang="ts">
+  import { Class, Doc, Ref } from '@hcengineering/core'
+  import { getMetadata } from '@hcengineering/platform'
+  import presentation, { MessageBox } from '@hcengineering/presentation'
+  import { Breadcrumb, Button, DropdownLabelsIntl, Header, Label, Scroller, showPopup } from '@hcengineering/ui'
+  import setting from '../plugin'
+
+  const classItems = [
+    { id: 'tracker:class:Issue', label: setting.string.ExportIssues },
```
haiodo, line 23: All of this should be contributable via model and probable different plugins.

services/export/pod-export/src/types.ts
```diff
@@ -0,0 +1,55 @@
...
+  // Base document properties
+  _class: string // Class reference (generic, not hardcoded)
+  space: string // Space reference
+  _id?: string // Optional document ID
+
+  // Document data
+  data: UnifiedAttributes<T> // Generic data type
+
+  // Special field handling
+  markdownFields?: string[] // Fields that contain markdown content
```
haiodo, line 26: Why we need all of them?

plugins/setting-resources/src/components/Export.svelte
haiodo: This could be in next PR.

services/export/pod-export/src/types.ts
akhismat: They are used when serializing into yaml+markdown format (in progress yet)

haiodo (approved):

PR #8173, opened 2025-03-07
Forced update db schema

BykhovDenis:


server/postgres/src/utils.ts
```diff
@@ -145,7 +144,10 @@ async function getTableSchema (client: postgres.Sql, domains: string[]): Promise
     }
   }
   for (const [domain, schema] of Object.entries(schemas)) {
-    addSchema(domain, schema)
+    const schemaMigration = setSchema(domain, schema)
+    if (schemaMigration !== undefined) {
+      await client.unsafe(schemaMigration)
```
haiodo, line 149: we need a tool for it and execute before deployment, since it could be many clients, and migrations could go bad.

PR #8184, opened 2025-03-10
feat: add datalake service

aonnikov:
motivation: rework datalake from CF worker to standalone service


services/datalake/pod-datalake/src/datalake/db.ts
```diff
@@ -0,0 +1,461 @@
...
+  }
+}
+export interface WorkspaceStatsResult {
+  count: number
+  size: number
+}
+
+export function createDb (ctx: MeasureContext, connectionString: string): BlobDB {
+  const sql = postgres(connectionString, {
+    max: 5,
```
haiodo, line 72: We could have more, please allow to configure via ENV variables in next PR.

services/datalake/pod-datalake/src/datalake/db.ts
```diff
@@ -0,0 +1,461 @@
...
+  const dbAdapter = new PostgresDBAdapter(sql)
+  return new LoggedDB(new PostgresDB(new RetryDBAdapter(dbAdapter, { retries: 5 })), ctx)
+}
+
+export async function withPostgres<T> (
+  ctx: MeasureContext,
+  connectionString: string,
+  fn: (db: BlobDB) => Promise<T>
+): Promise<T> {
+  const sql = postgres(connectionString, {
```
haiodo: Same stuff two times a row?

haiodo (dismissed):

services/datalake/pod-datalake/src/datalake/db.ts
aonnikov: unused from worker, will double check and remove

services/datalake/pod-datalake/src/server.ts
```diff
@@ -0,0 +1,199 @@
...
+    '/upload/form-data/:workspace',
+    withAuthorization,
+    withWorkspace,
+    wrapRequest(ctx, datalake, handleUploadFormData)
+  )
+
+  // S3 upload
+
+  app.get(
+    '/upload/r2/:workspace',
```
denis-tingaikin: Why do we need to specify r2 here?

aonnikov: nice catch, it should be `s3` because here we allow to directly access s3 buckets

haiodo (approved):

PR #8188, opened 2025-03-11
Survey plugin tweaks & bug fixes

mr1name:


haiodo:
we need this into staging, also please look we already had some debouncer in code:-)

mr1name:
@haiodo Sure, a second PR will be as soon as this one is approved in case any revisions are needed. About Deboucer - they all have slightly different logic, I haven't seen a suitable one, though maybe they should all be unified.

Chunosov (approved):

PR #8189, opened 2025-03-11
UBERF-9605: Test MTA hook integration

ArtyomSavchenko:
Add service that should handle MTA hook request from SMTP server, it should replace Cloudflare Email Worker for inbound emails


haiodo (approved):

aonnikov:
`hook` name is too generic per my point of view, is it for email?

ArtyomSavchenko:
> `hook` name is too generic per my point of view, is it for email?

I think this service can be used for several integrations, for example, mail, telegram, slack.
From my point of view this service should only send event into the queue with corresponding topic:
- mail hook -> mail topic
- telegram hook -> telegram topic
etc.

PR #8195, opened 2025-03-11
Ws tests stability

haiodo:


lexiv0re:
@haiodo wrong target branch

PR #8205, opened 2025-03-11
EQMS-1472: reuse the latest deleted draft instead of making a new version if possible

mr1name:


plugins/controlled-documents-resources/src/components/EditDocPanel.svelte
```diff
@@ -220,39 +220,43 @@
   }
 
   async function onCreateNewDraft (): Promise<void> {
```
lexiv0re, line 223: My only concern now is that we have the same-looking button doing different things which might be a little bit confusing. Could we also update the button's title for this specific case? To something like "Restore draft"

mr1name: Yeah, but here's the other thing... What I didn't think of is that if a draft was removed during the review, it will also retain all the other stuff attached to it. (e.g. review request). Should we clean it out?

lexiv0re: No, I think we can keep them. But we should then make sure we cancel all pending requests when the draft is being deleted.

plugins/controlled-documents-resources/src/components/EditDocPanel.svelte
```diff
@@ -238,7 +239,7 @@
         return
       }
 
-      if (latest.state === DocumentState.Deleted) {
+      if ($canRestoreDraft) {
```
lexiv0re, line 242: It seems to only be dependent on the latest document version but not the version of the document currently open? E.g. for latest effective or all deleted drafts after effective it would go this route which is not correct?

lexiv0re (approved):

PR #8215, opened 2025-03-12
UBERF-9616 Datalake blob meta

aonnikov:


services/datalake/pod-datalake/src/datalake/db.ts
```diff
@@ -284,6 +294,34 @@ export class PostgresDB implements BlobDB {
...
 
+  async getMeta (ctx: MeasureContext, blobId: BlobId): Promise<BlobMeta | null> {
+    const { workspace, name } = blobId
+
+    const rows = await this.db.execute<BlobMetaRecord[]>(
+      `
+      SELECT m.workspace, m.name, m.meta
+      FROM blob.blob b
+      LEFT JOIN blob.meta as m ON m.workspace = b.workspace AND m.name = b.name
+      WHERE b.workspace = $1 AND b.name = $2 and b.deleted_at IS NULL
```
ArtyomSavchenko, line 305: Do we need index for `deleted_at`?

services/datalake/pod-datalake/src/datalake/db.ts
```diff
@@ -284,6 +294,34 @@ export class PostgresDB implements BlobDB {
...
+
+    return rows.length > 0 ? rows[0].meta ?? {} : null
+  }
+
+  async setMeta (ctx: MeasureContext, blobId: BlobId, meta: BlobMeta): Promise<void> {
+    const { workspace, name } = blobId
+
+    await this.db.execute(
+      `
+      UPSERT INTO blob.meta (workspace, name, meta)
```
ArtyomSavchenko, line 318: Looks like `UPSERT` command exists in CockroachDB, but in standard PostgreSQL there is only INSERT ... ON CONFLICT
Wouldn't it better to use [INSERT ... ON CONFLICT](https://www.cockroachlabs.com/docs/v25.1/insert#on-conflict-clause) statement?

services/datalake/pod-datalake/schema/datalake.sql
```diff
@@ -0,0 +1,39 @@
+
+CREATE SCHEMA IF NOT EXISTS blob;
+
+-- B L O B
+
+CREATE TABLE IF NOT EXISTS blob.location AS ENUM ('eu', 'weur', 'eeur', 'wnam', 'enam', 'apac');
```
ArtyomSavchenko: Shouldn't we use `CREATE TYPE` instead of `CREATE TABLE` here?

services/datalake/pod-datalake/schema/datalake.sql
```diff
@@ -0,0 +1,39 @@
...
+
+\echo "Creating blob.data..."
+CREATE TABLE IF NOT EXISTS blob.data (
+  hash UUID NOT NULL,
+  location blob.location NOT NULL,
+  size INT8 NOT NULL,
+  filename STRING(255) NOT NULL,
+  type STRING(255) NOT NULL,
```
ArtyomSavchenko, line 14: Do we need PostgreSQL compatibility?
If so, VARCHAR can be used, for CockroachDB both types should work identically
The same thing for INT8 and BIGINT alias

services/datalake/pod-datalake/src/datalake/db.ts
aonnikov: no, we don't need
it would be needed if we used sorting on delete date, but for nul//not null queries it will have low cardinality and won't be used

services/datalake/pod-datalake/schema/datalake.sql
aonnikov: I can do this separately. Current version is not compatible with PostgreSQL.

services/datalake/pod-datalake/src/datalake/db.ts
aonnikov: Can be done separately if we decide we need postgresql compatibility.

services/datalake/pod-datalake/schema/datalake.sql
aonnikov: thanks, fixed

services/datalake/pod-datalake/src/datalake/db.ts
ArtyomSavchenko: Looks like the following index can be used in this case:
`CREATE INDEX ON blob.blob (deleted_at) WHERE deleted_at IS NULL;`
But I don't mind leaving it as is

aonnikov: Index on boolean field is usually a bad idea because it has low selectivity. Moreover, in this query we already filter by primary key (workspace and name) and deleted_at` is used as filter.

ArtyomSavchenko (approved):

haiodo (approved):

PR #8219, opened 2025-03-13
Turkcelestirme

erkanyldz58:


BykhovDenis:
Hello. If you want to add Turkish languaguage please don't override Spain, create tr.json files instead.

PR #8224, opened 2025-03-13
QFix: (drive) add title providers for files and folders [develop]

mr1name:


plugins/drive-resources/src/index.ts
```diff
@@ -171,6 +171,18 @@ export async function CanDeleteFileVersion (
   return docs.every((p) => p.$lookup?.attachedTo !== undefined && p.$lookup?.attachedTo.file !== p._id)
 }
 
+export async function FileTitleProvider (client: Client, ref: Ref<File>, doc?: File): Promise<string> {
+  const object = doc ?? (await client.findOne(drive.class.File, { _id: ref }))
```
ArtyomSavchenko, line 175: Should we add projection with `title` field, since other fields are unused?

mr1name: This question can be addressed to thousands of places in the codebase. (For example, to each of the same providers). My understanding is that we use projections only in critical places. Moreover, this query is almost never used in primary scenarios, as it is expected that the object is passed in the arguments.

haiodo (approved):

PR #8233, opened 2025-03-14
fix: use huly.local in webpack proxy config

aonnikov:


haiodo (approved):

mingfang:
why was this done?
this broke everything for me and i had to revert everything back to localhost to restore my local dev setup.

PR #8247, opened 2025-03-17
Increase minimum tab width and hide close buttons on inactive tabs

denisfl:
With a large number of open tabs, it becomes difficult to navigate between them. Certainly, it would be great to keep only a small number of tabs open. However, since there's no limit on opening tabs, I suggest improving tab readability and preventing their accidental closure.

image

Readability can be improved by increasing the minimum tab width. I increased the minimum tab width value by one point. Since I don't have any data on how users work with the tab panel, I rely on personal experience and observations of similar mechanics in various browsers (Chrome-based browsers and Firefox).

Assuming that user information is of greater value, I suggest hiding the "Close" buttons for all inactive tabs. This can help avoid accidentally closing tabs when switching between them.

SCR-20250317-kbbm

_Each tab gets its well-deserved millimeters of screen space._


haiodo (approved):

denisfl:
I noticed that the tests have failed. I'll take a timeout to figure this out.

PR #8286, opened 2025-03-19
Do not preload video by default

denis-tingaikin:
Since we've added on backend video preview picture, we do not need to expend UI resources to load the video.


kristina-fefelova:
Do we need `videoPreload` prop at all? is it used?

haiodo (approved):

PR #8288, opened 2025-03-19
Add a token for stream requests

denis-tingaikin:


haiodo (approved):

services/datalake/pod-datalake/src/handlers/video.ts
```diff
@@ -28,10 +30,12 @@ export async function requestHLS (ctx: MeasureContext, workspace: string, name:
     return
   }
   const streamReq: StreamRequest = { format: 'hls', source: name, workspace }
+  const token = generateToken(systemAccountUuid)
```
aonnikov, line 33: recommend to add `extra` with service name so we know which service makes requests

denis-tingaikin: what is extra? I think we need to set aud/issuer

aonnikov: See `generateToken` function:
```
export function generateToken (
  accountUuid: PersonUuid,
  workspaceUuid?: WorkspaceUuid,
  extra?: Record<string, string>,
  secret?: string
): string {
  ...
}
```

denis-tingaikin: ok, I see extra is a part of payload. Will fix :)

PR #8294, opened 2025-03-20
Fix link preview size

kristina-fefelova:
- save  title, description, image, size in attachment meta
- fix image size


haiodo (dismissed):

haiodo (approved):

denis-tingaikin (commented):
looks good,

A quick question, how will we handle a text from the description? For example, will it blink on the description loaded?

UPD:  OK, I see we simply store it in the db ;)

PR #8306, opened 2025-03-21
qfix: add support for recording videos from desktop

denis-tingaikin:


plugins/recorder-resources/src/screen-recorder.ts
```diff
@@ -19,21 +19,33 @@ import { TusUploader, type Uploader, type Options } from './uploader'
 export class ScreenRecorder {
   private readonly recorder: Recorder
   private readonly uploader: Uploader
+  static getMediaStream = async (options?: DisplayMediaStreamOptions): Promise<MediaStream> =>
```
denis-tingaikin: @aonnikov Do you have any thoughts on how could we solve this without static?

aonnikov: have you tried to override `navigator.mediaDevices.getDisplayMedia` instead?

denis-tingaikin: nice idea, fixed ;)

haiodo (approved):

PR #8310, opened 2025-03-21
Small visual fixes fot calendar events

Chunosov:


packages/ui/src/components/Dropdown.svelte
```diff
@@ -55,12 +56,13 @@
...
     {justify}
     {disabled}
+    {padding}
     showTooltip={{ label, direction: labelDirection }}
     on:click={() => {
       if (!opened && !disabled) {
         opened = true
         showPopup(DropdownPopup, { title: label, items, icon, withSearch }, container, (result) => {
-          if (result) {
+          if (result !== undefined) {
```
haiodo: Could it be null?

Chunosov: Actually we do different things in different places - compare with nulls, compare with undefined, compary with falsy :-) So it seems the latter is better, restored it. (I changed beause vscode underlines this lines with eslint warning, it doesn't like comarison to falsy)

haiodo (approved):

PR #8320, opened 2025-03-24
Bug fix

### LINKED ISSUE #8071: Bug: Duplicated Labels in Issue Tracker
### **Description**

When adding and removing a newly created label in an issue, the system allows the label to be added multiple times, even when it is unchecked.

### **Steps to Reproduce**

1. Create a new issue in Tracker;
2. Create a new label and add it to the issue;
3. Click to remove the label;
4. Click again to add and remove the label multiple times;
5. Observe that the label keeps getting added multiple times instead of being removed.

### **Expected Behavior**

The label should be removed from the issue when unchecked and should not be duplicated.

### **Observed Behavior**

The label is duplicated indefinitely when added and removed multiple times, even if it is unchecked.

### **Evidence**

📹 Video demonstration:

https://github.com/user-attachments/assets/fb064588-c448-4179-8767-70753f33e25a

🖼️ Screenshot of issue with duplicated labels:

image

---

OBS: This is my first issue report in GitHub, if something is missing, please tell me. 🙂

rockcoolsaint:
When creating an issue tracker, the dispatch command wasn't removing the tag and consequently not counting the labels properly

The bug was as a result of using the `_id`  key in the object as opposed to the `tag` key in the `objects.labels` filter

closes #8071

### DIFF
--- a/plugins/tracker-resources/src/components/CreateIssue.svelte
+++ b/plugins/tracker-resources/src/components/CreateIssue.svelte
@@ -957,7 +957,7 @@
         addTagRef(evt.detail)
       }}
       on:delete={(evt) => {
-        object.labels = object.labels.filter((it) => it._id !== evt.detail)
+        object.labels = object.labels.filter((it) => it.tag !== evt.detail)
       }}
     />
     <ComponentSelector


rockcoolsaint:
You can take a look @haiodo @aplatoff

plugins/tracker-resources/src/components/CreateIssue.svelte
```diff
@@ -957,7 +957,7 @@
         addTagRef(evt.detail)
       }}
       on:delete={(evt) => {
-        object.labels = object.labels.filter((it) => it._id !== evt.detail)
+        object.labels = object.labels.filter((it) => it.tag !== evt.detail)
```
haiodo, line 960: Hi, Fix is not fully correct, in TagsDropDownEditor.svelte it could dispatch delete with both _id if tag element in case we open editor in already created application and with .tag if not.

```
else if (result.action === 'remove') {
            const filtered = items.filter((it) => (it.tag ? it.tag === result.tag._id : it._id === result.tag._id))
            if (filtered.length > 0) {
              void removeTag(filtered[0].tag ?? filtered[0]._id)
            }
          }
```

rockcoolsaint:
I'm unable to continue fixing this bug because of an issue I'm experiencing as described in issue https://github.com/hcengineering/platform/issues/8347
Any clue I can use in fixing that?

@haiodo

haiodo:
Hi, @rockcoolsaint I've already added a fix for this issue for all other apps and tracker. 
So will close this PR, thanks for your work.

PR #8342, opened 2025-03-25
Account mailboxes

Chunosov:


server/account/src/collections/postgres.ts
```diff
@@ -389,6 +393,15 @@ export class PostgresAccountDB implements AccountDB {
     this.accountEvent = new PostgresDbCollection<AccountEvent>('account_events', client)
     this.otp = new PostgresDbCollection<OTP>('otp', client)
     this.invite = new PostgresDbCollection<WorkspaceInvite, 'id'>('invite', client, 'id')
+    this.mailbox = new PostgresDbCollection<Mailbox, 'mailbox'>(
+      'mailbox',
+      (client as any).mailboxClient ?? client,
+      'mailbox'
```
lexiv0re, line 399: I'd not use it as a key here otherwise it will be filling it with uuid on insert if not provided which might be confusing.

server/account/src/collections/mongo.ts
```diff
@@ -382,6 +386,8 @@ export class MongoAccountDB implements AccountDB {
     this.accountEvent = new MongoDbCollection<AccountEvent>('accountEvent', db)
     this.otp = new MongoDbCollection<OTP>('otp', db)
     this.invite = new MongoDbCollection<WorkspaceInvite, 'id'>('invite', db, 'id')
+    this.mailbox = new MongoDbCollection<Mailbox, 'mailbox'>('mailbox', db, 'mailbox')
```
lexiv0re: I'd not use it as a key here otherwise it will be filling it with uuid on insert if not provided which might be confusing.

server/account/src/utils.ts
```diff
@@ -80,6 +80,14 @@ export async function getAccountDB (uri: string, dbNs?: string): Promise<[Accoun
   } else {
     const client = getDBClient(sharedPipelineContextVars, uri)
     const pgClient = await client.getClient()
+
+    // For testing, mailboxes could be stored in a separate database closer to the mailbox service
```
lexiv0re: Do we really need it? Can we just store all accounts data in this DB when testing mailboxes?

server/account/src/operations.ts
```diff
@@ -1971,6 +1975,86 @@ export async function ensurePerson (
...
+  params: { mailbox: string }
+): Promise<void> {
+  const { account } = decodeTokenVerbose(ctx, token)
+  const { mailbox } = params
+
+  const mb = await db.mailbox.findOne({ mailbox, accountUuid: account })
+  if (mb == null) {
+    throw new PlatformError(new Status(Severity.ERROR, platform.status.MailboxError, { reason: 'mailbox-not-found' }))
+  }
+  if (mb.accountUuid !== account) {
```
lexiv0re: Not possible as you query mailboxes only for this account so all these scenarios currently fall into `mailbox-not-found`

server/account/src/collections/mongo.ts
Chunosov: Removed

server/account/src/utils.ts
Chunosov: Mail server does not have access to the dev db because it's on localhost under provider NAT. So without this hack we have to test with not-working mail server running on localhost.

server/account/src/operations.ts
Chunosov: Sure, this is some refactoring remainder :) removed

server/account/src/utils.ts
lexiv0re: I mean why do we need to have a part of the accounts locally and mailboxes part externally? If we have this external DB anyways can we just use it to have all the dev accounts information when we need to debug this mail functionality?

Chunosov: Ah, I see. I tried to run the whole db on that my test node. But it's too weak to run cockroach (it seems much heavier than postgres). The node runs postgres, but the account db schema uses cockroach-specific sql syntax that is not compatible with postgres. So the account does not start. So this is why this ugly solution arised :-)
Anyway, I'll remove this code to keep it simpler. I can add it temporary for testing.

haiodo (approved):

server/account/src/utils.ts
haiodo: you could use postgres, we still compatible with it:-)

Chunosov: Removed

lexiv0re: @haiodo accounts are no longer compatible with postgres

PR #8365, opened 2025-03-27
Fix meeting window width

Chunosov:


aonnikov (approved):

packages/theme/styles/components.scss
```diff
@@ -595,27 +595,34 @@
...
+      min-width: unset;
       width: 25rem;
     }
     &.menu {
+      min-width: unset;
       width: 17.5rem;
     }
     &.full {
+      min-width: unset;
       width: max-content;
```
ArtyomSavchenko, line 626: Shouldn't we use `width: 100%;` instead of it?
Or add `&.full-screen` modifier with `width: 100%;`  if `&.full` is used to display full content

Chunosov: I don't know, this is the copy of styles from `dialogs.css`.  If it work there, it will work here.

PR #8406, opened 2025-03-31
Mailbox fixes

Chunosov:
- Enable my-space app in desktop client
- Use token for securing the mail hook
- Create/remove social-id when create/remove mailbox
- Create local person and social-id for mail addressee

### DIFF
--- a/desktop/package.json
+++ b/desktop/package.json
@@ -12,8 +12,8 @@
     "webpack": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=development webpack --stats-error-details --progress -w",
     "devp": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=production CLIENT_TYPE=dev webpack --progress -w",
     "dev": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=development webpack --progress -w",
-    "start": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=production electron .",
-    "start-dev": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=development electron .",
+    "start": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=production electron --no-sandbox .",
+    "start-dev": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=development electron --no-sandbox .",
     "format": "format",
     "bump": "bump-package-version"
   },
--- a/plugins/setting-resources/src/components/MailboxItem.svelte
+++ b/plugins/setting-resources/src/components/MailboxItem.svelte
@@ -25,29 +25,19 @@
   } from '@hcengineering/ui'
   import setting from '@hcengineering/setting'
   import { MailboxInfo } from '@hcengineering/account-client'
-  import { MessageBox } from '@hcengineering/presentation'
+  import { getClient, MessageBox } from '@hcengineering/presentation'
   import { getAccountClient } from '../utils'
+  import contact, { getCurrentEmployee } from '@hcengineering/contact'
+  import { buildSocialIdString, SocialIdType } from '@hcengineering/core'
 
   export let mailbox: MailboxInfo
   export let mailboxIdx: number
+  export let loadingRequested: () => void
   export let reloadRequested: () => void
 
   let opened = false
 
-  function getMenuItems (mailbox: MailboxInfo): (DropdownIntlItem & { action: () => void })[] {
-    return [
-      {
-        id: 'delete',
-        icon: IconDelete,
-        label: setting.string.DeleteMailbox,
-        action: () => {
-          deleteMailbox(mailbox)
-        }
-      }
-    ]
-  }
-
-  function deleteMailbox (mailbox: MailboxInfo): void {
+  function deleteMailboxAction (): void {
     showPopup(
       MessageBox,
       {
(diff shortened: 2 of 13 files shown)


desktop/package.json
```diff
@@ -12,8 +12,8 @@
     "webpack": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=development webpack --stats-error-details --progress -w",
     "devp": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=production CLIENT_TYPE=dev webpack --progress -w",
     "dev": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=development webpack --progress -w",
-    "start": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=production electron .",
-    "start-dev": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=development electron .",
+    "start": "cross-env MODEL_VERSION=$(node ../common/scripts/show_version.js) VERSION=$(node ../common/scripts/show_tag.js) NODE_ENV=production electron --no-sandbox .",
```
lexiv0re, line 15: Just curious why?

server/account/src/operations.ts
```diff
@@ -2081,15 +2088,33 @@ async function deleteMailbox (
...
   }
 
-  await db.mailbox.deleteMany({ mailbox })
   await db.mailboxSecret.deleteMany({ mailbox })
+  await db.mailbox.deleteMany({ mailbox })
+  await releaseSocialId(db, account, SocialIdType.EMAIL, mailbox)
+  ctx.info('Mailbox deleted', { mailbox, account })
+}
+
+async function releaseSocialId (
```
lexiv0re, line 2108: Please move all utility functions to `utils.ts`

server/account/src/operations.ts
```diff
@@ -2081,15 +2088,33 @@ async function deleteMailbox (
...
+  db: AccountDB,
+  personUuid: PersonUuid,
+  type: SocialIdType,
+  value: string
+): Promise<void> {
+  const socialIds = await db.socialId.find({ personUuid, type, value })
+  for (const socialId of socialIds) {
+    await db.socialId.updateOne({ _id: socialId._id }, { value: `${socialId.value}#${socialId._id}` })
```
lexiv0re, line 2116: We should probably also mark them as deleted explicitly so we know e.g. not to use them for new objects creation etc.

services/mail/pod-inbound-mail/src/person.ts
```diff
@@ -69,20 +70,30 @@ export async function ensureLocalPerson (
...
       contact.class.Person,
       'socialIds',
       {
-        key: socialId,
+        key: buildSocialIdString({ type: SocialIdType.EMAIL, value: email }),
         type: SocialIdType.EMAIL,
         value: email
       },
-      generateId()
+      personId as any as Ref<SocialIdentity>
```
lexiv0re, line 94: Minor: there's a `SocialIdentityRef` type. It can be used w/o casting to any first.

desktop/package.json
Chunosov: In Ubuntu 24 they somehow "increased safety" making running electron apps impossible without creating something called "sandbox environment" which seems nobody care about, because most advices in Internet just say "run it without sandboxing" :-)

server/account/src/operations.ts
Chunosov: Do you suggest adding some "deleted" column and checking it on every object creation/modifcation?

plugins/setting-resources/src/components/MailboxItem.svelte
```diff
@@ -56,24 +46,71 @@
...
+        }
+      )
+    }
+    const channels = await client.findAll(contact.class.Channel, {
+      attachedTo: currentUser,
+      provider: contact.channelProvider.Email,
+      value: mailbox.mailbox
+    })
+    for (const channel of channels) {
+      await client.removeCollection(
```
ArtyomSavchenko, line 92: What do you think about using rate limiter for  `channels ` and `socialIds` requests?

services/mail/pod-inbound-mail/src/handlerMta.ts
```diff
@@ -36,6 +36,13 @@ interface MtaMessage {
 
 export async function handleMtaHook (req: Request, res: Response): Promise<void> {
   try {
+    if (config.hookToken !== undefined) {
+      const token = req.headers['x-hook-token']
+      if (token !== config.hookToken) {
+        throw new Error('Invalid hook token')
```
ArtyomSavchenko, line 42: Will the 500 response status return In this case?
I think we should return 401 response code in case of incorrect token

services/mail/pod-inbound-mail/src/message.ts
```diff
@@ -61,13 +61,42 @@ export async function createMessages (
...
+      mailId,
+      fromPerson.uuid,
+      fromPerson.socialId,
+      from.address,
+      fromPerson.firstName,
+      fromPerson.lastName
+    )
+  } catch (err) {
+    console.error(`[${mailId}] Failed to ensure local FROM person`, err)
+    console.error(`[${mailId}] Unable to create message without a proper FROM`)
```
ArtyomSavchenko, line 76: [Minor] Isn't it enough to log the error once?

plugins/setting-resources/src/components/MailboxItem.svelte
Chunosov: There should be only one item in each collection. Loops here are 'just in case'

services/mail/pod-inbound-mail/src/handlerMta.ts
Chunosov: No, it always retrurn 200 in order not to prevent stalwart from sending emails

services/mail/pod-inbound-mail/src/message.ts
Chunosov: This is for generality with global person failure - there will be an error with stack, and then a separate message about the process has been interrupted

ArtyomSavchenko (approved):

server/account/src/operations.ts
haiodo: Let's do in next PR

services/mail/pod-inbound-mail/src/person.ts
haiodo: In next PR please

server/account/src/operations.ts
haiodo: In next PR please

desktop/package.json
lexiv0re: Won't we have the same problem in production build of the desktop on Ubuntu?

PR #8425, opened 2025-04-01
Svetlov testwork recruit plugin refactor

Teamiha:
Starting work on the test task "to tidy up the project".
I decided to start with refactoring the recruiting module.
The CreateCandidate file was decomposed, the amount of code was reduced by dividing it into modules.
This reduced the file size by half, making it easier to read and edit in the future.

The solution has not yet been tested, since at the moment the project has a bug with accessibility, which prevents it from running locally.


Teamiha:
Tested for local version

PR #8435, opened 2025-04-02
QFix: Show proper current employee

haiodo:
QFix: Show proper current employee

### DIFF
--- a/plugins/workbench-resources/src/components/AccountPopup.svelte
+++ b/plugins/workbench-resources/src/components/AccountPopup.svelte
@@ -130,6 +130,25 @@
         group: 'end'
       })
     }
+
+    if (hasAccountRole(account, AccountRole.User)) {
+      actions.push({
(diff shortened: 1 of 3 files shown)


lexiv0re (approved):

plugins/workbench-resources/src/components/AccountPopup.svelte
```diff
@@ -130,6 +130,25 @@
         group: 'end'
       })
     }
+
+    if (hasAccountRole(account, AccountRole.User)) {
```
BykhovDenis, line 134: Why? Guest probably want to change workspace to...?)

BykhovDenis (approved):

plugins/workbench-resources/src/components/AccountPopup.svelte
haiodo: Hmps :-)

PR #8439, opened 2025-04-02
UBERF-9727: Allow adding social id to existing person

lexiv0re:
* Added new `addSocialIdToPerson` method to the accounts service

Related to: https://front.hc.engineering/workbench/platform/tracker/UBERF-9727

### DIFF
--- a/server/account/src/operations.ts
+++ b/server/account/src/operations.ts
@@ -101,7 +101,8 @@ import {
   getWorkspaceRole,
   normalizeValue,
   isEmail,
-  generatePassword
+  generatePassword,
+  addSocialId
 } from './utils'
 
 // Move to config?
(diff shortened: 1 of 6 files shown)


server/account/src/operations.ts
```diff
@@ -2107,6 +2108,21 @@ async function deleteMailbox (
...
+  ctx: MeasureContext,
+  db: AccountDB,
+  branding: Branding | null,
+  token: string,
+  params: { person: PersonUuid, type: SocialIdType, value: string, confirmed: boolean }
+): Promise<PersonId> {
+  const { person, type, value, confirmed } = params
+  const { extra } = decodeTokenVerbose(ctx, token)
+
+  verifyAllowedServices(['github'], extra)
```
haiodo, line 2121: May be we should have just a service extra? Instead of adding individual restrictions? Or let's have environment variable to pass them all to account?

server/account/src/utils.ts
```diff
@@ -1274,6 +1274,42 @@ export async function getInviteEmail (
...
+  if (!Object.values(SocialIdType).includes(type) || normalizedValue.length === 0) {
+    throw new PlatformError(new Status(Severity.ERROR, platform.status.BadRequest, {}))
+  }
+
+  const person = await db.person.findOne({ uuid: personUuid })
+  if (person == null) {
+    throw new PlatformError(new Status(Severity.ERROR, platform.status.PersonNotFound, { person: personUuid }))
+  }
+
+  const socialId = await db.socialId.findOne({ type, value: normalizedValue })
```
haiodo, line 1295: What if I will authorize from different user using same github socialId?

server/account/src/operations.ts
lexiv0re: Different services may need different service methods. E.g. the print service is never supposed to add social ids to persons etc. Exposing all of the methods to all the services may be dangerous if token from one of the service gets exposed somehow. So I would like to keep the rights as minimal as needed for things to work.

haiodo (approved):

PR #8452, opened 2025-04-03
UBERF-9724: Github use of updated accounts

haiodo:
UBERF-9724: Github use of updated accounts


plugins/chunter-resources/src/components/chat-message/ChatMessagePresenter.svelte
```diff
@@ -78,6 +82,13 @@
   $: personId = value?.createdBy
   $: person = personId !== undefined ? $personByPersonIdStore.get(personId) : undefined
 
+  $: socialId =
```
lexiv0re: I think we can always pass this id regardless if it's primary or not but then in the message template only show ids if type != huly. I created a task to always use huly as a primary id in the UI so it will work fine then.

plugins/contact-resources/src/utils.ts
```diff
@@ -385,17 +387,18 @@ export const personRefByAccountUuidStore = writable<Map<AccountUuid, Ref<Employe
...
  */
 export const personByPersonIdStore: Readable<Map<PersonId, Person>> = derived(
-  [personRefByPersonIdStore, personByIdStore],
-  ([personRefByPersonId, personById]) => {
+  [personRefByPersonIdStore, personByIdStore, personBySocialIdStore],
+  ([personRefByPersonId, personById, personBySocialIdStore]) => {
     const mapped = Array.from(personRefByPersonId.entries())
       .map(([personId, personRef]) => {
-        const person = personById.get(personRef)
+        const person = personById.get(personRef) ?? personBySocialIdStore.get(personRef)
```
lexiv0re: Should we instead union `personById` store and new `personBySocialIdStore` store into a derived store keeping old `personById` name (so we don't need to rename all existing uses). And don't even expose the first two from this utils file only the combined one.

plugins/contact-resources/src/utils.ts
```diff
@@ -485,9 +488,20 @@ onClient(() => {
...
-  siQuery.query(contact.class.SocialIdentity, {}, (res) => {
-    socialIdsStore.set(res)
-  })
+  siQuery.query(
+    contact.class.SocialIdentity,
+    {},
+    (res) => {
+      socialIdsStore.set(toIdMap(res))
+      const persons = res.map((it) => it.$lookup?.attachedTo).filter((it) => it !== undefined) as Person[]
+      personBySocialIdStore.set(toIdMap(persons))
```
lexiv0re: Minor: the name is confusing as it's actually something like `personByIdFromSocialIdsTable` :) With the existing name (`personBySocialIdStore`) I'd expect it to contain `PersonId` -> `Person` mapping rather than `Ref<Person>` -> `Person`.

services/github/pod-github/src/platform.ts
```diff
@@ -106,10 +110,63 @@ export class PlatformWorker {
...
+
     const storageConfig = storageConfigFromEnv()
     this.storageAdapter = buildStorageFromConfig(storageConfig)
   }
 
+  private async performMigrations (db: Db): Promise<void> {
+    const migrations = db.collection<any>('migrations')
+    // Check and apply migrations
+    const accountMigration = await migrations.findOne({ name: 'account-to-uuid' })
+    if (accountMigration == null) {
```
lexiv0re: Do we have only one instance of GitHub? It won't work well with multiple instances as they may start that same migration simultaneously.

services/github/pod-github/src/platform.ts
```diff
@@ -106,10 +110,63 @@ export class PlatformWorker {
...
+      // We need to update all workspace information accordingly
+
+      const token = generateToken(systemAccountUuid, '' as WorkspaceUuid, { service: 'github', mode: 'github' })
```
lexiv0re: Minor: Is `mode` in token used anywhere?

services/github/pod-github/src/platform.ts
```diff
@@ -244,7 +302,7 @@ export class PlatformWorker {
         } else {
           let client: Client | undefined
           try {
-            ;({ client } = await createPlatformClient(oldWorkspace as WorkspaceUuid, 30000)) // TODO: FIXME
+            ;({ client } = await createPlatformClient(oldWorkspace, 30000)) // TODO: FIXME
```
lexiv0re: Minor: is todo still relevant?

services/github/pod-github/src/platform.ts
```diff
@@ -388,150 +446,188 @@ export class PlatformWorker {
...
+                (it) => it.type === SocialIdType.GITHUB && it.value === dta?._id
+              )?._id
+              // We need to assign socialId to person in global account if missing and get it to match if exists.
+
+              if (githubSocialId === undefined) {
+                // We need to create a new social id for this account.
+                githubSocialId = await sysAccountClient.addSocialIdToPerson(
+                  person.personUuid as PersonUuid,
+                  SocialIdType.GITHUB,
+                  dta?._id ?? '',
```
lexiv0re, line 516: Is `dta._id` a GitHub `username` or `identifier`? We use only `usernames` in github social ids now, using identifiers may lead to social id duplication, it won't be possible to login using a GH social id with identifier and so on...

services/github/pod-github/src/worker.ts
```diff
@@ -458,221 +462,228 @@ export class GithubWorker implements IntegrationManager {
...
-  //   }
-  //   return person
-  // }
+  private async findPerson (userInfo: UserInfo, userName: string, uuid: PersonUuid): Promise<Ref<Person>> {
+    let person: Ref<Person> | undefined
+    // try to find by account.
+    if (userInfo.email != null && userInfo.email.trim().length > 0) {
+      const personAccount = await this.client.findOne(contact.class.SocialIdentity, {
+        type: SocialIdType.GITHUB,
+        value: userInfo.email
```
lexiv0re, line 474: Is `userInfo.email` a GitHub username? If it's an actual public email of the user should we look for `type: SocialIdType.EMAIL` here?

plugins/contact-resources/src/utils.ts
haiodo: yep, I've asked you a question if it possible to have employee without social id, if not possible then will do.

services/github/pod-github/src/platform.ts
haiodo: Just one for now.

services/github/pod-github/src/platform.ts
haiodo: Probable not, it was for a while :-)

services/github/pod-github/src/platform.ts
haiodo: username = login (haiodo) for me.

services/github/pod-github/src/worker.ts
haiodo: Will check

plugins/contact-resources/src/utils.ts
lexiv0re: It should not be possible but we can still combine them just in case so `personById = idMap(all employees + all persons from social ids)`

services/github/pod-github/src/platform.ts
lexiv0re: All good then

lexiv0re (dismissed):

services/github/pod-github/src/users.ts
```diff
@@ -47,23 +65,67 @@ export class UserManager {
...
 
   async removeUser (login: string): Promise<void> {
     this.userCache.clear()
     this.refUserCache.clear()
-    await this.usersCollection.deleteOne({ _id: login })
-  }
 
-  getAllUsers (): FindCursor<GithubUserRecord> {
-    return this.usersCollection.find({})
+    const secerts = await this.accountClient.listIntegrationsSecrets({ kind: 'github-user', key: login })
```
lexiv0re, line 126: Minor: typo

lexiv0re (approved):

services/github/pod-github/src/users.ts
haiodo: Will fix in next PR.

PR #8469, opened 2025-04-04
UBERF-9734: Set default account timezone

ArtyomSavchenko:


packages/account-client/src/client.ts
```diff
@@ -258,7 +261,8 @@ class AccountClientImpl implements AccountClient {
   async loginOtp (email: string): Promise<OtpInfo> {
     const request = {
       method: 'loginOtp' as const,
-      params: { email }
+      params: { email },
+      headers: getTimezoneHeader()
```
lexiv0re: Maybe we should just include this header in `rpc` function for all requests?

server/account/src/operations.ts
```diff
@@ -2122,6 +2137,21 @@ export async function addSocialIdToPerson (
...
 }
 
+export async function getAccountInfo (
+  ctx: MeasureContext,
+  db: AccountDB,
+  branding: Branding | null,
+  token: string,
+  params: { accountId: PersonUuid }
+): Promise<AccountInfo> {
+  const { accountId } = params
```
lexiv0re: Let's only allow authorized users to make this request for now until we have a use case that'd require an unauthorized call. You can simply call `decodeTokenVerbose(ctx, token)` for that.

server/account/src/utils.ts
```diff
@@ -118,7 +119,8 @@ export function wrap (
     request: any,
     token?: string
   ): Promise<any> {
-    return await accountMethod(ctx, db, branding, token, { ...request.params })
+    const timezone = request.headers !== undefined && request.headers['X-Timezone']
+    return await accountMethod(ctx, db, branding, token, { ...request.params }, { timezone })
```
lexiv0re: We need to adjust `accountMethod` interface. Actually, it hasn't been adjusted to the previous change as well.

server/account/src/types.ts
```diff
@@ -281,3 +281,7 @@ export interface MailboxOptions {
   maxNameLength: number
   maxMailboxCount: number
 }
+
+export interface LoginMeta {
```
lexiv0re: Let's call it just `Meta` as we may want to add more details to it later not related to login. Even timezone not really related to login :)

server/account/src/utils.ts
```diff
@@ -1337,3 +1339,24 @@ export async function getWorkspaceRole (
...
+  meta?: LoginMeta
+): Promise<void> {
+  try {
+    if (meta?.timezone === undefined) return
+    const existingAccount = account ?? (await db.account.findOne({ uuid: accountId }))
+    if (existingAccount === undefined || existingAccount === null) {
+      ctx.warn('Failed to find account')
+      return
+    }
+    if (existingAccount?.timezone != null) return
```
lexiv0re: Minor: `?` should not be needed here

server/account/src/operations.ts
ArtyomSavchenko: Thanks, fixed

aonnikov (approved):

PR #8493, opened 2025-04-08
fix: handle token error in collaborator service

aonnikov:


BykhovDenis (approved):

haiodo:
@aonnikov do we need it into stating?

aonnikov:
> @aonnikov do we need it into stating?

It is a minor issue that affects sentry only, but we can add it staging to reduce number of invalid issues reported

PR #8498, opened 2025-04-08
Cherry Pick of recent changes in qms

mr1name:


plugins/setting-resources/src/plugin.ts
```diff
@@ -118,6 +118,31 @@ export default mergeIds(settingId, setting, {
     StartOfTheWeek: '' as IntlString,
     SystemSetupString: '' as IntlString,
     DefaultString: '' as IntlString,
-    WorkspaceNamePattern: '' as IntlString
+    WorkspaceNamePattern: '' as IntlString,
+    Export: '' as IntlString,
```
lexiv0re: Why export is a part of this PR?

mr1name: No idea. Removed

lexiv0re (approved):

PR #8500, opened 2025-04-08
EQMS-1302: fixed RBAC bypass for space / team related wizards and popups

mr1name:


plugins/view-resources/src/utils.ts
```diff
@@ -1480,21 +1480,36 @@ async function getAttrEditor (key: KeyedAttribute, hierarchy: Hierarchy): Promis
...
   return (store.whitelist.has(space) || store.ps[space]?.has(_id)) ?? false
 }
 
+export function getPermittedAccounts (
+  _id: Ref<Permission>,
+  space: Ref<TypedSpace>,
+  store: PermissionsStore
+): Array<Ref<Account>> {
+  const arePermissionsDisabled = getMetadata(core.metadata.DisablePermissions) ?? false
+  if (arePermissionsDisabled) return Array.from(store.ms[space] ?? [])
```
lexiv0re, line 1503: It doesn't seem right to return all members' accounts here only to add them to the query filter instead of just omitting this filter at all when everyone is permitted. Can we improve this?
Then we also won't need to track space members in permissions store which also doesn't look right in the current solution.

mr1name: I'm not sure I'm understanding correctly what you're suggesting to do instead

lexiv0re: From what I can see you
1. Keep all spaces' members in the store
2. When permission checks are disabled you return all members of the space
3. This array of members then used in queries with filter like:
```
{
  _id: { $in: allMembers },
  ...otherFilters
}
```

What I'm suggesting is to add special handling to this case and instead of doing 1-3 just query with filter if permission checks are disabled:
```
{
  ...otherFilters
}
```

mr1name: This would imply that each place that tries to get a list of available users would manually do a check to see if the permissions have been disabled. Why not do this in one place?

lexiv0re: Ofc there are cons and pros but in general queries with `IN [very_long_array_of_values]` scale pretty badly and here we will be artificially creating a bunch of them. Imagine a space (e.g. a community channel) with 5000 members - I'm sure there will be issues with such queries.
For your concern I think it will be possible to create extra utilities on top of that to cover the most popular cases and  not to repeat the same logic everywhere and that'd worth the benefits we'll get for large spaces.

lexiv0re (approved):

PR #8523, opened 2025-04-10
UBERF-9759 Scroll attachments in chat

aonnikov:


plugins/attachment-resources/src/components/AttachmentPreviewPopup.svelte
```diff
@@ -0,0 +1,83 @@
...
+        limit: 1
+      }
+    )
+    const result = []
+    if (drawings !== undefined) {
+      for (const drawing of drawings) {
+        result.push(drawing)
+      }
+    }
+    return result
```
ArtyomSavchenko: [Minor] Can we use Array.from?
`return Array.from(drawings ?? [])`

aonnikov: Yep

Chunosov (approved):

haiodo (approved):

PR #8539, opened 2025-04-14
Basic rate limits

haiodo:
Basic rate limits

### DIFF
--- a/packages/api-client/src/rest/rest.ts
+++ b/packages/api-client/src/rest/rest.ts
@@ -42,8 +42,19 @@ export function createRestClient (endpoint: string, workspaceId: string, token: 
   return new RestClientImpl(endpoint, workspaceId, token)
 }
 
+const rateLimitError = 'rate-limit'
+
+function isRLE (err: any): boolean {
+  return err.message === rateLimitError
+}
+
 export class RestClientImpl implements RestClient {
   endpoint: string
+
+  slowDownTimer = 0
+
+  remaining: number = 1000
+  limit: number = 1000
   constructor (
     endpoint: string,
     readonly workspace: string,
@@ -85,10 +96,15 @@ export class RestClientImpl implements RestClient {
     const result = await withRetry(async () => {
       const response = await fetch(requestUrl, this.requestInit())
       if (!response.ok) {
+        await this.checkRateLimits(response)
         throw new PlatformError(unknownError(response.statusText))
       }
       return await extractJson<FindResult<T>>(response)
-    })
+    }, isRLE)
+
+    if (result.error !== undefined) {
+      throw new PlatformError(result.error)
+    }
 
     if (result.lookupMap !== undefined) {
       // We need to extract lookup map to document lookups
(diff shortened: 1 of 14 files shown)


packages/api-client/src/rest/rest.ts
```diff
@@ -124,6 +140,25 @@ export class RestClientImpl implements RestClient {
...
+    if (response.status === 429) {
+      // Extract rate limit information from headers
+      const retryAfter = response.headers.get('Retry-After')
+      const rateLimitReset = response.headers.get('X-RateLimit-Reset')
+      // const rateLimitLimit: string | null = response.headers.get('X-RateLimit-Limit')
+      const waitTime =
+        retryAfter != null
+          ? parseInt(retryAfter, 10)
+          : rateLimitReset != null
+            ? new Date(parseInt(rateLimitReset, 10)).getTime() - Date.now()
```
aonnikov: it should not be 10 here

aonnikov: Maybe do something like this:
`Math.max(new Date(parseInt(rateLimitReset, Date.now())).getTime() - Date.now()), 10)`

packages/api-client/src/rest/rest.ts
```diff
@@ -124,6 +140,25 @@ export class RestClientImpl implements RestClient {
...
+            : 1000 // Default to 1 seconds if no headers are provided
+
+      console.warn(`Rate limit exceeded. Waiting ${waitTime / 1000} seconds before retrying...`)
+      await new Promise((resolve) => setTimeout(resolve, waitTime))
+      throw new Error(rateLimitError)
```
aonnikov, line 158: throw after waiting for timeout? I think we should either throw an error or wait, not both. If we waited for XX seconds, then we can safely do requests again

aonnikov, line 158: Should it be `PlatformError`?

aonnikov (dismissed):

packages/api-client/src/rest/rest.ts
haiodo: Do not understand you and provided may be code is some really weird stuff. Especially parseInt by Date.now() radix
`parseInt(rateLimitReset, Date.now())`

packages/api-client/src/rest/rest.ts
haiodo: Nope, it will never will be propagated to clients.

haiodo: It is used with withRetry and require exception to retry.

packages/api-client/src/rest/rest.ts
aonnikov: Yeah, Somehow I decided that it is default value :)

aonnikov (approved):

packages/api-client/src/rest/rest.ts
```diff
@@ -160,16 +195,23 @@ export class RestClientImpl implements RestClient {
...
-      keepalive: true,
-      body: JSON.stringify(tx)
-    })
-    if (!response.ok) {
-      throw new PlatformError(unknownError(response.statusText))
+    const result = await withRetry(async () => {
+      const response = await fetch(requestUrl, {
+        method: 'POST',
+        headers: this.jsonHeaders(),
+        keepalive: true,
```
ArtyomSavchenko, line 202: Should we add keep-alive headers to `this.jsonHeaders` as well?

ArtyomSavchenko (approved):

kristina-fefelova (approved):

packages/api-client/src/rest/rest.ts
haiodo: This is not this PR intend to check for it :-)

PR #8542, opened 2025-04-14
feat: Press ArrowUp to edit last message in chats / threads

### LINKED ISSUE #8359: Press Up to edit previous message in Chat
It would be nice if pressing Up in the text field in Chat allowed editing the previous message. This feature is widely supported in many messengers.

n1cc00:
This adds a keyboard shortcut (ArrowUp) for the chat's message input field if the field is empty. It works for both chat and thread messages.

Closes #8359 

This is my first PR ever and I don't know too much about svelte so I hope this is useful

Edit: I didn't mean to try to merge instantly :D


packages/ui/src/components/Component.svelte
```diff
@@ -120,6 +120,8 @@
           on:validate
           on:submit
           on:select
+          on:keydown
```
aonnikov, line 123: these are too specific, please try to avoid using them, especially `editingEnded`

PR #8574, opened 2025-04-16
Fix TypedSpaces empty owners

utkaka:
If a TypedSpace was created without an owner we should assign one

### DIFF
--- a/server-plugins/contact-resources/src/index.ts
+++ b/server-plugins/contact-resources/src/index.ts
@@ -45,7 +45,7 @@ import core, {
   AccountUuid,
   TxCreateDoc,
   SortingOrder,
-  MarkupBlobRef
+  MarkupBlobRef, AccountRole, systemAccountUuid, TypedSpace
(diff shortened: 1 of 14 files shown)


server-plugins/contact-resources/src/index.ts
```diff
@@ -115,6 +118,56 @@ export async function OnEmployeeCreate (_txes: Tx[], control: TriggerControl): P
...
+    if (owners.length === 0 || (owners.length === 1 && owners[0] === systemAccountUuid)) {
+      if (employees === undefined) {
+        employees = await control.findAll(
+          control.ctx,
+          contact.mixin.Employee,
+          { role: AccountRole.User },
+          { sort: { createdOn: SortingOrder.Ascending } }
+        )
+      }
+      const account = employees?.find(
```
BykhovDenis: Maybe we can take just first member?

BykhovDenis (approved):

PR #8606, opened 2025-04-17
Restrict project creation

utkaka:
Allow projects to be created only for the Maintainer and Owner roles, or if the user role has permission to create projects.


BykhovDenis (approved):

models/core/src/spaceType.ts
```diff
@@ -24,7 +24,7 @@ const roles = [
   {
     _id: core.role.Admin,
     name: 'Admin',
-    permissions: [core.permission.UpdateObject, core.permission.DeleteObject]
+    permissions: [core.permission.UpdateObject, core.permission.DeleteObject, core.permission.CreateProject]
```
lexiv0re, line 27: Minor: @utkaka not really consistent with the existing naming.

utkaka: Would it be better to call it CreateTrackerSpace since it only applies to Tracker projects?

lexiv0re: No, it's in `core` so it knows nothing about `Tracker` or any other specific spaces. As it's a spaces' space its objects are spaces themselves so e.g. the existing `core.permission.UpdateObject` applied to this space means the permission of updating spaces. Here adding the standard `core.permission.CreateObject` would mean creating spaces (in tracker, documents, any other plugin).

PR #8615, opened 2025-04-17
Controlled documents import fixes

akhismat:


packages/importer/src/huly/cards.ts
```diff
@@ -81,20 +81,22 @@ export class CardsProcessor {
...
-            $position: 0
+    if (typesRefs.length > 0) {
+      const updateDefaultSpace: UnifiedUpdate<CardSpace> = {
+        _class: card.class.CardSpace,
+        _id: 'card:space:Default' as Ref<CardSpace>,
+        space: core.space.Model,
+        props: {
+          $push: {
+            types: {
+              $each: [...new Set(typesRefs)],
```
BykhovDenis, line 92: Why we need $each, instead of `$push: { types: [...new Set(typesRefs)] }`

akhismat: image

Actually can't do that - the type system requires `$each` here. As you can see from the error, direct array assignment isn't compatible with the expected type `Ref<MasterTag> | Position<Ref<MasterTag>> | undefined`. We'd need to modify core types to make it work:

The `DocumentUpdate` type uses `PushOptions` which expects either a single value or a `Position` object with `$each` and `$position` for array operations. This is part of our core types in `tx.ts`, 

For now, I think it's safer to stick with the current implementation. Let me know if you'd like to explore changing the base types though.

lexiv0re (approved):

BykhovDenis (approved):

PR #8665, opened 2025-04-22
UBERF-10303: Always sign up with OTP

### LINKED ISSUE #8647: Keep only OTP sign up

lexiv0re:
* Uses OTP sign up in production
* Fallbacks to the old sign up when MAIL_URL is not configured (primarily for dev & tests)

Related to: https://front.hc.engineering/workbench/platform/tracker/UBERF-10303
Closes https://github.com/hcengineering/platform/issues/8647


server/account/src/operations.ts
```diff
@@ -198,6 +198,8 @@ export async function loginOtp (
 /**
  * Given an email, password, first name, and last name, creates a new account and sends a confirmation email.
  * The email confirmation is not required if the email service is not configured.
+ *
+ * ---------DEPRECATED. Only to be used for dev setups without mail service. Use signUpOtp instead.
```
haiodo, line 202: May be we could use OTP key 0000 if mail service is not configured in dev setup?

haiodo (dismissed):

server/account/src/operations.ts
lexiv0re: I thought about it but it had some drawbacks. The main one is that the same OTP check method is used for both login and sign up and it would be possible to login with 000000 code as any existing account by just knowing the email which is not good for self-hosting setups that don't have the mail service configured yet. This way we at least keep them usable. Plus we keep the otp handling code free from any dev/test hacks which makes them more reliable.

haiodo (approved):

PR #8668, opened 2025-04-22
EQMS-1475: space browser for qms documents

mr1name:


lexiv0re:
@mr1name does it correctly handle the permissions of who can change members/owners from the new table?

mr1name:
@lexiv0re yes

lexiv0re (approved):

PR #8729, opened 2025-04-28
UBERF-10376: Allow todos filtering

haiodo:
image
image


BykhovDenis (approved):

aonnikov:
OMG so we filter and allow to create todo in the same input? This can be confusing ...

aonnikov (approved):

PR #8750, opened 2025-04-28
fix checking for guest in meetings

Chunosov:


plugins/love-resources/src/utils.ts
```diff
@@ -956,7 +956,9 @@ export async function tryConnect (
     await client.update(invite, { status: invite.room === room._id ? RequestStatus.Approved : RequestStatus.Rejected })
   }
 
-  const isGuest = (currentPerson as Employee).role === AccountRole.Guest
+  const isGuest =
+    (currentPerson as Employee).role === AccountRole.Guest ||
+    client.getHierarchy().as(currentPerson, contact.mixin.Employee).role === AccountRole.Guest
```
haiodo:
```suggestion
  const isGuest = client.getHierarchy().as(currentPerson, contact.mixin.Employee).role === AccountRole.Guest
```

haiodo: as check if it is already Employee and cast if it is required.

aonnikov (approved):

PR #8832, opened 2025-05-05
Update API client documentation

aonnikov:
1. Add information about REST client
2. Update examples to use huly.app instead of localhost
3. Minor changes


n1kolasM:
Documents/Collections/Mixins API are unavailable in Rest client without `createRestTxOperations`, which again deviates from `connect*` simplified path

haiodo (approved):

PR #8871, opened 2025-05-07
Emoji node

utkaka:


packages/presentation/src/components/markup/NodeContent.svelte
```diff
@@ -78,11 +79,21 @@
     {/if}
   {:else if node.type === MarkupNodeType.text}
     {node.text}
+  {:else if node.type === MarkupNodeType.emoji}
+    <span class="emoji" class:emojiOnly={single}>
+      {#if node.attrs?.kind === 'custom'}
```
aonnikov: Maybe use 'image' instead of 'custom'. We don't care if it is custom or not, it is more important if it is image or unicode.

packages/ui-next/src/utils.ts
```diff
@@ -30,15 +31,15 @@ export const defaultMessageInputActions: TextInputAction[] = [
...
+        emojiPlugin.component.EmojiPopup,
         {},
         element,
         (emoji) => {
           if (emoji === null || emoji === undefined) {
             return
           }
 
-          editorHandler.insertText(emoji)
+          editorHandler.insertText(emoji.text)
```
aonnikov: We should not insert emoji as text, I suggest to add separate method to insert emoji as node.

plugins/emoji-assets/lang/ru.json
```diff
@@ -0,0 +1,4 @@
+{
+  "string": {
```
aonnikov, line 2: I see some strings in emoji plugin but don't see translations here

plugins/emoji-resources/src/components/SkinToneTooltip.svelte
```diff
@@ -4,16 +4,18 @@
...
+  import { getEmojiSkins } from '../utils'
+
+  export let emoji: Emoji.Emoji
   export let selected: number
 
   const dispatch = createEventDispatcher()
 
-  const skins: string[] = generateSkinToneEmojis(getEmojiCode(emoji))
+  const emojiSkins = getEmojiSkins(emoji)
+  const skins: Emoji.Emoji[] = emojiSkins !== undefined ? [emoji, ...emojiSkins] : []
```
ArtyomSavchenko: Looks like reactivity is missing here:
```
$: emojiSkins = getEmojiSkins(emoji)
```

plugins/emoji-resources/src/store.ts
```diff
@@ -0,0 +1,107 @@
...
+          }
+          const skinEmoji = emoji.skins?.find((s) => s.hexcode === parsedEmoji.hexcode)
+          if (skinEmoji === undefined) return
+          result.push({ ...skinEmoji, key: '' })
+        })
+      }
+    })
+    return result
+  } catch (e) {
+    console.error(e)
```
ArtyomSavchenko: [Minor] Suggest to add error message:
`console.error('Failed to parse emojis', e)`

plugins/emoji-resources/src/store.ts
```diff
@@ -0,0 +1,107 @@
...
+      )
+    }
+  }
+}
+export const addFrequentlyEmojis = (emoji: EmojiWithGroup): void => {
+  if (emoji === undefined) return
+  const hexcode = isCustomEmoji(emoji) ? emoji.shortcode : emoji.hexcode
+
+  const frequentlyEmojisKey = getEmojisLocalStorageKey()
+  const frequentlyEmojis = window.localStorage.getItem(frequentlyEmojisKey)
```
ArtyomSavchenko: [Minor] Looks like code to get and parse emoji can be reused in several places:
```
function getStorageItem<T>(key: string, defaultValue: T): T {
  const storageKey = getEmojisLocalStorageKey(key)
  const item = window.localStorage.getItem(storageKey)
  if (item == null) return defaultValue
  
  try {
    return JSON.parse(item)
  } catch (e) {
    console.error(`Failed to parse ${key}:`, e)
    return defaultValue
  }
}

File: packages/text/src/nodes/emoji.ts [hunk 7]
```diff
@@ -0,0 +1,144 @@
...
+              type: this.name,
+              attrs: { emoji, kind, image }
+            })
+          }
+    }
+  },
+
+  parseHTML () {
+    return [
+      {
```
aonnikov, line 68: You should be able to parse image nodes as well.

packages/text/src/nodes/emoji.ts
```diff
@@ -0,0 +1,144 @@
...
+        'span',
+        mergeAttributes(
+          {
+            'data-type': this.name,
+            class: 'emoji'
+          },
+          HTMLAttributes
+        ),
+        [
+          'img',
```
aonnikov, line 123: Can we put image inside span? If we use `img` tag then we have conflict with image node.

aonnikov (approved):

PR #8885, opened 2025-05-08
Fix contact check

BykhovDenis:


plugins/recruit-resources/src/components/CreateCandidate.svelte
```diff
@@ -525,7 +525,12 @@
     ]
   }
 
-  $: if (object.firstName != null && object.lastName != null) {
+  $: if (
```
haiodo, line 528: Alex staff asked some time ago to allow to create a person without first name or last name.

haiodo (approved):

PR #8901, opened 2025-05-12
Saving the MediaPopup position when leaving

SasLord:
https://github.com/user-attachments/assets/91af138e-b930-4304-ad9b-b75bb5758494


aonnikov:
You should close popup, not keep it open. I think we can return `dispatch('close')`

aonnikov (approved):

PR #8913, opened 2025-05-13
pods / external

aav:
pods/external implementation, as defined in:

https://app.hc.engineering/workbench/platform/document/external-services-build-682232fa8c2a2eb48ffe47d5

### DIFF
--- a/pods/external/package.json
+++ b/pods/external/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "_phase:docker-build": "rushx docker:build",
     "docker:build": "sh bin/build.sh",
-    "docker:push": "sh bin/push.sh"
+    "docker:push": "sh bin/push.sh",
+    "format": "true"
   }
 }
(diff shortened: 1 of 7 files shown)


pods/external/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@hcengineering/pod-external",
+  "version": "0.6.0",
+  "author": "Anticrm Platform Contributors",
+  "license": "EPL-2.0",
+  "scripts": {
+    "_phase:docker-build": "rushx docker:build",
+    "docker:build": "sh bin/build.sh",
```
aonnikov, line 8: nit: add a shebang to the script and just call it here, no need to hardcode `sh` command

aonnikov (approved):

pods/external/package.json
aav: this will require an 'x' permission on scripts, i deliberately avoided that, because they are not user invocable under normal conditions

PR #8914, opened 2025-05-13
fix: close room popup on stop share

aonnikov:


haiodo (approved):

haiodo:
why only on stop?

aonnikov:
> why only on stop?

actually on both share and stop sharing

PR #8933, opened 2025-05-14
Popups and Tooltips in full screen mode

SasLord:
**Popups:**
screen-fs-popups

**Tooltips:**
screen-fs-tooltips


haiodo (changes requested):
I do not understand why we need this hacks.

haiodo (approved):

PR #8943, opened 2025-05-15
UBERF-10599: Fix ws not found in gmail

ArtyomSavchenko:
Fix `Cannot find module 'ws'` in gmail container
ws is not included in bundle, since api-client cannot resolve ws from optional dependency in this case
So installed ws in dockerfile as external dependency

### DIFF
--- a/packages/api-client/src/socket/node.ts
+++ b/packages/api-client/src/socket/node.ts
@@ -19,7 +19,12 @@ import { type ClientSocket, type ClientSocketFactory } from '@hcengineering/clie
 export const NodeWebSocketFactory: ClientSocketFactory = (url: string): ClientSocket => {
   // We need to override default factory with 'ws' one.
   // eslint-disable-next-line
-  const WebSocket = require('ws')
+  let WebSocket
+  try {
+    WebSocket = require('ws')
+  } catch (error) {
+    throw new Error('The "ws" package is required for NodeWebSocketFactory. ')
+  }
   type WebSocketData = Parameters<typeof ws.on>[1]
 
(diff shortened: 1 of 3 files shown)


aonnikov (dismissed):

BykhovDenis:
Do we really need it? (I mean other service works without it somehow)

aonnikov:
> Do we really need it? (I mean other service works without it somehow)

I'm  sure we don't need it but I don't know what is the real problem, this is a quick fix as I understand

haiodo:
why not add dependency to your pod directly?, github service works fine with it for example

haiodo:
@BykhovDenis they not;-) they have it in package.json deps

ArtyomSavchenko:
> why not add dependency to your pod directly?, github service works fine with it for example

I've tried to add direct dependency as well and it works for server-client and other packages with direct ws dependency, but in api-client there was an error during bundle build:
```
Build failed: Build failed with 1 error:
../../../packages/api-client/lib/socket/node.js:26:28: ERROR: Could not resolve "ws"
Error: Failed calling node ../../../common/scripts/esbuild.js.  Exit code: 1
```
I think it is caused by the fact that ws dependency is optional in api-client and the issue is not observed if mark it as mandatory dependency. It is optional dependency, since api-client can be used both in node.js and browser

ArtyomSavchenko:
I've updated the PR to use alternative solution:

1. Direct dependencies are used now instead of external
2. In api-client check ws dependency in runtime, looks like it is ok for optional dependency that depends on platform

aonnikov (approved):

PR #8967, opened 2025-05-16
add {PROVIDER}_DISPLAY_NAME env-var 

n1cc00:
This environment variable is added to the front-service to allow setting a custom provider name for OpenId Login button:

huly_openiddisplayname

### DIFF
--- a/packages/core/src/classes.ts
+++ b/packages/core/src/classes.ts
@@ -869,4 +869,9 @@ export interface AccountInfo {
   locale?: string
 }
 
+export interface ProviderInfo {
+  name: string
+  displayName?: string
+}
+
 export type SocialKey = Pick<SocialId, 'type' | 'value'>
--- a/packages/core/src/classes.ts
+++ b/packages/core/src/classes.ts
@@ -869,9 +869,4 @@ export interface AccountInfo {
   locale?: string
 }
 
-export interface ProviderInfo {
-  name: string
-  displayName?: string
-}
-
 export type SocialKey = Pick<SocialId, 'type' | 'value'>
--- a/plugins/login-resources/src/components/Providers.svelte
+++ b/plugins/login-resources/src/components/Providers.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { concatLink } from '@hcengineering/core'
+  import { concatLink, type ProviderInfo } from '@hcengineering/core'
   import { getMetadata } from '@hcengineering/platform'
   import { AnySvelteComponent, Button, Grid, deviceOptionsStore, getCurrentLocation } from '@hcengineering/ui'
   import { onMount } from 'svelte'
@@ -12,28 +12,41 @@
   interface Provider {
     name: string
     component: AnySvelteComponent
+    displayName: null
   }
 
   const providers: Provider[] = [
     {
       name: 'google',
-      component: Google
+      component: Google,
+      displayName: null
     },
     {
       name: 'github',
-      component: Github
+      component: Github,
+      displayName: null
     },
     {
       name: 'openid',
-      component: OpenId
+      component: OpenId,
+      displayName: null
     }
   ]
 
   let enabledProviders: Provider[] = []
 
   onMount(() => {
-    void getProviders().then((res) => {
-      enabledProviders = providers.filter((provider) => res.includes(provider.name))
+    void getProviders().then((res: ProviderInfo[]) => {
+      enabledProviders = providers
+        .map((provider) => {
+          const match = res.find((r) => r.name === provider.name)
+          if (!match) return null
+          return {
+            ...provider,
+            displayName: match.displayName
+          }
+        })
+        .filter((p): p is Provider & { displayName: string } => p !== null)
     })
   })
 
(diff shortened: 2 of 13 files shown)


plugins/login-resources/src/components/providers/OpenId.svelte
```diff
@@ -2,9 +2,12 @@
   import { Label } from '@hcengineering/ui'
   import OpenId from '../icons/OpenId.svelte'
   import login from '../../plugin'
+  import { getMetadata } from '@hcengineering/platform'
+
+  const openIdDisplayName = getMetadata(login.metadata.DisableSignUp) ?? 'OpenId'
```
lexiv0re: Wrong metadata being read?

server/front/src/starter.ts
```diff
@@ -119,6 +119,8 @@ export function startFront (ctx: MeasureContext, extraConfig?: Record<string, st
 
   const disableSignUp = process.env.DISABLE_SIGNUP
 
+  const openIdDisplayName = process.env.OPENID_DISPLAY_NAME
```
lexiv0re: Would it be better to define this variable in the accounts service along with all other variables defining the provider and return it along with the providers' ids in the `/providers` request? It'd help to simplify things:
1. The whole provider entity is defined in one place
2. Handled with one piece of code for both the web ui and the desktop
3. Will also be available to any other clients besides the web ui and the desktop (e.g. automation, custom integrations, etc.)

lexiv0re (changes requested):
Thanks for the contribution. Could you please also attach a screenshot of how it looks when configured with a custom name.

plugins/login-resources/src/components/providers/OpenId.svelte
n1cc00: Yup. Must have hit Ctrl-Z once too much there 😄

server/front/src/starter.ts
n1cc00: Makes sense. I still committed a fix for the current approach for now and will have a look into this next week

n1cc00: Okay so I made a new version including it in the /providers request. I added the functionality to GitHub and google providers too for the sake of consistency (GOOGLE_DISPLAY_NAME, GITHUB_DISPLAY_NAME). Even though it probably won't really be used there.

packages/core/src/classes.ts
```diff
@@ -869,4 +869,9 @@ export interface AccountInfo {
   locale?: string
 }
 
+export interface ProviderInfo {
```
lexiv0re: Feels like it doesn't belong to `core`. Please define it both in `auth-providers` and in `account-client` instead.

plugins/login-resources/src/components/Providers.svelte
```diff
@@ -12,28 +12,41 @@
...
+      enabledProviders = providers
+        .map((provider) => {
+          const match = res.find((r) => r.name === provider.name)
+          if (!match) return null
+          return {
+            ...provider,
+            displayName: match.displayName
+          }
+        })
+        .filter((p): p is Provider & { displayName: string } => p !== null)
```
lexiv0re: Minor: there's a `notEmpty` utility which can simplify this line to `filter(notEmpty)`

plugins/login-resources/src/components/Providers.svelte
```diff
@@ -12,28 +12,41 @@
...
     component: AnySvelteComponent
+    displayName: null
   }
 
   const providers: Provider[] = [
     {
       name: 'google',
-      component: Google
+      component: Google,
+      displayName: null
```
lexiv0re: Minor: `null`s looks strange here. Please consider replacing this structure with a simple mapping of `name` -> `component` and only use `Provider` type for `enabledProviders`.

plugins/login-resources/src/components/providers/OpenId.svelte
```diff
@@ -2,9 +2,12 @@
   import { Label } from '@hcengineering/ui'
   import OpenId from '../icons/OpenId.svelte'
   import login from '../../plugin'
+  import { getMetadata } from '@hcengineering/platform'
+
+  const openIdDisplayName = getMetadata(login.metadata.OpenIdDisplayName) ?? 'OpenId'
```
lexiv0re: Shouldn't it be taken from the new parameter now?

lexiv0re (changes requested):

plugins/login-resources/src/components/Providers.svelte
n1cc00: I see - after revision I think you can leave it out completely

packages/core/src/classes.ts
n1cc00: Wouldn't it be better to only define it in account-client package to avoid doubling? For now, I just imported it from there In Providers.svelte and in authProviders

n1cc00:
Hm, can someone help me why formatting check failed? When I format it with eslint it makes no changes to the file, it all seems correct.

n1cc00:
I have a question about formatting: I tried running ESLint, but it didn’t change Providers.svelte, and running Prettier completely changed the files in the wrong way (not following the correct formatting style). So for now, I applied the changes manually. What’s the correct command to fix files properly ?

lexiv0re:
It's `rush fast-format` for the whole repo or `rushx format` for a particular package.

lexiv0re (approved):
LGTM. Thanks!

PR #8968, opened 2025-05-16
UBERF-10632: Fix email thread creation date

ArtyomSavchenko:


services/mail/mail-common/src/message.ts
```diff
@@ -247,7 +248,8 @@ async function saveMessageToSpaces (
         modifiedBy,
         MessageType.Message,
         {
-          created: createdDate
+          created: createdDate,
+          edited: createdDate
```
kristina-fefelova: Looks extra, is it used?

kristina-fefelova: `await msgClient.createMessage( threadId,  chat.masterTag.Thread,   content,  modifiedBy MessageType.Message, undefined, createdDate)` something like this?

haiodo (approved):

PR #9005, opened 2025-05-21
Readonly workspace access

utkaka:


plugins/setting-resources/src/components/Owners.svelte
```diff
@@ -27,6 +27,7 @@
   const currentAccount = getCurrentAccount()
 
   const items: DropdownIntlItem[] = [
+    { id: AccountRole.ReadOnlyGuest, label: setting.string.User },
```
haiodo: do we need another label here?

plugins/setting-resources/src/components/General.svelte
```diff
@@ -66,6 +75,8 @@
 
     oldName = res.name
     name = oldName
+    console.log(res.allowReadOnlyGuest)
```
haiodo:
```suggestion
```

plugins/workbench-resources/src/components/Workbench.svelte
```diff
@@ -158,6 +158,9 @@
   const apps: Application[] = client
     .getModel()
     .findAllSync<Application>(workbench.class.Application, { hidden: false, _id: { $nin: excludedApps } })
+    .filter((it) => isAllowedToRole(it.accessLevel, account))
+
+  console.log(apps)
```
haiodo:
```suggestion
```

packages/ui/src/components/Button.svelte
```diff
@@ -72,6 +73,9 @@
   export let stopPropagation: boolean = true
   export let event: string | undefined = undefined
 
+  const account = getCurrentAccount()
```
haiodo: No way, No way, button should not know about anything from anywhere.

haiodo (changes requested):

packages/ui/src/components/Button.svelte
```diff
@@ -31,6 +31,7 @@
   import Label from './Label.svelte'
   import Spinner from './Spinner.svelte'
   import { Analytics } from '@hcengineering/analytics'
+  import { AccountRole, getCurrentAccount } from '@hcengineering/core'
```
haiodo:
```suggestion
```

plugins/notification-resources/src/components/BrowserNotificatator.svelte
```diff
@@ -85,14 +87,14 @@
...
-      await client.remove(value)
+      if (me?.role !== AccountRole.ReadOnlyGuest) await client.remove(value)
       return
     }
 
     const locObjectId = await getObjectIdFromLocation(getCurrentResolvedLocation())
 
     if (_id && _id === locObjectId) {
-      await client.remove(value)
+      if (me?.role !== AccountRole.ReadOnlyGuest) await client.remove(value)
```
haiodo: Let's remove all of such changes, since we do not send transactions in readonly guest mode at all.

plugins/workbench-resources/src/components/Applications.svelte
```diff
@@ -56,7 +56,11 @@
 
   $: topApps = apps.filter((it) => it.position === 'top')
   $: midApps = apps.filter(
-    (it) => !hiddenAppsIds.includes(it._id) && isAppAllowed(it, me) && it.position !== 'top' && it.position !== 'bottom'
+    (it) =>
+      !hiddenAppsIds.includes(it._id) &&
+      isAllowedToRole(it.accessLevel, me) &&
```
kristina-fefelova: Why role checked only for midApps?

utkaka: This is an old and redundant check (and which I forgot to remove) that is no longer needed because I've added a filter for all apps in Workbench. I'll remove this one.

plugins/login-resources/src/components/SelectWorkspace.svelte
```diff
@@ -193,7 +196,7 @@
           </div>
         {/each}
 
-        {#if workspaces.length === 0 && account?.token != null}
+        {#if workspaces.length === 0 && account?.token != null && !isReadOnlyGuest}
```
haiodo: It is bad to hide this buttons, we need to show info about registration is required to create a new workspace instead and button to go into registration/login page.

plugins/notification-resources/src/inboxNotificationsClient.ts
```diff
@@ -15,14 +15,14 @@
 import activity from '@hcengineering/activity'
 import {
   type Account,
-  type Client,
-  SortingOrder,
-  getCurrentAccount,
-  toIdMap,
   type Class,
+  type Client,
```
haiodo: Why we need this changes?

plugins/workbench-resources/src/components/AccountPopup.svelte
```diff
@@ -163,7 +165,7 @@
         label: setting.string.Signout,
         action: async () => {
           await logOut()
-          navigate({ path: [loginId] })
+          navigate({ path: [loginId, 'login'] })
```
haiodo: why just login not work?

plugins/workbench-resources/src/connect.ts
```diff
@@ -86,7 +97,7 @@ export async function connect (title: string): Promise<Client | undefined> {
     )
     // something went wrong with selecting workspace with the selected token
     await logOut()
-    navigate({ path: [loginId] })
+    navigate({ path: [loginId, 'login'] })
```
haiodo: same why?

plugins/workbench-resources/src/components/AccountPopup.svelte
utkaka: It works without it, I've removed redudant 'login'

plugins/workbench-resources/src/connect.ts
utkaka: It works without it, I've removed redudant 'login'

plugins/notification-resources/src/inboxNotificationsClient.ts
utkaka: We don't, reverted

plugins/login-resources/src/components/SelectWorkspace.svelte
utkaka: Now we show them a signup button with a warning message

haiodo (approved):

PR #9009, opened 2025-05-21
Hide meeting minutes to guests

utkaka:


haiodo (approved):

haiodo:
Do you plan to add checks on server side as well?

BykhovDenis (approved):

BykhovDenis:
> Do you plan to add checks on server side as well?

I think we should just implement spaces for office.

PR #9081, opened 2025-05-24
UBERF-11111: Add retry package

ArtyomSavchenko:
Add common retry package to reuse in different places:

1. Implement retry logic in function (withRetry) and a method decorator (Retryable)
2. Support several delay strategies and jitter
3. Provide predefined filters for retryable errors and allow to define custom conditions

### DIFF
--- a/utils/retry/src/retry.ts
+++ b/utils/retry/src/retry.ts
@@ -71,25 +71,25 @@ export async function withRetry<T> (
   while (attempt <= config.maxRetries) {
     try {
       return await operation()
-    } catch (err: any) {
-      lastError = err
+    } catch (error: any) {
+      lastError = error
       const isLastAttempt = attempt >= config.maxRetries
 
       if (isLastAttempt) {
         logger.error(`${operationName} failed after ${attempt} attempts`, {
-          error: err.message,
+          error,
           attempt,
           maxRetries: config.maxRetries
         })
-        throw err
+        throw error
       }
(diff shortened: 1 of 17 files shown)


utils/retry/src/retry.ts
```diff
@@ -0,0 +1,146 @@
...
+  while (attempt <= config.maxRetries) {
+    try {
+      return await operation()
+    } catch (err: any) {
+      lastError = err
+      const isLastAttempt = attempt >= config.maxRetries
+
+      if (isLastAttempt) {
+        logger.error(`${operationName} failed after ${attempt} attempts`, {
+          error: err.message,
```
aonnikov: I think we should pass original error into the logger

utils/retry/src/retry.ts
```diff
@@ -0,0 +1,146 @@
...
+        attempt,
+        nextAttempt: attempt + 1,
+        delayMs: Math.round(actualDelay)
+      })
+
+      // Wait before retry
+      await sleep(actualDelay)
+
+      // Increase delay for next attempt (exponential backoff)
+      delayMs = Math.min(delayMs * config.backoffFactor, config.maxDelayMs)
```
aonnikov: Maybe extract this into separate class that implement some delay strategy (fixed/backoff/etc)

aonnikov (dismissed):

aonnikov (dismissed):

aonnikov (approved):

PR #9092, opened 2025-05-26
UBERF-11178 Migrate datalake workspace id

aonnikov:


lexiv0re (dismissed):

aonnikov:
Migrated manually, not needed for now

lexiv0re:
@aonnikov what guarantees the existence of `global_account` namespace in the DB before the migration is run? For this there must be some kind of dependency that the account's migrations are completed.

aonnikov:
Do you think it is needed? We can do this by waiting for account service to start, but I don't think it changes the situation  much.
There are three scenarios:
1. `global_account` already exists.
2. `global_account` does not exist but will be created soon - datalake will fail on migration, restart, and after few restarts will finish the migration. Probably we need to do better error handling here.
3. `global_account` does not exist and won't be created because account is in MongoDB or somewhere else. As far as I remember, we decided to not support this scenario.

lexiv0re:
Yes, I was concerned about 2. Won't there be a race condition when `global_account` is created already but workspaces migration is not finished yet and datalake might run its migration on partial data with some workspaces info missing?
3. yes, we decided not to support

aonnikov:
> Yes, I was concerned about 2. Won't there be a race condition when `global_account` is created already but workspaces migration is not finished yet and datalake might run its migration on partial data with some workspaces info missing? 3. yes, we decided not to support

If we wait for account service to be ready before starting datalake, will it work? I pushed changes that try to perform some /rpc to ensure the account service started.

lexiv0re:
Yes, if we wait for any rpc from account that'd guarantee that accounts' migrations are completed. Did you push the changes here? I cannot see them :)

aonnikov:
Yep, pushed the changes

lexiv0re (approved):

PR #9103, opened 2025-05-26
hulygram UI integration

aav:


plugins/telegram-resources/src/components/Connect.svelte
```diff
@@ -13,151 +13,236 @@
...
 
-  async function sendCode (): Promise<void> {
-    const res = await sendRequest('/signin/code', { phone, code })
-    if (res.next === 'pass') {
-      secondFactor = true
-    } else if (res.next === 'end') {
-      dispatch('close', { value: phone })
-    }
-  }
+  async function request (method: 'GET' | 'POST' | 'DELETE', path?: string, body?: any): Promise<any> {
```
kristina-fefelova: I would prefer to move api methods and types  in a separate file (api.ts for example) to make the component easier

plugins/telegram-assets/lang/en.json
```diff
@@ -1,5 +1,8 @@
 {
   "string": {
+    "Loading": "Loading...",
```
kristina-fefelova, line 3: It's better to add translation in all languages ​​at once so you don't forget :)

plugins/telegram-resources/src/components/Connect.svelte
```diff
@@ -13,151 +13,240 @@
...
+      primary?: { label: IntlString, handler?: () => any, disabled?: boolean }
+      secondary?: { label: IntlString, handler?: () => any }
+    }
+  }
+
+  let integration: Integration = 'Loading'
+  let state: UIState = { mode: 'Loading' }
+
+  $: {
+    function h (handler: () => Promise<Integration>) {
```
kristina-fefelova: it seems that it doesn't make sense to describe a function inside a reactive block

plugins/telegram-resources/src/components/Connect.svelte
```diff
@@ -13,151 +13,240 @@
...
+        hint: ex.message,
+
+        buttons: {
+          primary: { label: ui.string.Ok, handler: close }
+        }
+      }
+    }
   }
+
+  init().catch(console.error)
```
kristina-fefelova: You already have catch inside `init`

plugins/telegram-resources/src/components/Connect.svelte
```diff
@@ -13,151 +13,240 @@
...
+    <div class="tool" on:click={close}>
       <IconClose size={'small'} />
     </div>
   </div>
+
   <div class="content">
-    {#if secondFactor}
-      <p><Label label={telegram.string.PasswordDescr} /></p>
+    {#if state.mode === 'Loading'}
+      <p>
```
kristina-fefelova: Do you really need to wrap all with `<p>` ?

plugins/telegram-resources/src/components/Connect.svelte
aav: Does it have any practical (i.e. performance implications)? If not otherwise isn't more narrow visibility - better?

plugins/telegram-resources/src/components/Connect.svelte
aav: without this catch, the linter complains that "async function is not awaited" (or something like this)

plugins/telegram-resources/src/components/Connect.svelte
aav: Probably not

plugins/telegram-resources/src/components/Connect.svelte
kristina-fefelova: the function will be recreated each time when this block is entered, and updates will also be called for everything where there are references to this function. In this simple component this will not affect performance, but even visually, I want to simplify the code

plugins/telegram-resources/src/components/Connect.svelte
kristina-fefelova: `void init()`

plugins/telegram-resources/src/components/Connect.svelte
aav: ok, let it be so :)

plugins/telegram-resources/src/components/Connect.svelte
aav: done

lexiv0re (approved):

PR #9118, opened 2025-05-27
UBERF-8425: Improve parallel ws upgrade logging within one ws service

lexiv0re:
Related to: https://front.hc.engineering/workbench/platform/tracker/UBERF-8425

### DIFF
--- a/server/workspace-service/src/service.ts
+++ b/server/workspace-service/src/service.ts
@@ -38,6 +38,7 @@ import {
 import { generateToken } from '@hcengineering/server-token'
 import { FileModelLogger, prepareTools } from '@hcengineering/server-tool'
 import path from 'path'
+import { randomUUID } from 'crypto'
 
 import { Analytics } from '@hcengineering/analytics'
 import {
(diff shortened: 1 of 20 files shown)


server/workspace-service/src/service.ts
```diff
@@ -188,8 +189,9 @@ export class WorkspaceWorker {
         await this.doSleep(ctx, opt)
       } else {
         void this.exec(async () => {
+          const operationJobUuid = randomUUID()
           await ctx
-            .with('workspaceOperation', { mode: workspace.mode }, (ctx) =>
+            .with('workspaceOperation', { operationJobUuid, mode: workspace.mode }, (ctx) =>
```
haiodo: it is not logging, it will use huge set of memory, please use regular logging instead with context

haiodo (changes requested):

server/workspace-service/src/service.ts
lexiv0re: Thanks, fixed

haiodo (approved):

PR #9141, opened 2025-05-30
Support embed markup nodes in text-markdown

n1kolasM:


aonnikov (dismissed):

mr1name:
@n1kolasM Does this support conversion from Markdown back to a markup tree?

aonnikov:
> @n1kolasM Does this support conversion from Markdown back to a markup tree?

I don't think so, we may need to write some code so convert `<...>` style links to embeddings.

n1kolasM:
> @n1kolasM Does this support conversion from Markdown back to a markup tree?

I think embeds should not be parsed from markdown at all. Embed is already just a visual representation of a link

mr1name:
@n1kolasM The problem is that we need to support bi-directional sync. For example, if you paste an embed into an issue connected to GitHub, once it syncs, all embeddings will be converted back to links.

mr1name:
It definitely should be a link on the Markdown side, but you need to attach additional metadata (e.g., data-type="embed" or some marker within the link) to be able to parse it back properly

aonnikov:
Yeah, when converting markup -> markdown -> markup we should try to keep the original markup, so if we had link represented as an embedded element, after `markdown -> markup` conversion we should keep it as embedding. As I understand, it is not always possible because this information is lost when converted to markdown.

> For example, if you paste an embed into an issue connected to GitHub, once it syncs, all embeddings will be converted back to links.

Even worse, the integration may stop if we detect content that not convertible in both ways.

n1kolasM:
> @n1kolasM The problem is that we need to support bi-directional sync. For example, if you paste an embed into an issue connected to GitHub, once it syncs, all embeddings will be converted back to links.

Then we need our own syntax for embed markdowns, honestly, I don't see a problem in such behavior

mr1name:
> I don't see a problem in such behavior

@n1kolasM It will break sync with GitHub, and later it will break documents once we start using Markdown as the storage format — making the feature essentially obsolete.

mr1name:
So, you should either use inline HTML or add a marker to the link itself.

n1kolasM:
Ok, than we can use only  `<a>` links with additional attribute to indicate embbedded content

mr1name:
If you plan to use inline HTML, the only thing you’ll need to add is `data-type="embed"` to the `a` node. That's basically already the parse rule used by the plugin.

haiodo (dismissed):

aonnikov (approved):

mr1name (approved):

PR #9144, opened 2025-05-30
UBERF-8425: Improved pg/acc/ws error handling

lexiv0re:
Related to: https://front.hc.engineering/workbench/platform/tracker/UBERF-8425

### DIFF
--- a/server/postgres/src/storage.ts
+++ b/server/postgres/src/storage.ts
@@ -223,7 +223,7 @@ class ConnectionMgr {
             await client.execute('ROLLBACK;')
             console.error({ message: 'failed to process tx', error: err.message, cause: err })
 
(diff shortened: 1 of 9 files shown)


server/postgres/src/storage.ts
```diff
@@ -223,7 +223,7 @@ class ConnectionMgr {
             await client.execute('ROLLBACK;')
             console.error({ message: 'failed to process tx', error: err.message, cause: err })
 
-            if (err.code !== '40001' || tries === maxTries) {
+            if (!['40001', 'CONNECTION_CLOSED'].includes(err.code) || tries === maxTries) {
```
haiodo: why not also add other codes like in account?

haiodo (dismissed):

haiodo (approved):

PR #9160, opened 2025-06-03
QFIX: Rpc response body size

haiodo:


aonnikov (dismissed):

dev/tool/src/index.ts
```diff
@@ -1677,7 +1677,7 @@ export function devTool (
           name = systemAccountUuid
         }
         const wsByUrl = await db.workspace.findOne({ url: workspace })
-        const account = await db.socialId.findOne({ key: name })
+        const account = await db.socialId.findOne({ value: name })
```
lexiv0re, line 1680: Minor: ambiguous condition might accidentally return not the expected social id although should be a very rare case.

lexiv0re (approved):

dev/tool/src/index.ts
haiodo: It is just for tests mostly.

PR #9169, opened 2025-06-04
Add a blank line after Code block in a chat

rostislavnagimov:
The Issue:
When a code block was inserted into the chat, users were unable to create a new line or paragraph below it to continue typing. This made it difficult to compose the text directly after the code block.

Changes:
The text editor has been updated to automatically ensure an empty paragraph is present immediately after any inserted code block. This allows users to seamlessly click or navigate below the code block and continue typing.

### DIFF
--- a/plugins/text-editor-resources/src/components/extension/codeblock.ts
+++ b/plugins/text-editor-resources/src/components/extension/codeblock.ts
@@ -23,6 +23,8 @@ import { Plugin, PluginKey } from '@tiptap/pm/state'
 import { Decoration, DecorationSet, type EditorView } from '@tiptap/pm/view'
 import { createLowlight } from 'lowlight'
 import { isChangeEditable } from './editable'
+import { TextSelection } from '@tiptap/pm/state'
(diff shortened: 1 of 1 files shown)


mr1name:
If you press the down arrow at the end of a code block, the editor will create a new paragraph — so why do we need to create one manually? You’d still need to press the arrow key either way. In any case, this kind of enhancement should be handled at the cursor logic level, not at the content level. (Or at least for the editor in general, rather than for a specific block). Otherwise, we’d end up having to wrap every possible block with paragraphs.

rostislavnagimov:
> If you press the down arrow at the end of a code block, the editor will create a new paragraph — so why do we need to create one manually? You’d still need to press the arrow key either way. In any case, this kind of enhancement should be handled at the cursor logic level, not at the content level. (Or at least for the editor in general, rather than for a specific block). Otherwise, we’d end up having to wrap every possible block with paragraphs.

From my understanding down key is not fully obvious in that case, so i think would be better to make it simple and usable with a cursor click

PR #9179, opened 2025-06-05
Minor localization improvement

rostislavnagimov:
Requested in product feedback:

https://hulycommunity.slack.com/archives/C06BBP6630E/p1747299358572339


eliasko (commented):
Leave Employee as it was only change Collaborators. As mentioned in the issue:

“Employee(s)” is correctly translated as “Mitarbeiter”.

However, “Collaborator(s)” is also translated as “Mitarbeiter”, which is incorrect.
The correct translation should be “Mitwirkende” (plural) or “Mitwirkender” (singular).

kristina-fefelova (approved):

PR #9190, opened 2025-06-09
Add ability to enable camera in personal office

utkaka:
The camera in the personal office is turned off by default, but users now have the option to enable it in their office


plugins/love-resources/src/components/RoomTypePresenter.svelte
```diff
@@ -0,0 +1,46 @@
...
+  export let object: Room | undefined
+  export let value: RoomType | undefined
+  export let onChange: (value: any) => void
+  export let kind: ButtonKind = 'no-border'
+  export let size: ButtonSize = 'small'
+  export let justify: 'left' | 'center' = 'center'
+  export let width: string | undefined = 'fit-content'
+
+  $: booleanValue = value === RoomType.Video
+  $: disabled = object === undefined || !('person' in object) || (object as Office).person !== getCurrentEmployee()
```
haiodo, line 30: so it will be enabled for me in my office, and disabled for me in any other office? why?

utkaka: So that users cannot turn on the camera in other people's offices. Should this be allowed?

BykhovDenis: Please use `isOffice` func. It should be `$: disabled = object === undefined || (isOffice(object) && object.person !== getCurrentEmployee())`

PR #9209, opened 2025-06-10
Implement the scroll state handling and restore

rostislavnagimov:
This change saves the scroll state for scrollable elements into the Tab and restores it to the same position when tab switches.

### DIFF
--- a/plugins/workbench-resources/src/components/Workbench.svelte
+++ b/plugins/workbench-resources/src/components/Workbench.svelte
@@ -91,13 +91,6 @@
     parseLinkId,
     updateFocus
   } from '@hcengineering/view-resources'
-  import type {
-    Application,
-    NavigatorModel,
-    SpecialNavModel,
-    ViewConfiguration,
-    WorkbenchTab
-  } from '@hcengineering/workbench'
   import { getContext, onDestroy, onMount, tick } from 'svelte'
   import { subscribeMobile } from '../mobile'
   import workbench from '../plugin'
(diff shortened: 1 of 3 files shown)


plugins/workbench-resources/src/components/Workbench.svelte
```diff
@@ -254,6 +250,112 @@
...
+
+async function restoreScrollState() {
+  const tabId = $tabIdStore;
+  if (!tabId || !contentPanel) {
+    return;
+  }
+
+  console.log('[RESTORE] Waiting for two ticks for components to render...');
+  await tick();
+  await tick();
```
aonnikov: looks like hack to me. why not one and not three? can we use more reliable way here, for example events?

kristina-fefelova: oh, I used same hack in old chat, it is because of `<Component..>`. and the more they are nested, the longer we have to wait `ticks`.

need to find a better solution

rostislavnagimov: I was making that based on the existing chat components says the "one tick is not enough to load chat messages"

Till this moment all other ways was not really working for me (tested something like 5-7 other solutions)

aonnikov:
@rostislavnagimov please fix formatting issues. You can run `rush format` locally to ensure proper formatting.

rostislavnagimov:
> @rostislavnagimov please fix formatting issues. You can run `rush format` locally to ensure proper formatting.

Formating fixed 

[36585ad](https://github.com/hcengineering/platform/pull/9209/commits/36585ad55e1e8f4f400642efe0f9a2554018b26b)

PR #9221, opened 2025-06-12
Add extra handlers to room controls

rostislavnagimov:
In some cases fullscreen room was not switching the mic and camera with key combinations opening new bookmark for cmd + d instead.


plugins/love-resources/src/components/Room.svelte
```diff
@@ -383,6 +387,31 @@
   }
 
   $: activeParticipants = getActiveParticipants(participants)
+
```
aonnikov: I think that we should not handle key down here. We already have actions that toggle mic and camera. We need to figure out why these actions sometimes do not work.
```
createAction(
    builder,
    {
      action: love.actionImpl.ToggleMic,
      label: love.string.Microphone,
      icon: love.icon.Mic,
      keyBinding: ['Meta + keyD'],
      category: love.category.Office,
      allowedForEditableContent: 'always',
      input: 'none',
      target: core.class.Doc,
      context: {
        mode: ['workbench', 'browser', 'panel', 'editor', 'input']
      }
    },
    love.action.ToggleMic
  )
```

rostislavnagimov: After some research i got that the problem root was related to the missing context mode in a Room components when it renders as a fullscreen

[e79be4d](https://github.com/hcengineering/platform/pull/9221/commits/e79be4dbd386362a8384c44b2b908d589b96fadc)

rostislavnagimov: Alternatively we can allow key combinations for popup mode, not sure what is better

aonnikov (dismissed):

aonnikov (approved):

PR #9229, opened 2025-06-13
EQMS-1569: fixed ui crash when editing product members

mr1name:


plugins/contact-resources/src/components/AccountArrayEditor.svelte
```diff
@@ -24,7 +24,7 @@
   import UserBoxList from './UserBoxList.svelte'
 
   export let label: IntlString
-  export let value: AccountUuid[] | undefined
+  export let value: AccountUuid | AccountUuid[] | undefined
```
lexiv0re, line 27: Let's fix it properly after the deploy.

lexiv0re (approved):

PR #9234, opened 2025-06-14
Updated layout of the Card header

SasLord:
screen-header-1

screen-header-2

screen-header-3


kristina-fefelova:
This colored circle looks like it's not from this product, maybe we shouldn't show the tags in the header at all when we dont have space

kristina-fefelova (approved):

PR #9254, opened 2025-06-16
Qfix for admin access

haiodo:


plugins/workbench-resources/src/connect.ts
```diff
@@ -365,9 +365,9 @@ export async function connect (title: string): Promise<Client | undefined> {
   }
 
   // Ensure employee and social identifiers
-  const employee = await ensureEmployee(ctx, me, newClient, socialIds, getGlobalPerson)
+  const employee = workspaceLoginInfo.role === AccountRole.Admin ? undefined : await ensureEmployee(ctx, me, newClient, socialIds, getGlobalPerson)
 
-  if (employee == null) {
+  if (employee == null && workspaceLoginInfo.role !== AccountRole.Admin) {
```
lexiv0re: Not sure how the UI will like `undefined` current employee :) Guess we just need to test to find this out.

lexiv0re (dismissed):

plugins/workbench-resources/src/connect.ts
haiodo: Undefined is not good indeed, will think and fix.

BykhovDenis (approved):

PR #9259, opened 2025-06-16
fix: show collaborator connection errors

aonnikov:
Motivation:
When we cannot connect to collaboration service, users see blank description and no error. This PR shows error message when failed to connect.

### DIFF
--- a/plugins/text-editor-resources/src/provider/utils.ts
+++ b/plugins/text-editor-resources/src/provider/utils.ts
@@ -15,12 +15,12 @@
 
 import { type Blob, type CollaborativeDoc, type Ref, generateId } from '@hcengineering/core'
 import { encodeDocumentId } from '@hcengineering/collaborator-client'
-import { getMetadata } from '@hcengineering/platform'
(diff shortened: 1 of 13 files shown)


plugins/text-editor-resources/src/provider/utils.ts
```diff
@@ -36,28 +36,30 @@ export function createLocalProvider (ydoc: Ydoc, doc: CollaborativeDoc): Provide
...
-      parameters: { content }
-    })
+  const provider = new HocuspocusCollabProvider({
+    url: collaboratorUrl,
+    name: documentId,
+    document: ydoc,
+    token,
+    parameters: { content },
+    onConnect: () => {
+      void setPlatformStatus(OK)
```
haiodo, line 51: May be better to show a placeholder inside editor? Any other service could clear status

aonnikov: Maybe on top of the content in the editor. We can have local cache so cannot show the error instead of the content

haiodo (approved):

Issue #9265, opened 2025-06-16
Sentry envelope error prevents removing default fields of "Issue" in "Classic project" space type (Tracker)


hyoretsu:
Pretty convoluted, but if you go to Settings, Space Types, Classic Project, then Issue under "Task Types", create a new "Process States" and try to delete the default one, a Sentry request fails to `/envelope` and this shows in console

```
DeleteConfirmationPopup.svelte:43 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'length')
    at S.$$.update (DeleteConfirmationPopup.svelte:43:17)
    at new Q (31527.93c6a20….js:31:40907)
    at Sr (49245.7f36f3d….js:13:107783)
    at new Dr (49245.7f36f3d….js:13:109008)
    at Hv (bundle.bd3a6fa3ebbd4…2c0.js:10544:244993)
    at new Zv (bundle.bd3a6fa3ebbd4…2c0.js:10544:253284)
    at KC (bundle.bd3a6fa3ebbd4…2c0.js:10544:254268)
    at Object.p (bundle.bd3a6fa3ebbd4…2c0.js:10544:255543)
```

Also this:

```
fetch.js:46 
            
            
           POST https://o4506745347964928.ingest.sentry.io/api/4506745359695872/envelope/?sentry_version=7&sentry_key=19731177cb434d395dbaa60456da0b6e&sentry_client=sentry.javascript.svelte%2F9.22.0 net::ERR_BLOCKED_BY_CLIENT
```

PR #9281, opened 2025-06-17
Fix error on removing default task states

Caranell:
Fixes issue #9265. The problem originated because default task states (e.g., 'todo'/'done') had an undefined `creators` property. Attempting to access `undefined.length` subsequently threw an error, which prevented users from deleting these objects.

### DIFF
--- a/plugins/contact-resources/src/components/DeleteConfirmationPopup.svelte
+++ b/plugins/contact-resources/src/components/DeleteConfirmationPopup.svelte
@@ -75,7 +75,7 @@
       </div>
       <div class="mb-2">
         <Label label={view.string.DeletePopupCreatorLabel} />
-        {#each creators as person}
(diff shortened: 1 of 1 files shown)


aonnikov (dismissed):

plugins/contact-resources/src/components/DeleteConfirmationPopup.svelte
```diff
@@ -33,14 +33,14 @@
   const me = getCurrentEmployee()
   const objectArray = Array.isArray(object) ? object : [object]
   const dispatch = createEventDispatcher()
-  let creators: Ref<Person>[]
+  let creators: Ref<Person>[] | undefined
```
aonnikov, line 36: nit: maybe initialize with empty array and get rid of checks on undefined?

aonnikov (approved):

PR #9298, opened 2025-06-18
Tracker filters

rostislavnagimov:
The issue: filters in a tracker were set globally, so if you have two same tabs you cannot set two different set of filters (and some people in a product_feedback were asking to have this feature).

### DIFF
--- a/plugins/tracker-resources/src/components/issues/IssuesView.svelte
+++ b/plugins/tracker-resources/src/components/issues/IssuesView.svelte
@@ -8,6 +8,10 @@
   import { FilterBar, SpaceHeader, ViewletContentView, ViewletSettingButton } from '@hcengineering/view-resources'
   import tracker from '../../plugin'
   import CreateIssue from '../CreateIssue.svelte'
+  import { get } from 'svelte/store'
+  import { onMount, onDestroy, tick } from 'svelte'
+  import type { WorkbenchTab, TabUiState } from '@hcengineering/workbench'
+  import { tabIdStore, tabsStore, updateTabUiState } from '@hcengineering/workbench-resources/src/workbench'
 
   export let space: Ref<Space> | undefined = undefined
   export let query: DocumentQuery<Issue> = {}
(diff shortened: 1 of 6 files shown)


plugins/tracker-resources/src/components/issues/IssuesView.svelte
```diff
@@ -28,11 +36,129 @@
   $: if (query) updateSearchQuery(search)
   let resultQuery: DocumentQuery<Issue> = { ...searchQuery }
 
+  function saveTrackerIssueState (): void {
+    console.log('[DEBUG_STATE_ISSUES] Save State: Starting for tab', currentTabId)
```
BykhovDenis, line 39: extra logs, please remove

plugins/view-resources/src/components/SpaceHeader.svelte
```diff
@@ -5,6 +5,7 @@
   import { Viewlet } from '@hcengineering/view'
   import ViewletSelector from './ViewletSelector.svelte'
   import FilterButton from './filter/FilterButton.svelte'
+  import { WorkbenchTab } from '../../../workbench/types'
```
BykhovDenis, line 8: No. You can't use workbench types in view, because workbench depends on view... You should extract it to the view package, or use Ref<Doc> instead

BykhovDenis (changes requested):

plugins/view-resources/src/components/SpaceHeader.svelte
aonnikov: This actually indicates the problem with the approach - the header and button components know on which tab rendered, which is not correct per my point of view.

plugins/tracker-resources/src/components/issues/IssuesView.svelte
rostislavnagimov: Cleaned up

plugins/view-resources/src/components/SpaceHeader.svelte
rostislavnagimov: Moved the logic to the Workbench layer, keeping all the View components pure

aonnikov:
Can you explain the idea, how this works?

rostislavnagimov:
> Can you explain the idea, how this works?

We save the filter state in a store with a tabId relation and restore the state when user gets back on a tab. That is the basic idea.

PR #9360, opened 2025-06-25
QFix: fixed parsing of trailing parentless text and empty markers in the markdown -> markup converter

mr1name:


aonnikov (approved):
LGTM, but please add some tests

PR #9384, opened 2025-06-25
Try to fix missing communication broadcast events

kristina-fefelova:


server/server-pipeline/src/communication.ts
```diff
@@ -49,15 +49,21 @@ export class CommunicationMiddleware extends BaseMiddleware implements Middlewar
...
           }
+          if (isAsync) {
+            void context.head?.handleBroadcast(ctx)
+          }
         },
-        enqueue: (ctx, result: Event) => {
+        enqueue: (ctx, result: Event, isAsync) => {
           const { contextData, evt } = CommunicationMiddleware.wrapEvent(ctx, result)
           contextData.broadcast.queue.push(evt)
+          if (isAsync) {
```
haiodo: Why? enqueue is only for queue. It should not call it here.

server/server-pipeline/src/communication.ts
```diff
@@ -49,15 +49,21 @@ export class CommunicationMiddleware extends BaseMiddleware implements Middlewar
   static create (communicationApiFactory: CommunicationApiFactory): MiddlewareCreator {
     return async (ctx, context, next): Promise<Middleware> => {
       const communicationApi = await communicationApiFactory(ctx, context.workspace, {
-        broadcast: (ctx, sessions, result: Event) => {
+        broadcast: (ctx, sessions, result: Event, isAsync) => {
           const { contextData, evt } = CommunicationMiddleware.wrapEvent(ctx, result)
           for (const s of sessions) {
             contextData.broadcast.sessions[s] = (contextData.broadcast.sessions[s] ?? []).concat(evt)
           }
+          if (isAsync) {
```
haiodo: Why we need to call it here? It expected to be called only for user transaction.

haiodo (approved):

PR #9442, opened 2025-07-03
Fix emoji data import for missing locales

utkaka:


plugins/emoji/src/utils.ts
```diff
@@ -43,20 +43,30 @@ async function fetchEmojis (locale: Locale, options?: FetchEmojisOptions & { com
...
+        .default as ShortcodesDataset
+      shortcodes.push(shortcodeData)
+    }
 
-  return joinShortcodes(emojis, shortcodes)
+    return joinShortcodes(emojis, shortcodes)
+  } catch (e) {
+    return compact
+      ? await fetchEmojis('en', { ...options, compact: true })
+      : await fetchEmojis('en', { ...options, compact: false })
```
haiodo, line 60:
```suggestion
    return await fetchEmojis('en', { ...options, compact })
```

utkaka: It won't pass validation. There are two overloads that I made to preserve compatibility with the original library signatures:
```
async function fetchEmojis (locale: Locale, options: FetchEmojisOptions & { compact: true }): Promise<CompactEmoji[]>
async function fetchEmojis (locale: Locale, options?: FetchEmojisOptions & { compact?: false }): Promise<Emoji[]>
```

aonnikov (approved):

haiodo (approved):

PR #9471, opened 2025-07-04
Disable emoji img margin 

kristina-fefelova:
@utkaka  please check and maybe change fix

I dont understand why we need it https://github.com/hcengineering/platform/blob/develop/packages/theme/styles/common.scss#L889

because of this the emoji is not centered and the layout is crooked


utkaka:
I wanted to align custom emojis with Unicode ones, but it still doesn't work well everywhere. Now I'm planning a small rework of emojis (custom emojis in the new chat + moving to the emoji plugin from TipTap) and there I want to fix the main styles.

utkaka (approved):

PR #9490, opened 2025-07-08
Added optional REPLY_TO for email service 

f1yn:
While testing self-hosted Huly, I noted that my SMTP setups (the unidirectional ones) weren't available in a way where replies to emails would be caught.

This created situations where a password reset would be sent, or a notification - but replies to those emails would end up sucked into a SMTP hell where I'm unable to see them due to the fixed way I'm handling email domains. This PR adds an optional REPLY_TO environment variable that will apply itself when emails are being emitted from the same address as SOURCE (system emails).

I also updated the readme, since I noted that it didn't have the SOURCE env value documented.

> Rebased from #9489 - I forgot to enable signed commits on this specific machine

### DIFF
--- a/services/mail/pod-mail/src/config.ts
+++ b/services/mail/pod-mail/src/config.ts
@@ -19,6 +19,7 @@ dotenvConfig()
 export interface Config {
   port: number
   source?: string
+  replyTo?: string,
   sesConfig?: SesConfig
   smtpConfig?: SmtpConfig
 }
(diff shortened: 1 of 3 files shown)


f1yn:
Just a heads up, both commits are (now) signed off `Signed-off-by: Flynn <gpg@flyn.ca>` 👍🏻

services/mail/pod-mail/src/config.ts
```diff
@@ -19,6 +19,7 @@ dotenvConfig()
 export interface Config {
   port: number
   source?: string
+  replyTo?: string,
```
BykhovDenis:
```suggestion
  replyTo?: string
```

BykhovDenis: Fix formatting

BykhovDenis:
Hello. Thank you for your contribution. Could you please fix formatting to pass our CI?

BykhovDenis (approved):

PR #9521, opened 2025-07-10
UBERF-12227: Stabilize UI tests

lexiv0re:
Related to: https://front.hc.engineering/workbench/platform/tracker/UBERF-12227

### DIFF
--- a/plugins/chunter-resources/src/components/chat-message/ChatMessagePresenter.svelte
+++ b/plugins/chunter-resources/src/components/chat-message/ChatMessagePresenter.svelte
@@ -270,7 +270,7 @@
       {#if !isEditing}
         {#if withShowMore}
           <ShowMore limit={compact ? 80 : undefined}>
-            <div class="clear-mins">
+            <div class="clear-mins" {...!pending && { 'data-delivered': true }}>
               <MessageViewer message={displayText} />
               {#if (value.attachments ?? 0) > 0}
                 <div class="mt-2" />
@@ -279,7 +279,7 @@
             </div>
           </ShowMore>
         {:else}
-          <div class="clear-mins">
+          <div class="clear-mins" {...!pending && { 'data-delivered': true }}>
             <MessageViewer message={displayText} />
             {#if (value.attachments ?? 0) > 0}
               <div class="mt-2" />
--- a/tests/sanity/tests/model/channel-page.ts
+++ b/tests/sanity/tests/model/channel-page.ts
@@ -12,11 +12,15 @@ export class ChannelPage extends CommonPage {
 
   readonly inputMessage = (): Locator => this.page.locator('div[class~="text-editor-view"]')
   readonly buttonSendMessage = (): Locator => this.page.locator('g#Send')
-  readonly textMessage = (messageText: string): Locator =>
-    this.page.locator('.hulyComponent .activityMessage', { hasText: messageText })
+  readonly textMessage = (messageText: string, strict = false): Locator =>
+    strict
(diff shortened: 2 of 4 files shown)


tests/sanity/tests/model/channel-page.ts
```diff
@@ -99,6 +99,10 @@ export class ChannelPage extends CommonPage {
   async sendMessage (message: string): Promise<void> {
     await this.inputMessage().fill(message)
     await this.buttonSendMessage().click()
+    // since we use immediate messages delivery with async processing behind the scenes
```
haiodo: Better to use retry logic, please check. So try -> on failure wait some timeout -> try again.

lexiv0re: yes, good idea, I just wasn't sure if that was actually the problem because none of CI failures are reproduced locally, so I pushed it primarily to test the CI :)

lexiv0re: Ok so I tested 5 times and it never failed and was never flacky so probably it helped. Now the problem with the smarter retry logic is that we really know whether it failed not immediately on this step but much later. E.g. in the flaky test I was fixing it:
1. sends the message, checks it's there (it's there ofc because of the immediate delivery). Alternatively, we'd have to wait for the delivery time out and make sure the message doesn't turn gray - the only reliable condition I can think of for checking if the message has been created on the backend or not. But it would be taking much longer.
2. leaves the channel
3. joins the channel again through the channels browser
4. opens the channel
5. doesn't find the expected message here and fails

@haiodo Do you have any idea how to immediately test for the success/failure of the message creation?

aonnikov: Another reliable condition is that user receives the message. I.e you can check in another session that the message is delivered.

haiodo: @lexiv0re we have a REST api so we could use it already in UI tests :-)

BykhovDenis (approved):

plugins/chunter-resources/src/components/chat-message/ChatMessagePresenter.svelte
```diff
@@ -270,7 +270,7 @@
       {#if !isEditing}
         {#if withShowMore}
           <ShowMore limit={compact ? 80 : undefined}>
-            <div class="clear-mins">
+            <div class="clear-mins" {...!pending && { 'data-delivered': true }}>
```
aonnikov, line 273: nit: rather than adding `data-delivered` to all messages, add `data-pending` to only those that are not delivered yet.

aonnikov (approved):

plugins/chunter-resources/src/components/chat-message/ChatMessagePresenter.svelte
lexiv0re: `data-delivered` felt a little bit more intuitive and straightforward then negated `data-pending` as if it were exposed in the UI it would show the delivery sign as many popular messengers do. Do you think it produces significant overhead to swap to pending?

aonnikov: As I understand the meaning of this is different from what they have in messengers. They have 'Sent', 'Delivered', and 'Read' statuses. `data-pending` means that the message has been delivered to the server, not client.

lexiv0re: Yes, it only shows the first step in this chain now - `Sent` (delivered from the sender to the server). So yes, the name is a bit confusing :)

PR #9523, opened 2025-07-11
Communication indexer

n1kolasM:
Will require reindexing after update to consume communication messages and change mapping of elastic fields, will delete old index due to version change


haiodo:
@NikitaSkrynnik You also need to add a core migration to push a full-reindex for workspace.

haiodo (approved):

PR #9538, opened 2025-07-13
Use internal mongodb url for account in tests

n1kolasM:
For rootless docker MONGO_URL should be host-gateway but it is localhost, use internal url for consistency with other services


haiodo:
Variable is taken from .env and it is same, so I do not see why we need this change.

haiodo:
Also assume we use tests/docker-compose.override.yaml for PG, and with this change it will not be used.

n1kolasM:
DB_URL is actually overridden with `DB_URL=mongodb://localhost:27018` somehow during tests/prepare.sh. It is exported in tests/tool.sh, docker runs before tool calls, yet somehow it leaks

PR #9541, opened 2025-07-13
Analytics: add OAuth authentication and guest access events

SasLord:
Added analytics events for tracking:

- OAuth authentication start and completion via Google/GitHub
- Guest login and workspace selection
- Distinction between regular users and guests in metrics

Changes:

- Extended AnalyticProvider interface with guest parameter
- Added new events to LoginEvents
- Updated Auth.svelte and Providers.svelte components
- Enhanced analytics in guest-resources and workbench-resources

### DIFF
--- a/plugins/login-resources/src/components/Auth.svelte
+++ b/plugins/login-resources/src/components/Auth.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
-  import { Loading, getCurrentLocation } from '@hcengineering/ui'
+  import { Loading } from '@hcengineering/ui'
   import { logIn } from '@hcengineering/workbench'
-  import { trackOAuthCompletion } from '@hcengineering/analytics'
+  import { trackOAuthCompletion } from '@hcengineering/analytics-providers'
   import { onMount } from 'svelte'
 
   import {
@@ -13,21 +13,6 @@
     navigateToWorkspace
   } from '../utils'
 
-  function getProviderFromUrl (): string | null {
-    const location = getCurrentLocation()
-    const referrer = document.referrer
-
-    if (referrer.includes('accounts.google.com')) {
-      return 'google'
-    } else if (referrer.includes('github.com')) {
-      return 'github'
-    } else if (location.query?.provider != null && location.query.provider !== '') {
-      return location.query.provider
-    }
-
-    return null
-  }
-
   onMount(async () => {
     const autoJoinInfo = getAutoJoinInfo()
     if (autoJoinInfo != null) {
@@ -36,9 +21,8 @@
     }
 
     const result = await getLoginInfoFromQuery()
-    const provider = getProviderFromUrl()
 
-    trackOAuthCompletion(result, provider)
+    trackOAuthCompletion(result)
 
     if (result != null) {
       await logIn(result)
--- a/plugins/login-resources/src/components/Auth.svelte
+++ b/plugins/login-resources/src/components/Auth.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
   import { Loading, getCurrentLocation } from '@hcengineering/ui'
   import { logIn } from '@hcengineering/workbench'
+  import { trackOAuthCompletion } from '@hcengineering/analytics'
   import { onMount } from 'svelte'
 
   import {
@@ -9,8 +10,7 @@
     getAutoJoinInfo,
     goTo,
     isWorkspaceLoginInfo,
-    navigateToWorkspace,
-    trackOAuthCompletion
+    navigateToWorkspace
   } from '../utils'
 
   function getProviderFromUrl (): string | null {
--- a/plugins/login-resources/src/components/Auth.svelte
+++ b/plugins/login-resources/src/components/Auth.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { Loading } from '@hcengineering/ui'
+  import { Loading, getCurrentLocation } from '@hcengineering/ui'
   import { logIn } from '@hcengineering/workbench'
   import { onMount } from 'svelte'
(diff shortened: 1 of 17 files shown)


plugins/login-resources/src/utils.ts
```diff
@@ -1002,3 +1005,24 @@ export function getAccountDisplayName (loginInfo: LoginInfo | null | undefined):
 
   return loginInfo.account
 }
+
+export function trackOAuthCompletion (result: LoginInfo | WorkspaceLoginInfo | null, provider?: string | null): void {
```
haiodo: I think login should not have this logic, please move into analytics package, and just pass a provider.

plugins/workbench-resources/src/connect.ts
```diff
@@ -388,19 +388,30 @@ export async function connect (title: string): Promise<Client | undefined> {
     await setPlatformStatus(new Status(Severity.INFO, platform.status.SystemAccount, {}))
   }
 
-  const socialId = me.fullSocialIds.find((si) => si._id === me.primarySocialId)?.value
+  const socialId = me.fullSocialIds.find((si) => si._id === me.primarySocialId)?.key
```
haiodo, line 397: @lexiv0re Please check if this is valid change. Thanks.

haiodo (changes requested):
@lexiv0re Please check socialId stuff.

plugins/workbench-resources/src/connect.ts
```diff
@@ -388,19 +388,30 @@ export async function connect (title: string): Promise<Client | undefined> {
...
   const email = me.fullSocialIds.find((si) => si.type === 'email')?.value ?? socialId
```
lexiv0re: Please use enum for social id type

plugins/workbench-resources/src/connect.ts
lexiv0re: It only affects the analytics event data so not critical. @SasLord I'd recommend logging both `_id` and `key` of the primary social id.

plugins/workbench-resources/src/connect.ts
lexiv0re: Minor: it might be worth changing the analytics event API and not to have `email` as the mandatory special field because now the system doesn't expect everyone to have an email associated to the account.

plugins/workbench-resources/src/connect.ts
```diff
@@ -388,19 +388,30 @@ export async function connect (title: string): Promise<Client | undefined> {
...
 
   const data: Record<string, any> = {
     social_id: socialId ?? account,
```
lexiv0re: Minor: There should always be at least one social id for an account so the fallback doesn't make a lot of sense. I'd rather send another (error) event to the analytics if there's no social id for an account.

plugins/workbench-resources/src/connect.ts
lexiv0re: Also, finding primary social id is not very informative as it will always be a `Huly` id with `uuid` value which is not very human-readable. You should log account.uuid for every event anyways to use for grouping/aggregation and social id key may be used for better human readability. For that you'd need to log non-`Huly` social id.

plugins/login-resources/src/utils.ts
lexiv0re: @SasLord should it be sent from the UI at all? All these events are available in the accounts service so can be sent from the server side.

plugins/login-resources/src/components/Auth.svelte
```diff
@@ -12,6 +13,21 @@
     navigateToWorkspace
   } from '../utils'
 
+  function getProviderFromUrl (): string | null {
+    const location = getCurrentLocation()
+    const referrer = document.referrer
+
+    if (referrer.includes('accounts.google.com')) {
```
haiodo: I do not like we have it here. Let's move it into analytics.

plugins/workbench-resources/src/connect.ts
```diff
@@ -389,19 +390,34 @@ export async function connect (title: string): Promise<Client | undefined> {
     await setPlatformStatus(new Status(Severity.INFO, platform.status.SystemAccount, {}))
   }
 
-  const socialId = me.fullSocialIds.find((si) => si._id === me.primarySocialId)?.value
-  const email = me.fullSocialIds.find((si) => si.type === 'email')?.value ?? socialId
+  const hasEmail = (si: SocialId): boolean => {
```
lexiv0re, line 393: Minor: it's more like `isNonHulyId`

plugins/workbench-resources/src/connect.ts
```diff
@@ -389,19 +390,34 @@ export async function connect (title: string): Promise<Client | undefined> {
...
+    return [SocialIdType.EMAIL, SocialIdType.GOOGLE, SocialIdType.GITHUB].some((type) => type === si.type)
+  }
+  const email = me.fullSocialIds.find((si) => hasEmail(si))?.key
```
lexiv0re, line 396: Minor: `nonHulyId`

plugins/workbench-resources/src/connect.ts
```diff
@@ -389,19 +390,34 @@ export async function connect (title: string): Promise<Client | undefined> {
...
+  const socialId = me.fullSocialIds.find((si) => si._id === me.primarySocialId)?.key
 
   const data: Record<string, any> = {
-    social_id: socialId ?? account,
+    social_id: email ?? socialId ?? account,
```
lexiv0re, line 400: Minor: not sure why we need to fallback to `account` here

lexiv0re (approved):

haiodo (approved):

PR #9610, opened 2025-07-29
UBERF-11414: Integrations

ArtyomSavchenko:
- Introduces typed IntegrationKind and standardizes integration constants across services
- Adds new integration management UI with filtering, configuration, and state display capabilities
- Adds integration client to reuse some common integration logic
Screenshot 2025-08-01 at 13 06 42
Screenshot 2025-08-01 at 13 08 13


packages/integration-client/src/events.ts
aonnikov, line 1: Not: if this emitter is for local events only, then it makes sense to use some existing events library

aonnikov (approved):

packages/integration-client/src/events.ts
ArtyomSavchenko: I think I can try to use Redis pub/sub instead of it
I will add separate PR for this

BykhovDenis (approved):

PR #9698, opened 2025-08-19
Get your own integrations for admin user

ArtyomSavchenko:


server/account/src/serviceOperations.ts
```diff
@@ -694,6 +695,10 @@ export async function listIntegrations (
 
   if (isAllowedService) {
     socialIds = socialId != null ? [socialId] : undefined
+    if (socialIds === undefined && account !== systemAccountUuid) {
```
haiodo: Let's better remove admin check from verifyAllowedServices and fix all places where admin: true is potentially used.

haiodo (changes requested):
Please do not add a workarounds, we need to fix admin: 'true' to a proper service name.

server/account/src/serviceOperations.ts
```diff
@@ -586,7 +586,12 @@ export async function addSocialIdToPerson (
   const { person, type, value, confirmed, displayValue } = params
   const { extra } = decodeTokenVerbose(ctx, token)
 
-  verifyAllowedServices(['github', 'telegram-bot', 'gmail', 'tool', 'workspace', 'hulygram', 'google-calendar'], extra)
+  if (extra?.admin !== 'true') {
+    verifyAllowedServices(
+      ['github', 'telegram-bot', 'gmail', 'tool', 'workspace', 'hulygram', 'google-calendar'],
```
aonnikov, line 591: nit: I suggest to use environment variable instead of hardcoding services

aonnikov (approved):

server/account/src/serviceOperations.ts
lexiv0re: I'm not sure about env vars as we'll end up having a separate var for every method. I think we'll need to discuss how to better resolve it for all the methods at once.

PR #9700, opened 2025-08-19
Add script to restore Github integrations 

ArtyomSavchenko:


dev/tool/src/restoreGithub.ts
```diff
@@ -0,0 +1,80 @@
...
+const GITHUB_INTEGRATION: IntegrationKind = 'github' as any
+
+export async function restoreGithubIntegrations (dbUrl: string, dryrun: boolean): Promise<void> {
+  try {
+    const pg = getDBClient(dbUrl)
+    const pgClient = await pg.getClient()
+    const token = generateToken(systemAccountUuid, undefined, { service: 'admin', admin: 'true' })
+    const accountClient = getAccountClient(token)
+
+    const integrationSettings = await pgClient<
```
haiodo, line 22: It could be few of them In workspace, you need to group by account and pass an array as installationId

haiodo (changes requested):

dev/tool/src/restoreGithub.ts
ArtyomSavchenko: Thanks, fixed

haiodo (approved):

PR #9703, opened 2025-08-20
Qfix office layout

utkaka:
- Fixed layout of fullscreen room when screen sharing is completed
- The camera preview is initialized with the resolution that is actually used in meetings, eliminating the need to use "overflow:hidden", since the video stream dimensions match the dimensions of the interface elements.


plugins/media-resources/src/components/MediaPopupCamPreview.svelte
```diff
@@ -23,7 +23,10 @@
   let video: HTMLVideoElement | null = null
 
   async function updateStream (device: MediaDeviceInfo): Promise<void> {
-    const constraints = device !== null ? { video: { deviceId: { exact: device.deviceId } } } : { video: true }
+    const constraints =
+      device !== null
+        ? { video: { deviceId: { exact: device.deviceId }, width: 1280, height: 720 } }
+        : { video: { width: 1280, height: 720 } }
```
aonnikov: I recommend to use `width: { ideal: 1280 }` instead to avoid overconstrained error (in case someone's camera does not support 1280x720

haiodo (approved):

PR #9716, opened 2025-08-25
fix: limit clone depth

aonnikov:


packages/core/src/clone.ts
```diff
@@ -34,13 +34,21 @@ export function getTypeOf (obj: any): string {
...
+  as?: (doc: any, m: any) => any,
+  needAs?: (value: any) => any | undefined
+): any {
   if (typeof obj === 'undefined') {
     return undefined
   }
   if (typeof obj === 'function') {
     return obj
   }
+  if (depth <= 0) {
```
haiodo: Please support -1 as unlimited and use as default.

haiodo (changes requested):
Please support old behaviour

packages/core/src/clone.ts
aonnikov: Added `undefined` as default value

haiodo (dismissed):

haiodo (approved):

PR #9730, opened 2025-08-28
Emoji improvements

utkaka:
- Fixed custom emojis in the new chat
- Added Noto Color font. Users now able to choose between system & Noto emojis
image

image


plugins/emoji-resources/src/utils.ts
```diff
@@ -109,3 +117,43 @@ export function getEmojiSkins (emoji: ExtendedEmoji): Emoji.Emoji[] | undefined
   if (isCustomEmoji(emoji)) return undefined
   return emoji.skins
 }
+
+export function parseTextWithEmojis (text: string): ParsedTextWithEmojis {
+  const matches = [...text.matchAll(emojiGlobalRegex), ...text.matchAll(shortcodeGlobalRegex)]
+  if (matches.length === 0) return { nodes: [text], emojisOnly: false }
+  matches.sort((a, b) => a.index - b.index)
```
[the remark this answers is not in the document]
utkaka: It's not optional

packages/presentation/src/components/markup/NodeContent.svelte
```diff
@@ -73,7 +74,26 @@
       {/each}
     {/if}
   {:else if node.type === MarkupNodeType.text}
-    {node.text}
+    {#await getResource(emojiPlugin.functions.ParseTextWithEmojis) then parseFunction}
```
haiodo: I'm not happy with it, Lets make a non async -> callback function, store parsedText into a local variable, remember result of parseFunction and reuse-it, so in must case it will not be async call at all.

packages/presentation/src/components/MessageViewer.svelte
```diff
@@ -16,18 +16,31 @@
...
   export let message: string
   export let preview = false
 
   $: node = markupToJSON(message)
   $: empty = isEmptyMarkup(message)
 
+  let parseEmojisFunction: ((text: string) => ParsedTextWithEmojis) | undefined = undefined
+
+  onMount(async () => {
+    try {
```
aonnikov: this code is copied multiple times in different places. can you move it to emoji plugin?
```
onMount(async () => {
   parseEmojisFunction = await getParseEmojisFunction()
 })
```

aonnikov (dismissed):

haiodo (approved):

PR #9761, opened 2025-09-02
UBERF-13433: Migrate channels to threads

ArtyomSavchenko:
Screenshot 2025-09-02 at 15 46 36


models/chat/src/migration.ts
```diff
@@ -29,17 +31,49 @@ export const chatOperation: MigrateOperation = {
...
     },
     {
       _class: chat.masterTag.Thread
     }
   )
 }
+
+async function migrateParentInfo (client: Client): Promise<void> {
+  const txOp = new TxOperations(client, core.account.System)
+  const cards = await client.findAll(card.class.Card, { parentInfo: { $exists: true } })
```
kristina-fefelova: `client.traverse` ?

aonnikov: yeah, cannot use `findAll` for cards

models/chat/src/migration.ts
```diff
@@ -13,9 +13,67 @@
...
+    const parents = card.parentInfo.map((info) => {
+      if (info._class !== channelMasterTag) {
+        return info
+      }
+      return {
+        ...info,
+        _class: chat.masterTag.Thread
+      }
+    })
+    await txOp.update(card, { parentInfo: parents })
```
aonnikov: I think we need to use raw updates instead of txes

models/chat/src/migration.ts
ArtyomSavchenko: Thanks, fixed

aonnikov (dismissed):

kristina-fefelova (approved):

PR #9762, opened 2025-09-02
Fix avatar for system user in Home

ArtyomSavchenko:


plugins/card-resources/src/components/HomeCardPresenter.svelte
```diff
@@ -50,7 +55,9 @@
 
   $: socialId = message?.creator ?? card.modifiedBy
   $: personStore = getPersonByPersonIdStore([socialId])
-  $: person = $personStore.get(socialId)
+  $: personRef = card._class === contact.class.UserProfile ? (card as UserProfile).person : undefined
```
kristina-fefelova, line 58: Let's define `ObjectIcon` for `UserProfile` (you can check `Directs`)


So that there is no need to write special processing in each place

kristina-fefelova (approved):

aonnikov (approved):

PR #9778, opened 2025-09-04
Ladies and gentlemen, introducing hulypulse-client and Typing

lleokaganov:
hulypulse-client and its first use for the Typing service.

I had to move the hulypulse-client daemon port to 8099 because 8095 was already occupied by something in our system.


packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+// import WebSocket from "ws"
```
aonnikov: nit: here and in other places remove unused code

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+
+  static async connect(url: string | URL): Promise<HulypulseClient> {
+    const client = new HulypulseClient(url)
+    await client.connect()
+    return client
+  }
+
+  // subscribe(key, callback)
+  // return: Promise.resolve if success (true - subscribed now, false - already subscribed), Promise.reject if error
+  public async subscribe(key: string, callback: Callback): Promise<boolean> {
```
aonnikov: I think that for all operations we need to add generics support so users use proper data types instead of `any`. For example:

```
public async subscribe<T> (key: string, callback: Callback<T>): Promise<boolean> {
   ...
}
```

aonnikov: Instead of returning `true` or `false` I recommend to return callback that should be called to unsubscribe. i.e.:
```
const unsub = await client.subscribe(...)
...
unsub()
```

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+        this.subscribes[key].push(callback)
+        throw new Error(reply.error)
+      }
+    }
+    return true
+  }
+
+  // put(key, value, TTL?)
+  // return: Promise.resolve(true) if success
+  public async put(key: string, data: string, TTL?: number): Promise<boolean> {
```
aonnikov: `data` is not string, it should be an object because internally we use JSON for data. 

```
public async put<T extends object>(key: string, data: T, ...): Promise<void> {
    ...
}
```

aonnikov: We also may want to move TTL to params:
```
export interface PutParams {
    ttl?: number
}

public async put<T>(key: string, data: T, params?: PutParams): Promise<boolean> {
    ...
}
```

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+    try {
+      const reply = await this.send({ type: "put", key, data, ...(TTL !== undefined ? { TTL } : {}) })
+      if (reply.error) {
+        throw new Error(reply.error)
+      }
+      return Promise.resolve(true)
+    } catch (err) {
+      return Promise.reject(""+err)
```
aonnikov: no need to return promise, just do `throw err`

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+  }
+
+  // get(key)
+  // return: Promise.resolve(value) if success, Promise.resolve(false) if not found, Promise.reject if error
+  public async get(key: string): Promise<any> {
```
aonnikov: The result should be `Promise<T | undefined>` where undefined is used for cases when value not exists.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+      }
+      return reply.result?.data
+    } catch (err) {
+      return Promise.reject(""+err)
+    }
+  }
+
+  // get_full(key)
+  // return: Promise.resolve({data, etag, expires_at}) if success, Promise.resolve(false) if not found, Promise.reject if error
+  public async get_full(key: string): Promise<any> {
```
aonnikov: Use proper type instead of `any`.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+        throw new Error(reply.error)
+      }
+      return Promise.resolve(true)
+    } catch (err) {
+      return Promise.reject("" + err)
+    }
+  }
+
+  // send({}): Promise<any>
+  public async send(arr: any): Promise<any> {
```
aonnikov: 1. Make private unless we want to expose internal API to users
2. Use proper type instead of any in both places

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+    const id = this.correlationId++
```
aonnikov: `correlationId` will be unique within browser, if two clients come from different browsers they may have the same correlation Id. Instead use some unique value.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+    arr.correlation = id.toString()
+
+    return new Promise((resolve, reject) => {
+      if (!this.ws || this.ws.readyState !== WebSocket.OPEN) {
+        return Promise.reject(new Error("WebSocket is not open."))
+      }
+      this.pending.set(id, { resolve, reject })
+      this.ws!.send(JSON.stringify(arr))
+      setTimeout(() => {
```
aonnikov: We add a timeout for every message, this is not good. At least we need to clean them when sent.
I also don't wee where the promise is resolved, it will block send.

plugins/chunter-resources/src/components/ChannelTypingInfo.svelte
```diff
@@ -15,47 +15,57 @@
...
 
-  onMount(() => {
-    const interval = setInterval(() => {
-      void updateTypingPersons(typingInfo)
-    }, typingDelay)
-    return () => {
-      clearInterval(interval)
+  function handleTypingInfo(msg: any, key: String, index: number): void {
+    try {
+      const id: Ref<Person> = msg.key.split('/').pop() as Ref<Person>
```
aonnikov: How do you filter by channel? This code will show that someone is typing even when they type in different channel (even private).

plugins/chunter-resources/src/components/ChannelTypingInfo.svelte
```diff
@@ -15,47 +15,57 @@
...
+
+      if (msg.message === "Set") {
+        if(id === me) {
+          return
+        }
+        if (!typingInfo.includes(id)) {
+          typingInfo.push(id)
+        }
+      } else if (msg.message === "Del" || msg.message === "Unlink" || msg.message === "Expired") {
```
aonnikov: This is too complex for client. The callback should provide two values:
1. Key of what changed
2. Changed value if it is set, undefined if it deleted by any reason

plugins/chunter-resources/src/components/ChannelTypingInfo.svelte
```diff
@@ -15,47 +15,57 @@
...
+        typingInfo = typingInfo.filter((x) => x !== id)
+      }
+      void updateTypingPersons()
+    } catch (error) {
+      console.error('handleTypingInfo Error', error)
     }
+  }
+
+  onMount(() => {
+    subscribeTyping(handleTypingInfo)
```
aonnikov: Unsubscribe?

plugins/presence-resources/src/pulse.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export function getPulseClient(): HulypulseClient | undefined {
+  return pulseclient
+}
+
+export async function createPulseClient(): Promise<HulypulseClient> {
+  if(!pulseclient) {
+    const token = getMetadata(presentation.metadata.Token)
+    // pulseclient = await HulypulseClient.connect("wss://hulypulse_mem.lleo.me/ws?token=" + token)
+    pulseclient = await HulypulseClient.connect("ws://localhost:8099/ws?token=" + token)
```
aonnikov: Pulse URL should be taken from metadata similar to token and workspace

plugins/presence-resources/src/pulse.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+  return pulseclient
+}
+
+export async function setTyping(me: any): Promise<void> {
+    const workspace = getWorkspace()
+    const id = getCurrentAccount().socialIds[0];
+    await pulseclient?.put(`${workspace}/typing/${me}`, "typing", typingDelaySeconds)
```
aonnikov: I think we should pass only `/typing/...` without workspace, because it is a workspace client, workspace can be configured only once.

aonnikov (changes requested):

packages/hulypulse-client/src/client.ts
lleokaganov: We don't need unsub.

packages/hulypulse-client/src/client.ts
lleokaganov: You need to uncomment this string for run tests.

packages/hulypulse-client/src/__tests__/client.test.ts
```diff
@@ -0,0 +1,50 @@
+import { HulypulseClient } from "../client";
+
+jest.setTimeout(120000);
+
+    let client: any;
+
+    beforeAll(async () => {
+        client = await HulypulseClient.connect("ws://localhost:8095/ws");
+        // client = await HulypulseClient.connect("wss://hulypulse_mem.lleo.me/ws");
```
GeneralGDA: Do we need some personal ids (lleo) in a shared code, especially in a comment form?

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
+//
+// Copyright © 2024-2025 Hardcore Engineering Inc.
```
GeneralGDA: 2024-2025 → 2025?

packages/hulypulse-client/src/client.ts
GeneralGDA: Tests should be ran automatically on a CI.

aonnikov: Tests should work without commenting/uncommenting code.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+
+  private readonly TIMEOUT = 3000
+
+  private correlationId = 1
+  private pending = new Map<
+    number,
+    { resolve: (v: any) => void; reject: (e: any) => void }
+  >()
+
+  private constructor(private readonly url: string | URL) {}
```
GeneralGDA: Why not just an URL? string allows pushing anything here not just an url, like color or something else.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+      ws.onclose = (event) => {
+        // console.warn("WebSocket.onclose")
+        this.reconnect()
+      }
+
+      ws.onmessage = (event) => {
+        try {
+
+          if (event.data === 'ping') {
+            this.ws?.send('pong')
```
GeneralGDA, line 264: Maybe make named consts for both 'ping' and 'pong'?

plugins/presence-resources/src/pulse.ts
```diff
@@ -0,0 +1,61 @@
+// Copyright © 2023 Hardcore Engineering Inc.
```
GeneralGDA: 2023 → 2025

packages/hulypulse-client/src/client.ts
GeneralGDA: They both duplicated several times in the code.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+export class HulypulseClient implements Disposable {
+  public ws: WebSocket | null = null
+  public closed = false
+  private reconnectTimeout: any | undefined
+  private readonly RECONNECT_INTERVAL = 1000
+  private subscribes: Record<string, Callback[]> = {}
+
+  private pingTimeout: any | undefined
+  private pingInterval: any | undefined
+  private readonly PING_INTERVAL = 30 * 1000
```
GeneralGDA: I would add a suffix like MS, to denote what units are used for timeout. Like PING_INTERVAL_MS. Am I right that this is milliseconds?

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+// import WebSocket from "ws"
+
+type Callback = (data: any, key: string, index: number) => void
+
+export class HulypulseClient implements Disposable {
+  public ws: WebSocket | null = null
```
GeneralGDA: Why those ('ws' and 'closed') implementation details are public?

packages/hulypulse-client/src/__tests__/client.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        expect( await client.get("test/online/123") ).toEqual("MY");
+    },500);
+
+    test("Delete", async () => {
+        expect( await client.delete("test/online/123") ).toEqual(true);
+        expect( await client.delete("test/online/123") ).toEqual(false);
+        expect( await client.get("test/online/123") ).toEqual(false);
+    },500);
+
+    test("Subscribe", async () => {
```
GeneralGDA: May be add tests for unsubscribe? To check that we do not get any more notifications after it?

packages/hulypulse-client/src/__tests__/client.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        let r: any;
+        let cb = function(msg: any, key: string, index: number) {
+            r = { ...msg, ...{key2:key,index} };
+        }
+        expect( await client.subscribe("test/online/", cb) ).toEqual(true);
+        expect( await client.subscribe("test/online/", cb) ).toEqual(false);
+        expect( await client.put("test/online/123", "Two",1) ).toEqual(true);
+        expect(r.message).toEqual("Set");
+        expect(r.key).toEqual("test/online/123");
```
GeneralGDA: If we expect here exactly the same string as in line 38, it is better to introduce a variable for those literals. So the code explicitly communicates that we are expecting exactly the same, but not the same by accident. Same goes to other constants in other test cases.

packages/hulypulse-client/src/__tests__/client.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        expect(client).toBeInstanceOf(HulypulseClient);
+    },1000);
+
+    afterAll(() => {
+        client.close();
+    },1000);
+
+    test("Put", async () => {
+        expect( await client.put("test/online/123", "MY",200) ).toEqual(true);
+    },500);
```
GeneralGDA: I would suggest to add names for all those timeouts. Also it would be useful to explain in the comment, why some methods require a timeout of 500 ms, but other — 1000 ms. May be we should use the same timeout for all the tests?

plugins/presence-resources/src/pulse.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const typingDelaySeconds = 2
+
+let pulseclient: HulypulseClient | undefined 
+
+function getWorkspace(): string | undefined {
+  return getMetadata(presentation.metadata.WorkspaceUuid)
+}
+
+export function getPulseClient(): HulypulseClient | undefined {
```
utkaka: I think it would be better to store and initialize the client in the package itself, since we want to use it in many different places, not just in the presence plugin.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+  public closed = false
+  private reconnectTimeout: any | undefined
+  private readonly RECONNECT_INTERVAL = 1000
+  private subscribes: Record<string, Callback[]> = {}
+
+  private pingTimeout: any | undefined
```
GeneralGDA: May be it is better to use typing here: number instead of any?

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+//
+
+// import WebSocket from "ws"
+
+type Callback = (data: any, key: string, index: number) => void
+
+export class HulypulseClient implements Disposable {
+  public ws: WebSocket | null = null
+  public closed = false
+  private reconnectTimeout: any | undefined
```
GeneralGDA, line 217: Above ws defined as WebSocket | null. Here we are using undefined. Taking into account that futher you are using exact comparisons ( !== ) it is safer to use consistent typing for the absence of a value. Either use null everywhere. Or undefined.

plugins/presence-resources/src/pulse.ts
lleokaganov: А вот чот я тут не уверен. А завтра появится задача читать какие-то глобальные события/переменные из корневого workspace, и что тогда?

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+  private readonly RECONNECT_INTERVAL = 1000
+  private subscribes: Record<string, Callback[]> = {}
+
+  private pingTimeout: any | undefined
+  private pingInterval: any | undefined
+  private readonly PING_INTERVAL = 30 * 1000
+  private readonly PING_TIMEOUT = 5 * 60 * 1000
+
+  private readonly TIMEOUT = 3000
```
GeneralGDA: Above we have a PING_TIMEOUT. May be we should name this one as SEND_TIMEOUT? It is used only once as I can see. It would also make more clear what is this timeout for.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+      return Promise.reject(""+err)
+    }
+  }
+
+  // delete(key)
+  // return: Promise.resolve if success (true - deleted now, false - not found), Promise.reject if error
+  public async delete(key: string): Promise<boolean> {
+    try {
+      const reply = await this.send({ type: "delete", key })
+      if (reply.error) {
```
GeneralGDA: This if block duplicated several times. May be it should be generalized?

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+
+  private async connect(): Promise<void> {
+
+    return new Promise((resolve, reject) => {
+
+      const ws = new WebSocket(this.url.toString())
+      this.ws = ws
+
+      ws.onopen = () => {
+        // resubscribe
```
GeneralGDA: Instead of using a comment, I would suggest to move the come into a named function.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+
+
+  [Symbol.dispose](): void {
+    // console.warn("Disposing HulypulseClient")
+    this.close()
+  }
+
+  private reconnect (): void {
+    // console.warn('Reconnecting...')
+    if(this.reconnectTimeout) clearTimeout(this.reconnectTimeout)
```
GeneralGDA: I would suggest to use homogeneous style of null check at least across the file. Here universal (agains null and undefined) check is used. I also saw strict comparison usege (!==). From my point of view, there is no silver bullet: both of the strategies (fuzzy check vs strict check) have some flaws. At least consistency, as wider, as possible, reduces the bugs.

GeneralGDA: I also suggest to use {} even for one liners to prevent further maintenance bugs with forgotten {} for non-one-liners.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+  }
+
+  // delete(key)
+  // return: Promise.resolve if success (true - deleted now, false - not found), Promise.reject if error
+  public async delete(key: string): Promise<boolean> {
+    try {
+      const reply = await this.send({ type: "delete", key })
+      if (reply.error) {
+        if(reply.error=="not found") {
```
GeneralGDA: Why "not found" is so special, that we need to treat as non-exceptional value? Actually, I'm not very deep in the distributed apps, so may be it is so. @aonnikov, what do you think? 

From my point of view, current API has two ways of reporting errors: boolean return values and exceptions. So the user should handle both of them in a different style.

May be it is better to use exceptions for everything that is not OK. And for clients in need make a wrapper, translating exceptions to booleans for certain cases?

plugins/presence-resources/src/pulse.ts
```diff
@@ -0,0 +1,61 @@
...
+    const workspace = getWorkspace()
+    const id = getCurrentAccount().socialIds[0];
+    await pulseclient?.put(`${workspace}/typing/${me}`, "typing", typingDelaySeconds)
+}
+
+export function subscribeTyping(callback: any): void {
+    const workspace = getWorkspace()
+    pulseclient?.subscribe(`${workspace}/typing/`, callback)
+}
+
```
utkaka, line 64: @aonnikov I understand that chat presence and typing status are related, but is it worth to use the presence plugin for typing status? Can we use Pulse directly?

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+    } catch (err) {
+      return Promise.reject(""+err)
+    }
+  }
+
+  // get_full(key)
+  // return: Promise.resolve({data, etag, expires_at}) if success, Promise.resolve(false) if not found, Promise.reject if error
+  public async get_full(key: string): Promise<any> {
+    try {
+      const reply = await this.send({ type: "get", key })
```
GeneralGDA: I would suggest to introduce the named consts for all the protocol massages. It would reduce duplicates. Also all the protocol could be made focused in one place, rather than being spread across the whole file - esier to maintain and read (the protocol conventions) in the future.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+    if (!this.subscribes[key] || !this.subscribes[key].includes(callback)) {
+      return false
+    }
+    this.subscribes[key] = this.subscribes[key].filter(cb => cb !== callback)
+    if (this.subscribes[key].length === 0) {
+      delete this.subscribes[key]
+      const reply = await this.send({ type: "unsub", key })
+      if (reply.error) {
+        this.subscribes[key] = this.subscribes[key] || []
+        this.subscribes[key].push(callback)
```
GeneralGDA: This rollback strategy is flawed in the next case. Consider a callback with a huge closure. Like an anonymous method somewhre in UI using a lot of stuff: canvas, images, etc. Now this client gets unloaded (like Svelte component, for example) and fails to unsubscribe. What should it do? Try to unsub in a loop, holding all the resources?

I can also imagine clients (people) that do not expect such transactional behavior from unsubscribe routine, so for me one-shot unsubscribes (as a following resource leaks) would be common in future. May be we should automatically schedule unsubscribe here? We already doing auto reconnect, for instance.

plugins/presence-resources/src/pulse.ts
aonnikov: @utkaka I think that we need to do some refactoring here and move code related to pulse to `pulse` and `pulse-resoruces` plugins. Presence and typing stuff can leave in different plugins.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+  public async subscribe(key: string, callback: Callback): Promise<boolean> {
+    if (!this.subscribes[key]) {
+      this.subscribes[key] = []
+    }
+
+    if (this.subscribes[key].includes(callback)) {
+      return false // Already subscribed
+    }
+
+    this.subscribes[key].push(callback)
```
GeneralGDA: May be it is better to add after successful send? In such way we do not need the following rollback code.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+	        if (event.data === 'pong') {
+      	    clearTimeout(this.pingTimeout)
+      	    return
+	        }
+
+          const msg = JSON.parse(event.data.toString())
+          // Handle incoming messages (Set, Expired, Del)
+          if (msg.message) {
+            for (const key in this.subscribes) {
+              if (msg.key.startsWith(key)) {
```
GeneralGDA, line 275: Why 'startsWith'? Is it a feature that the keys are kind of hierarchical? Like "foobar" will also receive messages for "foo"? I think it should be documented, and checked in the tests: non-trivial behavior.

plugins/presence-resources/src/pulse.ts
aonnikov: Не появится. У нас сейчас всё живёт в каком-то воркспейсе и мы должны убедиться, что пользователь приходит с правильным токеном в нужный воркспейс.
Необходимость передавать воркспейс в каждом запросе выглядит утомительной. Хочется упрощать жизнь пользователю, а не наоборот.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,311 @@
...
+  public async send(arr: any): Promise<any> {
+    const id = this.correlationId++
+    arr.correlation = id.toString()
+
+    return new Promise((resolve, reject) => {
+      if (!this.ws || this.ws.readyState !== WebSocket.OPEN) {
+        return Promise.reject(new Error("WebSocket is not open."))
+      }
+      this.pending.set(id, { resolve, reject })
+      this.ws!.send(JSON.stringify(arr))
```
GeneralGDA: Force suppress compiler null checks ( this.ws! ) is a bad practice. I believe we do not need this "!" here: you've already checked the ws for null higher, so compiler is good to figure out that here it is not null.

packages/hulypulse-client/src/client.ts
aonnikov: timeouts are not always numbers ... As far as I remember it either NodeJS.Timeout or number, so it is a general (bot not good) practice in our code to use `any | undefined` for them.

packages/hulypulse-client/src/client.ts
aonnikov: `url | URL` is more convenient than just `url` or `URL` but we have to do some validation that the passed value is valid URL.

But in this place exact place I would pass token additionally, the user don't have to know this specific of how to pass token into the service.

packages/hulypulse-client/src/client.ts
aonnikov: The `key` might have already been removed from the map (line 215).

Agree, we don't have to push this back into the list of subscriptions, but instead we may need to retry attempt.  Not necessary to do this synchronously, we can do this in background.

packages/hulypulse-client/src/client.ts
GeneralGDA: Why "more convenient"? It opens a possibility to pass some string that is not URL by mistake. Why not just use "any" as it is even more convinient?

packages/hulypulse-client/src/client.ts
aonnikov: Not only for messages, the protocol must be typed. I.e:

```
interface GetMessage {
    type: 'get'
    key: string
}

interface PutMessage {
    type: 'put'
    key: string
    ...
}

type ProtocolMessage = GetMessage | PutMessage | ...
```

packages/hulypulse-client/src/client.ts
GeneralGDA: This is a private field, not a public access point. Using "any" here opens possibility of assigning here (by accident/mistake) any value. For example, currently, the only non-null value assigned like this: this.pingTimeout = setTimeout
I can replace setTimeout with any function name, taking a lambda, and the code compiles.

packages/hulypulse-client/src/client.ts
aonnikov: Yes, we need to use exceptions only, not need to use additional flag.

`not found` should not be an error at all. As I understand, it indicates that value for given key is not set. It is perfectly normal case, not an exception.

plugins/presence-resources/src/pulse.ts
aonnikov: Not in the package please, the package provides just the client. For UI client we need a single copy of the client and it is the problem of UI only how to handle this.

plugins/presence-resources/src/pulse.ts
lleokaganov: "I think we should pass only /typing/... without workspace, because it is a workspace client, workspace can be configured only once."

I’ve thought a lot about this, but no, I don’t understand. So where would we store the workspace then? It’s not involved in the connection initialization, only the token is. Parsing the token seems kind of silly and complicated. Passing the currently unnecessary workspace as a separate variable just to keep it on the client? That also feels strange. It’s unclear what the point of that would be.

plugins/presence-resources/src/pulse.ts
lleokaganov: It seemed to me that this is the most suitable place — the launch definitely happens here when entering a new workspace. If there’s another place for it, let’s move it there.

packages/hulypulse-client/src/__tests__/client.test.ts
lleokaganov: Alright, I removed it if it bothers you. It was left over from debugging. If I have to debug other functions, I’ll need to switch to the external server and recall its address again. It’s more convenient for me to monitor and read/write streams there, whereas here, inside the Docker server, it’s not very clear how to view/change/read variables in a convenient format.

packages/hulypulse-client/src/client.ts
lleokaganov: Awesome idea!!! Thanks a ton — I’ll definitely do that! 🚀

packages/hulypulse-client/src/client.ts
lleokaganov: Passing a string that isn’t a URL is a minor issue. Much more dangerous is if the string is a URL — and that URL belongs to an attacker! That would be a disaster. But since none of the users are providing URLs, and it’s just a constant hardcoded in the code, I believe we don’t need to check for either case. What do you think?

packages/hulypulse-client/src/client.ts
lleokaganov: This is described in detail in the README: https://github.com/hcengineering/hulypulse

In short: the subscription key must always end with '/', otherwise the server will return an error.
Thus, when subscribing to workspace/foo/, we will receive workspace/foo/bar, workspace/foo/bar/man, and so on. All of them will start with workspace/foo/

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: You are right. But I assumed that the subscription should start working immediately at the moment of subscribing, so as not to miss the next event.
If you think this can be neglected, I’ll move the procedure to the end.

packages/hulypulse-client/src/client.ts
lleokaganov: The array in question essentially serves as a duplicate of the subscription on the server.
If removing the subscription on the server fails, we cannot pretend it was successfully removed and delete it from the array.
Otherwise, we may run into collisions where the server continues processing the subscription, while on the client side it is considered removed and no one is listening to it anymore.

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: A server response error: not found in our case is not really an error, since it simply means that the requested variable does not exist at the moment (it may appear a second later). In this case, we return undefined rather than an error. If you have other ideas, I’d be glad to hear them. However, I wouldn’t want to retroactively rewrite the Rust server already running in production just because of such minor details.

packages/hulypulse-client/src/client.ts
lleokaganov: tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: Different operations trigger different types of errors, and they are handled in different ways depending on the task. I’m afraid the code will become hard to read if we start moving every couple of lines into a separate procedure.

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: How would you suggest changing the line?
  public ws: WebSocket | null = null

packages/hulypulse-client/src/client.ts
lleokaganov: How would you suggest changing the line?

packages/hulypulse-client/src/__tests__/client.test.ts
lleokaganov: At the moment we don’t have any tests left — they all broke after the refactoring and the issues with WebSocket.

packages/hulypulse-client/src/__tests__/client.test.ts
lleokaganov: At the moment we don’t have any tests left — they all broke after the refactoring and the issues with WebSocket.

packages/hulypulse-client/src/__tests__/client.test.ts
lleokaganov: At the moment we don’t have any tests left — they all broke after the refactoring and the issues with WebSocket.

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

plugins/presence-resources/src/pulse.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

plugins/presence-resources/src/pulse.ts
lleokaganov: Please help me set this up. For now, I’ve left a TODO in the code, but I couldn’t figure out how to create a variable for the daemon on my own.

plugins/chunter-resources/src/components/ChannelTypingInfo.svelte
lleokaganov: When the client is destroyed, the subscription will disappear automatically.

plugins/chunter-resources/src/components/ChannelTypingInfo.svelte
lleokaganov: I’m afraid this is too expensive — intercepting every stream event and analyzing or modifying it before the callback.
The callback might not even be interested in that at all.
Maybe it’s only configured to catch the deletion of a long-lived variable — how could we know that in advance?

lleokaganov: Thank you, that’s wise. I did exactly that.

plugins/chunter-resources/src/components/ChannelTypingInfo.svelte
lleokaganov: This function is designed to be associated only with the subscription "{workspace}/typing/"

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: Each browser creates its own connection. The correlation parameter exists only within a single WebSocket and will not propagate to other WebSockets.

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: There’s still a problem with the tests.

packages/hulypulse-client/src/client.ts
GeneralGDA: Yes, that would work.

packages/hulypulse-client/src/client.ts
GeneralGDA: I'm not a security guy. But in the perspective of code maintenance any issue (minor or not) with a zero fix cost (removal of the string |) should be fixed.

packages/hulypulse-client/src/client.ts
GeneralGDA: We are not suggesting rewriting the server. We are suggesting rewriting the client. On the client side the case of not found could be an exceptional situation. For example a misprint: subscribe("boo") → unsubscribe("goo"). I do not really catch why this case is not an exception?

packages/hulypulse-client/src/client.ts
GeneralGDA: `  private pingTimeout: ReturnType<typeof setTimeout> | undefined `

packages/hulypulse-client/src/__tests__/client.test.ts
GeneralGDA: You are not the only person who will work with the code. You are suggesting that every maintainer should leave a personal URL convenient for them here? And we should watch for people leaving the team and keep the list up to date?

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,376 @@
...
+  | SubscribeMessages
+  | UnsubscribeMessages
+  | SubscribesList
+  | InfoMessage;
+
+export class HulypulseClient implements Disposable {
+  private ws: WebSocket | null = null
+  private closed = false
+  private reconnectTimeout: any | undefined
+  private readonly RECONNECT_INTERVAL = 1000
```
GeneralGDA: May be an _MS suffix here too?

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,376 @@
...
+    this.pingInterval = setInterval(() => {
+      if (this.ws !== null && this.ws.readyState === WebSocket.OPEN) {
+        this.ws.send(WEBSOCKET_MESSAGE_CONSTANT_PING_REQUEST)
+      }
+      clearTimeout(this.pingTimeout)
+      this.pingTimeout = setTimeout(() => {
+        if (this.ws !== null) {
+          console.log('no response from server')
+          clearInterval(this.pingInterval)
+          this.ws.close(1000)
```
GeneralGDA: May be introduce a name for the 1000 ms? Like we did for other timeouts in other places?

packages/hulypulse-client/src/client.ts
GeneralGDA: We are suggesting to change failure handling strategy, not to get rid of it. Current design forces UI-based clients to keep memory-heavy resources in memory for the undefined (possibly infinite) period of time.

packages/hulypulse-client/src/client.ts
GeneralGDA: Lines 321-326, 332-337, 348-353 are exact copies of each other without any differences. In case of changes one would change 3 places with a probability to mess up. get and get_full by their schematics are 95% chance should have exact exception handling strategy. delete is a wrapper for the 'send' too which implies the same exception handling scheme.

packages/hulypulse-client/src/client.ts
aonnikov: This client is not for catching misprints and typos 😄 If I called `unsubscribe` method, I would expect:
1. If there was a subscription, it is stopped and I do not receive updates anymore
2. The method is idempotent and reenterable, i.e. if there were no subscription, then I do not receive any error. This is important for simplifying interaction with the client. If someone needs to know whether there was a subscription, we can add a boolean status indicating status.

You can compare this with `Map` - when you delete something that does not exist in the map you don't get an exception. If you try to get something that not exists in the map, you don't get an exception either. It would be extremely difficult to use maps otherwise, every call would be wrapped into try-catch with complex handling.

packages/hulypulse-client/src/client.ts
aonnikov: I'm ok with having `string | URL`, it is completely valid for `WebSocket` class and they have all necessary verification that the passed value is a valid URL. From the user perspective of view, this allows to write less code, i.e. instead of `new Client(new Url(url))` I can just do `new Client(url)`.

packages/hulypulse-client/src/client.ts
GeneralGDA: This is a web client. The presence of network and the actual code in it (exceptions) already forces users to wrap every method in try-catch.

My point is to keep only one exceptional situations handling strategy: either exceptions, or "error codes" (in any form: booleans, maybe monads, etc, does not matter), but not both. In other case users will be forced (if they want a correct code, which by definition must handle all the exceptions) to check both booleans and exceptions.

In your example, buy the way, the Map implementation uses only one strategy.

aonnikov: Ah, yes, I agree. We need to get rid of error codes, just use exceptions. Ideally client methods should not raise any error at all, may be not exactly this client, but at some level we don't want to know about errors, instead I would expect it handle errors internally - retry / ignore / reconnect / etc.

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,376 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+type Callback = (key: string, data: JSONValue, info: { msg: JSONObject, subscribed_key: string, run_index: number }) => void
```
aonnikov: type Callback<T> = (key: string, data: T, info: { msg: JSONObject, subscribed_key: string, run_index: number }) => void

Use `T` or `T extends object` depending on what we need.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,376 @@
...
+
+const WEBSOCKET_MESSAGE_CONSTANT_PING_REQUEST = "ping";
+const WEBSOCKET_MESSAGE_CONSTANT_PONG_RESPONSE = "pong";
+
+type JSONPrimitive = string | number | boolean | null;
```
aonnikov: type Value = object | string | number | boolean

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,376 @@
...
+        }
+      }
+
+    })
+  }
+
+  private async resubscribe(reject: (reason?: any) => void): Promise<void> {
+    for (const key in this.subscribes) {
+      this.send({ type: "sub", key }).catch((error) => {
+        reject(new Error(`Resubscription failed for key=${key}: ${error.message ?? error}`));
```
aonnikov: just throw Error

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed:

type JSONValue = string | number | boolean | null | JSONValue[] | { [key: string]: JSONValue }
interface JSONObject { [key: string]: JSONValue }

packages/hulypulse-client/src/client.ts
lleokaganov: Так не получилось. Пишет unresolved T.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,383 @@
...
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+const WS_CLOSE_NORMAL = 1000
+
+type Callback = (
+  key: string,
+  data: JSONValue,
```
aonnikov: This one should be properly typed, otherwise it will be unsafe to use the client

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,383 @@
...
+}
+
+interface DeleteMessage {
+  type: 'delete'
+  key: string
+  ifMatch?: string
+  correlation?: string | number
+}
+
+interface SubscribeMessages {
```
aonnikov, line 183: nit: I think there is a typo in the interface name, it should be `SubscribeMessage` instead of `SubscribeMessages` because it represents protocol message for `sub` operation

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,383 @@
...
+            this.ws?.send(WEBSOCKET_MESSAGE_CONSTANT_PONG_RESPONSE)
+            return
+          }
+
+          if (event.data === WEBSOCKET_MESSAGE_CONSTANT_PONG_RESPONSE) {
+            clearTimeout(this.pingTimeout)
+            return
+          }
+
+          const msg = JSON.parse(event.data.toString())
```
aonnikov: It is not clear what the response can be, it is not typed here. Try to do:
```
const msg = JSON.parse(event.data.toString()) as ProtocolMessage
```

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,383 @@
...
+                  } catch (err) {
+                    console.error(`Error in callback #${index} with key "${key}":`, err)
+                  }
+                })
+              }
+            }
+          }
+          // Handle correlation responses
+          else if (msg.correlation) {
+            let id = Number(msg.correlation)
```
aonnikov: This block of code should be rewritten in more safe manner.
1. `correlation` is `string | number` but here you make assumption that it is a number
2. it is unsafe to do `get(id)!`

```
const pending = this.pending.get(msg.correlation)
if (pending !== undefined) {
    clearTimeout(pending.send_timeout)
    pending.resolve(msg)
    this.pending.delete(msg.correlation)
}
```

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,383 @@
...
+            console.warn('Unknown message format:', msg)
+          }
+        } catch (e) {
+          console.error('Failed to parse message', e)
+        }
+      }
+    })
+  }
+
+  private async resubscribe(reject: (reason?: any) => void): Promise<void> {
```
aonnikov: reject is unused?

1. The function seem to be not asynchronous, but is marked as `async`
2. It uses asynchronous functions, but does not properly wait for them (or mark usages with `void`)

This will cause a problem with unhandled rejections:
1. `resubscribe` calls `send` and immediately exits
2. `send` throws an error which is catched, but new error is thrown
3. this error won't be handled

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,383 @@
...
+    await client.connect()
+    return client
+  }
+
+  public async info(): Promise<string> {
+    const reply = await this.send({ type: 'info' })
+    if (reply.error) {
+      throw new Error(reply.error)
+    }
+    return reply?.result
```
aonnikov: `reply?.result` is `string | undefined`, the `info` cannot return string in this case

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,383 @@
...
+  | UnsubscribeMessages
+  | SubscribesList
+  | InfoMessage
+
+export class HulypulseClient implements Disposable {
+  private ws: WebSocket | null = null
+  private closed = false
+  private reconnectTimeout: any | undefined
+  private readonly RECONNECT_INTERVAL_MS = 1000
+  private subscribes: Record<string, Callback[]> = {}
```
aonnikov: I think it should be `Map`, not `Record` because Map is more suitable for this. Record is just an object.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,383 @@
...
+    const id = this.correlationId++
+    const message = { ...msg, correlation: id.toString() } as M
+
+    return new Promise((resolve, reject) => {
+      if (!this.ws || this.ws.readyState !== WebSocket.OPEN) {
+        return reject(new Error('WebSocket is not open.'))
+      }
+      const send_timeout = setTimeout(() => {
+        if (this.pending.has(id)) {
+          this.pending.get(id)!.reject(new Error('Timeout waiting for response'))
```
aonnikov: The same as above, rewrite without type assertion

packages/hulypulse-client/src/client.ts
lleokaganov: Here it’s not about a message. These are API methods. And in the naming I based it on the expected result of calling the method: get/put/delete performs an operation with a single message, while sub/unsub results in subscribing/unsubscribing to multiple messages. Of course, if you believe renaming is necessary - no problem, I’ll rename it. But aren’t we just wasting time on trivialities when we’re digging out and fixing such "errors" in the code?

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: According to the API protocol, correlation is of type string — for the general case. It’s possible that some other clients will use a hash or assign the same names to requests of the same type. But in our specific client implementation, we use a number. That’s why we can’t get rid of Number(msg.correlation), even if we move it into a separate function. How do you suggest we proceed?

packages/hulypulse-client/src/client.ts
lleokaganov: What do you mean here? Right now, we only have one protocol — the API protocol. It includes options that I specifically described and named putMessage, subscribeMessages, but there is no protocolMessage. And if we place our helper functions alongside the actual protocol options, it will cause confusion. Or what exactly did you mean?

packages/hulypulse-client/src/client.ts
lleokaganov: What is meant here? Are you referring to JSONValue? But we already defined it below — you even helped phrase it correctly yourself:
type JSONValue = string | number | boolean | null | JSONValue[] | { [key: string]: JSONValue }

packages/hulypulse-client/src/client.ts
lleokaganov: Thanks for pointing that out. That was my mistake, I’ve fixed it.

packages/hulypulse-client/src/client.ts
lleokaganov: Thanks for the advice. I’ve reworked it.

packages/hulypulse-client/src/client.ts
aonnikov: No, this defines some abstract object that can be used as data. Normally, people want to use more strict types.
For example, for typing I may want to use something like (pseudocode): 
```
interface TypingInfo {
  who: Ref<Person>
  where: Ref<Channel>
}

await client.subscribe('/typing', (data: TypingInfo) => {
    ...
})
```

In this example, data passed to the callback will have proper types that allows to write safer code. If it is not typed, but declared as `any`, then there will be no type check in compile time and it becomes much easier to make a mistake.

packages/hulypulse-client/src/client.ts
lleokaganov: I didn’t quite understand what you meant here. What type exactly? How should it be rewritten?

packages/hulypulse-client/src/client.ts
aonnikov: If it is string in the protocol, you can use `string` instead of `string | number` for `correlationId` and get rid of number conversion (key in `pending` map will also be string).

packages/hulypulse-client/src/client.ts
lleokaganov: Tnx, fixed as:

return reply.result || ''

packages/hulypulse-client/src/client.ts
aonnikov: The problem is that `JSON.parse` returns `any`, and that does not tell anything about te content of message returned by pulse server.
What I mean, is that there should be a proper type for server response. For example these lines below assume that that `msg` contains `message` field.

```
         // Handle incoming messages (Set, Expired, Del)
          if (msg.message) {
```

You can make it more explicit by adding proper interface for response:

```
interface Response {
    message: 'Set' | 'Expired' | 'Del'
}
```

This will allow you to type check your response and avoid mistakes.

packages/hulypulse-client/src/client.ts
lleokaganov: Thanks, now I understand. I’ve changed everything to string. Although I don’t think it’s a good idea to waste extra memory on a string where a conversion to number could be used instead.

packages/hulypulse-client/src/client.ts
aonnikov: This one is very bad, it tells compiler to believe you that `get(id)` returns something, not undefined.
```
this.pending.get(id)!
```

Instead do this:
```
const pending = this.pending.get(id)
if (pending !== undefined) {
    ...
}
```

packages/hulypulse-client/src/client.ts
lleokaganov: Oh, no. The protocol is too complex to lay out endless trees of all its variations.

aonnikov: Then you need to simplify the protocol.There should not be endless number of response variations.

packages/hulypulse-client/src/client.ts
lleokaganov: OK, fixed

packages/hulypulse-client/src/client.ts
lleokaganov: I’m losing the point and don’t understand what I’m supposed to fix now. I get the impression that this story of minor tweaks to already working code will be endless. I’ll need some time to accept this thought and come to terms with it. And is all the other platform code reviewed just as thoroughly, or is it just an issue with me?

aonnikov: type safety is NOT minor tweak!

packages/hulypulse-client/src/client.ts
lleokaganov: The thing is, a large variety of data flows through the WebSocket. These can be responses to operations, responses to informational requests, and of course the actual messages that were subscribed to. These are different formats. The protocol started being developed before me, then the server was implemented, debugged, and approved. If our task is to completely redesign the format of all messages passing through the WebSocket to bring them to a unified form, then first of all, I don’t yet understand how to do that since they are all different. Secondly, I would need authorization to fully rework the Rust server, debug it again, and allocate time for this work.

packages/hulypulse-client/src/client.ts
lleokaganov: If this is a security question, I don’t yet understand the nature of the threat, possible attack scenarios, or an example of a vulnerability. And I honestly don’t understand what exactly I’m supposed to do here in this part of the code, or how.

packages/hulypulse-client/src/client.ts
aonnikov: The point is not to rework rust service, but properly document possible responses from the server. The best way to "document" them is to define types. Otherwise, how do you (and other maintainers of this code) know that the client handles response properly?
Even if responses are of different format you can easily do this:

```
interface ResponseTypeA { ... }
interface ResponseTypeB { ... }
interface ResponseTypeC { ... }

type Response = ResponseTypeA | ResponseTypeB | ResponseTypeC
```

plugins/chunter-resources/src/components/chat-message/ChatMessageInput.svelte
```diff
@@ -107,21 +104,19 @@
...
   }
 
   async function updateTypingInfo (): Promise<void> {
     if (!withTypingInfo) return
 
     throttle.call(() => {
-      const room = { objectId: object._id, objectClass: object._class }
-      const typing = { person: me, lastTyping: Date.now() }
-      updateMyPresence(room, { typing })
+      setTyping(me)
```
aonnikov: This is not correct, `objectId` is missing, this will show that you are typing in every channel.

packages/hulypulse-client/src/client.ts
aonnikov: > what exactly I’m supposed to do here in this part of the code, or how.

Try to use the client, not only "key" but data as well. In this PR you have only one example that does not use data.

plugins/chunter-resources/src/components/chat-message/ChatMessageInput.svelte
lleokaganov: Это очень правильное замечание.
Я поменял setTyping и learTyping, передаю (me, String(object._id))
Поменял соответствующие функции.
Но возникла проблема с подпиской. Подписка инициируется из:
platform/plugins/chunter-resources/src/components/ChannelTypingInfo.svelte 
Подписка инициируется так:
  onMount(() => {
    subscribeTyping(handleTypingInfo, String(object._id))
  })
  Но в этом разделе нет object, и откуда он появляется и где его взять, я не знаю. Тут мне потребуется помощь авторов кода.

packages/hulypulse-client/src/client.ts
lleokaganov: Кажется я начинаю понимать, о чем речь. Это данные ПОЛЬЗОВАТЕЛЯ - информация, которую сохранил в базе данных кто-то. Разумеется, мы заранее не знаем, в каком они формате, API позволяет сохранить в базе произвольный string. Соответственно, нам не известно, кто и как в будущем станет использовать клиента и что сохранять в базе. Поэтому здесь должен быть string. Но меня просили исправить на JSON, что не вполне корректно, я исправил. Но какой именно и с какими полями это будет JSON на все случаи будущего использования клиента - об этом клиент не знает и знать не должен.

lleokaganov: Сервис typing - лишь частный случай использования, он использует просто ключ в качестве флажка, и тут мы поле data в базу вообще не отправляем (записываем пустое или true, не помню сейчас). Но вообще в поле data сервер ожидает любой произвольный string, хоть json, хоть jpeg uuencode.

aonnikov: Для этого и предлагается использовать Generic тип T, чтобы пользователь САМ указал, какого типа у него данные, а клиент в этом месте поверил ему и вернул ему это тип.
Если ты вернёшь в колбэке строку, или даже какой-то JSON, то пользоваться этим будет неудобно - каждый колбэк обрастёт дополнительной валидацией, приведением типа и т.д

aonnikov: Сервис typing не так прост, как может показаться. Нужно "выставить флажок" что пользователь печатает в определенном канале (например в General), для этого в пульс нужно передать идентификатор пользователя и идентификатор канала. Использовать для этого ключ - неудобно, мы не можем из строки в удобном виде извлечь информацию. А вот пользовательские данные использовать будет более удобно, если мы их правильно типизируем.

packages/hulypulse-client/src/client.ts
```diff
@@ -0,0 +1,392 @@
...
+
+  public async list(): Promise<string> {
+    const reply = await this.send({ type: 'list' })
+    if (reply.error) {
+      throw new Error(reply.error)
+    }
+    return reply.result || ''
+  }
+
+  public async subscribe<T>(key: string, callback: Callback): Promise<() => Promise<boolean>> {
```
aonnikov: callback: Callback<T>

aonnikov (approved):

utkaka (approved):

GeneralGDA (approved):

PR #9781, opened 2025-09-04
Migrate card spaces to exclude channels

ArtyomSavchenko:


models/chat/src/migration.ts
```diff
@@ -139,3 +151,29 @@ export async function performParentInfoMigration (client: MigrationClient, bulkS
     await iterator.close()
   }
 }
+
+async function migrateCardSpaces (client: MigrationClient): Promise<void> {
+  const cardSpaces = await client.find<CardSpace>(DOMAIN_SPACE, {
```
haiodo, line 156: Please use iteration API.

haiodo (changes requested):
Please fix migration

models/chat/src/migration.ts
ArtyomSavchenko: I agree that iteration is definitely worth using for large collections, but is it really necessary for card spaces?
There are 3 card spaces in the Platform workspace and I can't imagine a situation where there will be too many of them
I have no objections to using an iterator, I just want to clarify
Screenshot 2025-09-04 at 18 01 57

haiodo: ok

haiodo (approved):

PR #9799, opened 2025-09-08
update cockroach and redpanda versions

aonnikov:


lexiv0re:
Are they in par with the production versions? I think we should be testing production versions or if we upgrade them here should upgrade the prod as well.

aonnikov:
The plan is to upgrade here, test, then migrate on staging/prod :) We cannot migrate without testing.

haiodo:
If we not test a version and will do upgrade, will it be better?

lexiv0re:
Of course we should test first :) Just wanted to make sure we're planning to upgrade the prod after testing :)

aonnikov:
We are definitely planning because for RedPanda maintenance ends soon, and for CockroachDB it already ended on 2025-02-12.

haiodo (approved):

PR #9809, opened 2025-09-09
UBERF-13564: Fix logout caused by connectivity issues

haiodo:


plugins/login-resources/src/utils.ts
```diff
@@ -394,30 +394,30 @@ export async function getRegionInfo (doNavigate: boolean = true): Promise<Region
...
+    return [OK, loginInfo, true]
   } catch (err: any) {
     if (err instanceof PlatformError && err.status.code === platform.status.Unauthorized) {
       const loc = getCurrentLocation()
       loc.path[0] = 'login'
       loc.path[1] = 'login'
       loc.path.length = 2
       navigate(loc)
-      return [unknownStatus('Please login'), null]
+      return [unknownStatus('Please login'), null, true]
```
aonnikov, line 411: nit: instead of unknownStatus we can just return `err.status`.

plugins/login-resources/src/utils.ts
```diff
@@ -394,30 +394,30 @@ export async function getRegionInfo (doNavigate: boolean = true): Promise<Region
...
     } else if (err instanceof PlatformError) {
       Analytics.handleEvent(LoginEvents.SelectWorkspace, { name: workspaceUrl, ok: false })
       await handleStatusError('Select workspace error', err.status)
 
-      return [err.status, null]
+      return [err.status, null, false]
```
aonnikov: should it be `true` instead? If we received a PlatformError, this means we managed to perform request to account and do not need to retry.

aonnikov (dismissed):

plugins/login-resources/src/utils.ts
haiodo: Let it be

plugins/login-resources/src/utils.ts
haiodo: Fixed.

lexiv0re (approved):

PR #9873, opened 2025-09-16
EQMS-1633 Fix qms inline comments

aonnikov:


BykhovDenis (dismissed):

plugins/text-editor-resources/src/components/extension/qms/qmsInlineCommentMark.ts
```diff
@@ -202,39 +164,78 @@ export const QMSInlineCommentMark = Mark.create<QMSInlineCommentMarkOptions, QMS
...
+            descendant.marks.forEach((mark) => {
+              const uuid = getMarkUuid(mark)
+              if (uuid !== undefined) {
+                pastedUuids.add(uuid)
+              }
+            })
+          })
+        })
+
+        let hasDuplicatedUuids = false
```
lexiv0re, line 217: So if we copy from a different doc the marks will stay? What about the comments referencing them?

plugins/controlled-documents-resources/src/components/document/EditDocContent.svelte
```diff
@@ -241,19 +240,8 @@
...
-                    documentCommentsHighlightUpdated(selectedNodeId !== null ? { nodeId: selectedNodeId } : null)
-                  }
-                },
-                onNodeClicked: (uuid) => {
-                  if (selectedNodeId !== uuid) {
-                    selectedNodeId = uuid
-                  }
-
+                onNodeClicked: (uuids) => {
+                  selectedNodeId = Array.isArray(uuids) ? uuids[0] : uuids
```
lexiv0re: I'm not sure this logic is good. E.g. if the first comment is already resolved and not displayed in the right panel at all.

lexiv0re: As a user I'd expect to skip resolved comments (unless the checkmark "Show resolved" is checked) and loop over the associated array of comments when clicking on the same node.

aonnikov: Fixed both issues:
1. Ignore resolved when the option is disabled
2. Scroll through comments on click

plugins/text-editor-resources/src/components/extension/qms/qmsInlineCommentMark.ts
aonnikov: As discussed on the call, yes, we cannot remove them because of lack of information in this exact place.

lexiv0re (approved):

PR #9989, opened 2025-10-01
use of postgress-base

haiodo:
@BykhovDenis @aonnikov  Please take an extensive look at storage.ts changes.


aonnikov (dismissed):

server/postgres/src/storage.ts
```diff
@@ -389,6 +389,9 @@ abstract class PostgresAdapterBase implements DbAdapter {
           if (Object.keys(remainingData).length > 0) {
             updates.push(`data = ${params.add(converted.data, '::json')}`)
           }
+
+          updates.push(`"%hash%" = ${vars.add(this.curHash())}`)
```
BykhovDenis: We have it at 373

BykhovDenis (dismissed):

server/postgres/src/storage.ts
haiodo: Thanks. Removed.

aonnikov (approved):

PR #9991, opened 2025-10-01
Do not add collabs on mentions whrn security enabled

kristina-fefelova:


BykhovDenis:
I'm not sure that this is good idea...

haiodo (approved):

PR #10005, opened 2025-10-02
Display card parent

ArtyomSavchenko:
Screenshot 2025-10-02 at 18 04 11


BykhovDenis (approved):

haiodo:
will it be visible also on right panel?

ArtyomSavchenko:
> will it be visible also on right panel?

For now it is visible only if panel width is large enough
Looks like we should display parent or back button in compact way in this case

PR #10015, opened 2025-10-03
Qfix: billing livekit errors

utkaka:


services/love/src/billing.ts
```diff
@@ -104,7 +104,7 @@ export async function updateLiveKitSessions (ctx: MeasureContext): Promise<void>
         sessionId: session.sessionId,
         sessionStart: session.createdAt,
         sessionEnd: session.endedAt ?? session.lastActive,
-        bandwidth: Math.max(Number(session.bandwidthOut), Number(session.bandwidthIn)),
+        bandwidth: Math.max(Number(session.bandwidthOut ?? '0'), Number(session.bandwidthIn ?? '0')),
```
aonnikov, line 107: I wonder if this is a correct condition? What to we pay for - in, out, or both?

utkaka: If you request statistics for a specific session, one field is returned:
`bandwidth, // billable bytes of bandwidth used`
But when you request a list of sessions, these two fields are returned for each. And empirically, I've found that the maximum of the two fields is returned in a session request. But the documentation doesn't mention this, and it might be worth further consideration and verification.

aonnikov (approved):

PR #10025, opened 2025-10-05
qfix: do not allow to disable contacts plugin

aonnikov:


BykhovDenis (dismissed):

haiodo:
@aonnikov It was exactly this way, I think we need a mode visible and simplify logic.

BykhovDenis (approved):

PR #10033, opened 2025-10-06
qfix: log window reload reason

aonnikov:
log window reload reason to make it easier understand page refresh reason

### DIFF
--- a/dev/prod/src/platform.ts
+++ b/dev/prod/src/platform.ts
@@ -387,6 +387,7 @@ export async function configurePlatform() {
         if (err.message.includes('Loading chunk') && i != 4) {
           continue
         }
+        console.log('reload due to loading error')
         location.reload()
       }
     }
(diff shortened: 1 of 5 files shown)


lexiv0re:
@aonnikov are these logs seen in the terminal console?

lexiv0re (approved):

aonnikov:
These logs are seen in browser console and in terminal console

dev/prod/src/platform.ts
```diff
@@ -387,6 +387,7 @@ export async function configurePlatform() {
         if (err.message.includes('Loading chunk') && i != 4) {
           continue
         }
+        console.log('reload due to loading error')
```
haiodo, line 390: May be we should send it to analytics?

haiodo (approved):

dev/prod/src/platform.ts
aonnikov: Yeah, will add in another PR

PR #10034, opened 2025-10-06
Show emojis in mentions

kristina-fefelova:
Screenshot 2025-10-06 at 12 35 55


aonnikov (dismissed):

packages/presentation/src/components/markup/ObjectNode.svelte
```diff
@@ -35,12 +36,14 @@
   let broken: boolean = false
   const withoutDoc: Ref<Doc>[] = [contact.mention.Here, contact.mention.Everyone]
 
+  $: clazz = _class ? hierarchy.getClass(_class) : undefined
```
haiodo: Will it throw if class not found?

kristina-fefelova: yes..looks like we need to use `findClass`

haiodo (approved):

PR #10037, opened 2025-10-06
Validate and normalize txes received from user

n1kolasM:
Correctness mostly relies on type checking, if types of Txes change, typescript will require fixes to validation, new Tx subtypes will be rejected


server/middleware/src/normalizeTx.ts
```diff
@@ -0,0 +1,259 @@
...
+      attachedTo: attachedTo as Ref<Doc>,
+      attachedToClass: attachedToClass as Ref<Class<Doc>>,
+      collection
+    })
+    if (baseCUD._class === core.class.TxCreateDoc) {
+      const { attributes } = source as Record<keyof TxCreateDoc<Doc>, unknown>
+      if (typeof attributes !== 'object' || attributes === null) {
+        return undefined
+      }
+      const createDoc: ExplicitTx<TxCreateDoc<Doc>> = Object.assign(baseCUD, { attributes })
```
haiodo, line 229: Please look at TxProcessor more carefully, it has methods to create doc from TxCreateDoc etc.

n1kolasM: I don't understand how it would help here; There's no document creation in this middleware, only validation and reassembling of transactions

aonnikov: In communication we use [zod](https://zod.dev) for schema validation, maybe we can use it here as well?

n1kolasM: I thought about zod, but I am not sure it is possible to require zod schema to be in sync with Tx types as I did for this validation

haiodo (approved):

PR #10048, opened 2025-10-08
qfix: validate workspace uuid in datalake

aonnikov:


lexiv0re (dismissed):

lexiv0re:
Can we also change the type in the DB's blob.workspace to uuid?

aonnikov:
I think yes, but migration won’t be trivial

lexiv0re:
Okay, then maybe keep it in mind for future.

haiodo (approved):

PR #10063, opened 2025-10-10
Prevent duplicating ToDos in documents content

n1kolasM:


plugins/text-editor-resources/src/components/extension/todo/todo.ts
```diff
@@ -77,7 +85,10 @@ export const TodoItemExtension = TaskItem.extend<TodoItemExtensionOptions>({
     return {
       ...this.parent?.(),
       todoid: getDataAttribute('todoid', { default: null, keepOnSplit: false }),
-      userid: getDataAttribute('userid', { default: null, keepOnSplit: false })
+      userid: getDataAttribute('userid', { default: null, keepOnSplit: false }),
+      parentid: getDataAttribute('parentid'),
```
aonnikov: Do you think we need to persist three more attributes? We will have to ensure that we keep them when doing sync with github at least. Do you know if there is a way to make an attribute that does not serialize to prosemirror?

n1kolasM: I've checked serialized document and these additional attributes did not appear in serialized form, and it looks like alternatives require recreating a lot of parsing/serializing logic

aonnikov (dismissed):

plugins/text-editor-resources/src/components/extension/todo/todo.ts
aonnikov: Let's try this approach then

aonnikov (approved):

PR #10074, opened 2025-10-13
Chore: use fixed versions for hcengineering packages

lexiv0re:
* uses fixed versions for hcengineering packages
* also, improves sass deprecation silencing
* adds commented local deps to huly.server and huly.utils


lexiv0re:
Closed in favor of https://github.com/hcengineering/platform/pull/10078 as decided to not use fixed versions but follow semver more carefully.

PR #10076, opened 2025-10-13
Fix admin integrations

ArtyomSavchenko:


ws-tests/api-tests/src/__tests__/rest.test.ts
```diff
@@ -75,7 +75,7 @@ describe('rest-api-server', () => {
     accountClient = getAccountClient(config.ACCOUNTS_URL, apiWorkspace1.token)
     adminAccountClient = getAccountClient(
       config.ACCOUNTS_URL,
-      generateToken(systemAccountUuid, undefined, { admin: 'true' }, 'secret')
+      generateToken(systemAccountUuid, undefined, { service: 'admin', admin: 'true' }, 'secret')
```
lexiv0re: `service: 'admin'` doesn't look good...

ArtyomSavchenko: I replaced it with `workspace` service

server/account/src/operations.ts
```diff
@@ -2068,7 +2068,7 @@ export async function ensurePerson (
 ): Promise<{ uuid: PersonUuid, socialId: PersonId }> {
   const { account, workspace, extra } = decodeTokenVerbose(ctx, token)
   const allowedService = verifyAllowedServices(
-    ['tool', 'workspace', 'schedule', 'mail', 'github', 'hulygram'],
+    ['tool', 'workspace', 'schedule', 'mail', 'github', 'hulygram', 'admin'],
```
lexiv0re: Remove `admin` from here and other places as well?

ArtyomSavchenko: I added this for backward compatibility, since it was used in some migration scripts
However, now I think it's better to just update the scripts to use `tool` service

ArtyomSavchenko: Done

lexiv0re (approved):

PR #10083, opened 2025-10-14
Use navigation by types instead of spaces

ArtyomSavchenko:
Screenshot 2025-10-14 at 13 24 40

### DIFF
--- a/models/card/src/index.ts
+++ b/models/card/src/index.ts
@@ -447,15 +447,24 @@ export function createModel (builder: Builder): void {
             position: 'top'
           }
         ],
-        spaces: [
+        spaces: [],
+        groups: [
           {
-            id: 'spaces',
-            label: core.string.Spaces,
-            spaceClass: card.class.CardSpace,
-            addSpaceLabel: core.string.Space,
-            icon: card.icon.Space,
-            // intentionally left empty in order to make space presenter working
-            specials: []
+            id: 'types',
+            label: card.string.MasterTags,
+            groupByClass: card.class.MasterTag,
+            icon: card.icon.MasterTags,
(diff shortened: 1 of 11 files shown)


kristina-fefelova:
Do we need  top-level section `types` ?

ArtyomSavchenko:
> Do we need top-level section `types` ?

Do you mean add a section for types next to `All Cards` and `Spaces`?
I think this would be useful
In my opinion, it would also be good to add the ability to add a new type from there without going into the settings

kristina-fefelova:
> > Do we need top-level section `types` ?
> 
> Do you mean add a section for types next to `All Cards` and `Spaces`? I think this would be useful In my opinion, it would also be good to add the ability to add a new type from there without going into the settings

I mean, what's the point of this `types` section? Does it do anything other than create additional nesting?

Screenshot 2025-10-14 at 14 54 38

IMHO: there's no need to create types here, it's not a common operation.

ArtyomSavchenko:
> > > Do we need top-level section `types` ?
> > 
> > 
> > Do you mean add a section for types next to `All Cards` and `Spaces`? I think this would be useful In my opinion, it would also be good to add the ability to add a new type from there without going into the settings
> 
> I mean, what's the point of this `types` section? Does it do anything other than create additional nesting?
> 
> Screenshot 2025-10-14 at 14 54 38
> IMHO: there's no need to create types here, it's not a common operation.

I think the types look worse without the section.
We could consider making the section non-collapsible or displaying the heading separately, but I don't know if it's worth the time.
Screenshot 2025-10-14 at 18 30 34

haiodo (approved):

PR #10114, opened 2025-10-20
Subscriptions

lexiv0re:
* Adds subscription management


services/payment/pod-payment/src/providers/polar/provider.ts
```diff
@@ -0,0 +1,112 @@
...
+    this.subscriptionPlans = {}
+    const plans = subscriptionPlans.split(';')
+    for (const plan of plans) {
+      const [type, rawProductIds] = plan.split(':')
+      const productIds = rawProductIds.split(',')
+      this.subscriptionPlans[type] = productIds
+    }
+    // TODO: verify all plans are present in the config - take them from model?
+    // hardcoded check for now
+    const mustHave = ['common@tier', 'rare@tier', 'epic@tier', 'legendary@tier']
```
aonnikov, line 84: If we take this from model, should it be something like `billing:tier:Common`?

lexiv0re: yes, we can, although I think we should still have `type` and `plan` fields in the descriptor object explicitly even if we declare them this way

services/payment/pod-payment/src/server.ts
```diff
@@ -0,0 +1,236 @@
...
+  } catch (err: unknown) {
+    ctx.error('Failed to process payment request', { err })
+    res.status(500).end()
+  }
+}
+
+export async function createServer (ctx: MeasureContext, config: Config): Promise<{ app: Express, close: () => void }> {
+  const app = express()
+  app.use(cors())
+  app.use(express.json({ limit: '50mb' }))
```
aonnikov: nit: limit makes sense only for services that accept huge payload, can be removed here

lexiv0re: i thought it also helps to prevent dealing with unwanted oversized payloads which someone with bad intentions might be sending? If not I can remove, if it helps I'd reduce the limit to 5mb or something like this?

aonnikov: As far as I remember there is a default limit that is small enough. 50mb is used to increase the limit.

haiodo (approved):

PR #10169, opened 2025-10-29
Update changelog and fix api client reference

ArtyomSavchenko:


haiodo:
did you use staging-new to create a changelog?

ArtyomSavchenko:
> did you use staging-new to create a changelog?

Initially no, now I updated it to version generated based on staging-new

haiodo (approved):

PR #10288, opened 2025-12-10
Check permission in create card popup

BykhovDenis:


plugins/card-resources/src/components/CreateCardPopup.svelte
```diff
@@ -140,6 +141,8 @@
       _space = event.detail.space
     }
   }
+
+  $: allowed = _space && canCreateObject(type, _space, $permissionsStore)
```
ArtyomSavchenko, line 145: Should permissions be checked in `card-resources/src/components/NewCardForm.svelte` as well?

BykhovDenis: Thank you. Will add in another PR

PR #10336, opened 2025-12-29
Card versioning

BykhovDenis:


models/card/src/migration.ts
```diff
@@ -412,3 +417,20 @@ async function migrateRolePermissions (client: Client): Promise<void> {
     await txOp.update(role, { permissions: [] })
   }
 }
+
+async function fillVersioning (client: MigrationClient): Promise<void> {
+  const cards = await client.find<Card>(DOMAIN_CARD, { baseId: { $exists: false } })
```
ArtyomSavchenko: I assume `MigrationClient.find` return all entries at once that can cause problems during migration.
Should we use bulk processing here?

BykhovDenis: I'm not sure that we have a lot of cards for any real workspace((( 
Thank you, fixed

ArtyomSavchenko: Thanks
At least many cards should be on the front, since there are still a large number of cards created by mail and telegram integrations previously.

PR #10367, opened 2026-01-05
Multiple user request input

BykhovDenis:


plugins/process-resources/src/utils.ts
```diff
@@ -449,40 +450,48 @@ export async function requestUserInput (
...
-        await promise
-      }
+            resolve()
+          }
+        )
+      })
+      await promise
     }
   }
+  console.log('changed', userContext)
```
ArtyomSavchenko: [Minor] Should we clean up this logging?

ArtyomSavchenko (approved):

plugins/process-resources/src/utils.ts
BykhovDenis: Thank you

PR #10411, opened 2026-01-16
Support documents table diff and refresh

ArtyomSavchenko:
Screenshot 2026-01-16 at 17 25 21
Screenshot 2026-01-16 at 17 31 22


aonnikov:
LGTM

PR #10468, opened 2026-02-01
feat: Ability to select multiple docs and to export to pdf

ArtyomSavchenko:
Screenshot 2026-02-01 at 22 44 01

### DIFF
--- a/plugins/print-resources/src/components/PrintBulkToPDF.svelte
+++ b/plugins/print-resources/src/components/PrintBulkToPDF.svelte
@@ -19,6 +19,7 @@
   import presentation, { Card, getClient, PDFViewer } from '@hcengineering/presentation'
   import view from '@hcengineering/view'
   import { Button, Label, Loading } from '@hcengineering/ui'
+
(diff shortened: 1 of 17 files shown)


plugins/print-resources/src/components/PrintBulkToPDF.svelte
```diff
@@ -0,0 +1,190 @@
...
+              </div>
+            </div>
+            <div class="flex justify-end">
+              <Button kind="secondary" label={presentation.string.Cancel} on:click={cancel} />
+            </div>
+          </div>
+        {:else}
+          <div class="flex-1 min-h-0 max-h-64 overflow-auto">
+            <div class="flex flex-col gap-1 w-full">
+              {#each results as result}
```
aonnikov: Do we need to wrap into scrollable container?

ArtyomSavchenko: Thanks, fixed

aonnikov (approved):

PR #10507, opened 2026-02-14
feat: Redesign workspace join

ArtyomSavchenko:
Suggest to join using current account if user is alredy logged in:
Screenshot 2026-02-16 at 15 06 38

Rename "Sign Up" and "Join" buttons to avoid ambiguity:
Screenshot 2026-02-16 at 15 06 48
Screenshot 2026-02-16 at 15 07 01


aonnikov:
The `Would you like ... %USERNAME%` sentence is too long, can we redesign the page to be:

```
%USERNAME%      -- this can be displayed on every login page in the header
You are invited to join %WORKSPACENAME%

<Join %WORKSPACENAME%>
<Change account>
```

Also on Join / Signup and Join pages it would be nice to have workspace name.

aonnikov (approved):

PR #10512, opened 2026-02-15
Support minified dev deployment

ArtyomSavchenko:
Full dev deployment: 33 services, ~4 GB RAM
Minified dev deployment: 17 services, ~2 GB RAM


dev/docker-compose-min.yaml
aonnikov, line 1: Is that possible to do the same with merged compose files?
https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/

Maybe we can set this up as list of files:

```
docker-compose.yaml         -- full setup
docker-compose.min.yaml  -- modifies the full setup
```

or
```
docker-compose.base.yaml -- base setup
docker-compose.love.yaml -- adds love to the base
docker-compose.stream.yaml -- adds stream to the base
...
```

aonnikov (approved):

PR #10519, opened 2026-02-17
Fix: password signup shows JSON parse error when MAIL_URL is configured

### LINKED ISSUE #10518: Bug: Password signup shows "Unknown error: Unexpected token" when email service is configured
### Describe the bug

When signing up with password (using the "Set Password Now" toggle on the signup page), the UI shows:

> Unknown error: Unexpected token 'N', "Not Found" is not valid JSON

...or:

> Unknown error: Unexpected token '<', "\<html\> \<h"... is not valid JSON

The account IS created successfully, but the user is stuck on the signup page with the error. They must navigate to Sign In and use OTP to log in.

### Root cause

The `signUp` endpoint in `server/account/src/operations.ts` intentionally returns `token: undefined` when `MAIL_URL` is configured (production deployments with email confirmation):

```typescript
return {
  account,
  name: getPersonName(person),
  socialId,
  token: !forceConfirmation ? generateToken(account) : undefined  // undefined in production
}
```

However, `SignupForm.svelte` calls `logIn(result)` unconditionally without checking if `result.token` is defined:

```typescript
// plugins/login-resources/src/components/SignupForm.svelte
const [loginStatus, result] = await signUp(object.username, object.password, object.first, object.last)
status = loginStatus
if (result != null) {
  await logIn(result)           // BUG: calls logIn even when result.token is undefined
  goTo('confirmationSend')
}
```

This triggers `setCookie()` which sends `PUT /cookie` without an `Authorization` header. The account service returns 401 with an empty/invalid body, and the client crashes trying to parse it as JSON.

**The OTP signup flow handles this correctly.** `doLoginNavigate()` in `plugins/login-resources/src/utils.ts` has the proper guard:

```typescript
if (result.token != null) {    // Correctly checks for token
  await logIn(result)
}
loc.path[1] = result.token != null ? 'selectWorkspace' : 'confirmationSend'
```

### Regression

Introduced in PR #8057 ("feat: cookie token", 2025-02-24). Before that PR, the password signup path used `setMetadata()` which tolerated `undefined` tokens. The PR changed it to `await logIn(result)` → `setCookie()` which requires a valid token.

### Secondary issue

The `PUT /cookie` endpoint in `server/account-service/src/index.ts` uses raw Node.js response handling when returning 401:

```typescript
ctx.body = JSON.stringify({ error: new Status(...) })
ctx.res.writeHead(401)
ctx.res.end()  // Bypasses Koa — ctx.body may not be sent
```

This can produce an empty response body or cause "upstream prematurely closed connection" errors in reverse proxies. The client then fails at `response.json()` because the body isn't valid JSON.

### Steps to reproduce

1. Deploy Huly with `MAIL_URL` configured (any SMTP/SES setup)
2. Go to the signup page
3. Toggle "Set Password Now" (or use the password signup form)
4. Enter first name, last name, email, and password
5. Click "Sign Up"
6. Error appears: "Unknown error: Unexpected token..."

### Expected behavior

After password signup with email confirmation required:
1. Account should be created (it is)
2. User should be redirected to "confirmation sent" page (not shown — error appears instead)
3. User checks email for confirmation/OTP and completes login

### Suggested fix

Add the `token != null` guard in `SignupForm.svelte`, matching the OTP flow:

```typescript
if (result != null) {
  if (result.token != null) {
    await logIn(result)
  }
  goTo('confirmationSend')
}
```

And fix the `PUT /cookie` 401 response to use Koa's response mechanism instead of raw `ctx.res`:

```typescript
ctx.status = 401
ctx.body = { error: new Status(Severity.ERROR, platform.status.Unauthorized, {}) }
```

### Environment

- Huly version: v0.7.353 (`hardcoreeng/*` Docker images)
- Deployment: Self-hosted via Docker Compose with CockroachDB
- Reverse proxy: nginx (Dokploy)
- SMTP: Configured and working (OTP emails deliver successfully)

### Workaround

The account IS created despite the error. Users can:
1. Dismiss the error
2. Go to **Sign In**
3. Enter email/password
4. Complete OTP verification via email
5. Login succeeds normally

spatialy:
## Summary

Fixes #10518

When signing up with a password on deployments where `MAIL_URL` is configured, the account service returns `token: undefined` to enforce email confirmation. `SignupForm.svelte` was calling `logIn()` unconditionally — this triggered `PUT /cookie` with no `Authorization` header, the server returned 401 with an unparseable body, and the client crashed with _"Unknown error: Unexpected token"_, leaving users stuck on the signup page even though their account was created successfully.

- **`SignupForm.svelte`**: guard `logIn()` with `result.token != null`, matching the pattern already used in `doLoginNavigate()` in `utils.ts`
- **`server/account-service/src/index.ts`**: fix `PUT /cookie` 401 response to use Koa's `ctx.status`/`ctx.body` instead of raw `ctx.res.writeHead()`/`ctx.res.end()`, so the JSON error body is correctly serialized
- **`signupTokenGuard.test.ts`**: unit tests covering the token guard logic for all cases (undefined token, absent token, present token, null result)

## Test plan

- [ ] Deploy with `MAIL_URL` set to a real or dummy SMTP URL
- [ ] Sign up with password ("Set Password Now") — should redirect to confirmation-sent page, no error
- [ ] Verify `PUT /cookie` is not called in the network tab (token is undefined, so `logIn()` is skipped)
- [ ] Deploy without `MAIL_URL` — sign up with password should still call `logIn()` and proceed to workspace selection as before
- [ ] `curl -X PUT http://localhost:3000/cookie` → should return valid JSON `{"error":{...}}` with HTTP 401

### DIFF
--- a/plugins/login-resources/src/__tests__/signupTokenGuard.test.ts
+++ b/plugins/login-resources/src/__tests__/signupTokenGuard.test.ts
@@ -1,0 +1,97 @@
+// Copyright © 2025 Hardcore Engineering Inc.
+//
+// Licensed under the Eclipse Public License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License. You may
+// obtain a copy of the License at https://www.eclipse.org/legal/epl-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Tests for the token guard fix in SignupForm.svelte (issue #10518).
+ *
+ * When MAIL_URL is configured the account service returns token: undefined
+ * to force email confirmation. The signup handler must skip logIn() in that
+ * case — calling logIn() without a token triggers PUT /cookie with no
+ * Authorization header, which returns 401 and crashes the client's JSON
+ * parser with "Unexpected token".
+ */
+
+interface LoginInfo {
+  account: string
+  name?: string
+  token?: string
+}
+
+/**
+ * Mirrors the fixed logic from SignupForm.svelte:
+ *
+ *   if (result != null) {
+ *     if (result.token != null) {
+ *       await logIn(result)
+ *     }
+ *     goTo('confirmationSend')
+ *   }
+ */
+async function handleSignupResult (
+  result: LoginInfo | null,
+  logIn: (info: LoginInfo) => Promise<void>,
+  goTo: (page: string) => void
+): Promise<void> {
+  if (result != null) {
+    if (result.token != null) {
+      await logIn(result)
+    }
+    goTo('confirmationSend')
+  }
+}
+
+describe('SignupForm token guard (issue #10518)', () => {
+  let logIn: jest.Mock
+  let goTo: jest.Mock
+
+  beforeEach(() => {
+    logIn = jest.fn().mockResolvedValue(undefined)
+    goTo = jest.fn()
+  })
+
+  it('skips logIn and redirects to confirmationSend when token is undefined (MAIL_URL configured)', async () => {
+    // Server returns token: undefined when email confirmation is required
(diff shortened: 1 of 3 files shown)


ArtyomSavchenko (approved):

ArtyomSavchenko:
@spatialy Thanks for your contribution

PR #10546, opened 2026-02-24
Time service

BykhovDenis:
@ArtyomSavchenko FYI redpanda version changed for local docker-compose, and could require clean up


aonnikov:
@BykhovDenis why you updated redpanda? We are not planning to update it on prod in near future so it is better to keep the same version for dev as well.

services/worker/src/worker.ts
```diff
@@ -13,24 +13,57 @@
...
+    try {
+      const expiredEvents = await db.getExpiredEvents()
+      if (expiredEvents.length > 0) {
+        for (const event of expiredEvents) {
+          await SendTimeEvent(event.workspace, event.topic, event.data)
+        }
+        await db.deleteEvents(expiredEvents)
+      }
+    } catch (err) {
+      console.error('Error in Time Machine polling loop:', err)
```
aonnikov: better to use `MeasureContext`

BykhovDenis: fixed

haiodo (approved):

aonnikov (approved):

tests/docker-compose.yaml
```diff
@@ -11,6 +11,21 @@ services:
     ports:
       - 1081:1080
     restart: unless-stopped
+  time-machine:
+    image: hardcoreeng/worker
```
aonnikov, line 15: nit: `worker` is too generic, suggest to rename to `time-worker`
you also need to add it to `common/scripts/docker.sh`

PR #10554, opened 2026-02-25
fix(account): handle empty $in array to prevent PostgreSQL IN () syntax error

### LINKED ISSUE #10553: Bug: getWorkspacesInfo generates invalid SQL IN () when called with empty array (PostgreSQL)
## Bug Description

The `getWorkspacesInfo` account method generates invalid SQL when called with an empty workspace ID array. PostgreSQL rejects `WHERE "uuid" IN ()` as a syntax error. This causes the GitHub integration (and potentially Gmail/Backup) to fail silently every ~6 minutes, and blocks connecting new GitHub repos.

## Environment

- **Version**: v0.7.375
- **Database**: PostgreSQL 17 (self-hosted)
- **Affected service**: `hardcoreeng/account`, called by `hardcoreeng/github`

## Error

```
PostgresError: syntax error at or near ")"
  position: "74"
  code: "42601"
  method: "getWorkspacesInfo"
  source: "github"
```

This repeats every ~6 minutes indefinitely. Connecting new GitHub repos fails with `platform:status:BadRequest` in the transactor.

## Root Cause

**`server/account/src/collections/postgres/postgres.ts` ~line 177** — `buildWhereClause` does not handle empty `$in` arrays:

```typescript
case '$in': {
  const inVals = Object.values(qKey as object)[0]
  const inVars: string[] = []
  for (const val of inVals) {      // empty array → no iterations
    currIdx++
    inVars.push(formatVar(currIdx, castType))
    values.push(val)
  }
  whereChunks.push(`"${snakeKey}" IN (${inVars.join(', ')})`)  // → IN ()
  break
}
```

When `inVals` is `[]`, this produces `IN ()` which is invalid SQL in PostgreSQL (and CockroachDB).

**`server/account/src/operations.ts` ~line 1617** — `getWorkspacesInfo` checks `ids == null` but not `ids.length === 0`, so empty arrays pass through to the query.

**`services/github/pod-github/src/platform.ts` ~line 1040** — the GitHub service calls `getWorkspacesInfo(workspaces)` without guarding for empty arrays. When there are no integrations, `workspaces` is `[]`.

## Impact

| Caller | Guards empty array? |
|--------|-------------------|
| `pod-github/src/platform.ts` ~1040 | No |
| `pod-gmail/src/gmailController.ts` ~196 | No |
| `server/backup/src/service.ts` ~283 | No |
| `pod-calendar/src/calendarController.ts` ~75 | **Yes** — `if (ids.length === 0) return` |

The calendar service is the only caller that properly guards against this.

## Suggested Fix

Either (or both):

**Option A** — Fix at the query level in `getWorkspacesInfoWithStatusByIds` (`server/account/src/utils.ts` ~line 1350):
```typescript
if (uuids.length === 0) return []
```

**Option B** — Fix at the SQL builder level in `buildWhereClause` (`postgres.ts`):
```typescript
case '$in': {
  const inVals = Object.values(qKey as object)[0]
  if (inVals.length === 0) {
    whereChunks.push('FALSE')
    break
  }
  // ... existing code
}
```

Option B is more defensive and prevents the same bug from other callers.

## Reproduction

1. Self-host Huly v0.7.375 on PostgreSQL (not CockroachDB)
2. Configure the GitHub integration (`GITHUB_APPID`, etc.)
3. Install the GitHub App on an organization
4. Observe `account` service logs — `getWorkspacesInfo` errors every ~6 minutes
5. Try to connect a repo to a project — fails with BadRequest

spatialy:
## Summary

- In `buildWhereClause` (`postgres.ts`), emit `FALSE` when a `$in` array is empty instead of generating invalid `IN ()` SQL that PostgreSQL rejects with a syntax error.
- In `getWorkspacesInfoWithStatusByIds` (`utils.ts`), add an early `return []` when `uuids` is empty so callers (GitHub, Gmail, Backup services) never reach the query with an empty list.

Fixes #10553

## Test plan

- [ ] Self-host on PostgreSQL with the GitHub integration configured but no repos connected — confirm the `IN ()` error no longer appears in account logs every ~6 minutes.
- [ ] Verify `getWorkspacesInfo` returns `[]` when called with an empty id array instead of throwing a PostgreSQL syntax error.
- [ ] Verify existing workspace-lookup behaviour is unchanged when a non-empty id array is provided.

### DIFF
--- a/server/account/src/utils.ts
+++ b/server/account/src/utils.ts
@@ -1351,6 +1351,7 @@ export async function getWorkspacesInfoWithStatusByIds (
   db: AccountDB,
   uuids: WorkspaceUuid[]
 ): Promise<WorkspaceInfoWithStatus[]> {
+  if (uuids.length === 0) return []
   const statuses = await db.workspaceStatus.find({ workspaceUuid: { $in: uuids } })
   const statusesMap = statuses.reduce<Record<string, WorkspaceStatus>>((sm, s) => {
     sm[s.workspaceUuid] = s
--- a/server/account/src/utils.ts
+++ b/server/account/src/utils.ts
@@ -1351,7 +1351,7 @@ export async function getWorkspacesInfoWithStatusByIds (
   db: AccountDB,
   uuids: WorkspaceUuid[]
 ): Promise<WorkspaceInfoWithStatus[]> {
-  if (uuids.length === 0) return []
+  if (!Array.isArray(uuids) || uuids.length === 0) return []
   const statuses = await db.workspaceStatus.find({ workspaceUuid: { $in: uuids } })
   const statusesMap = statuses.reduce<Record<string, WorkspaceStatus>>((sm, s) => {
     sm[s.workspaceUuid] = s
--- a/server/account/src/collections/postgres/postgres.ts
+++ b/server/account/src/collections/postgres/postgres.ts
@@ -176,6 +176,10 @@ implements DbCollection<T> {
       switch (operator) {
         case '$in': {
           const inVals = Object.values(qKey as object)[0]
+          if (inVals.length === 0) {
+            whereChunks.push('FALSE')
+            break
+          }
           const inVars: string[] = []
           for (const val of inVals) {
             currIdx++


spatialy:
@ArtyomSavchenko, this is preventing GitHub sync from working properly in the latest version deployed with Postgres. Can you check?

server/account/src/utils.ts
```diff
@@ -1351,6 +1351,7 @@ export async function getWorkspacesInfoWithStatusByIds (
   db: AccountDB,
   uuids: WorkspaceUuid[]
 ): Promise<WorkspaceInfoWithStatus[]> {
+  if (uuids.length === 0) return []
```
ArtyomSavchenko: Hi @spatialy 
I think it would be good to also handle the case when `uuids` is not an array and return `[]` as well

aonnikov (approved):

ArtyomSavchenko (approved):

server/account/src/utils.ts
spatialy: done

PR #10555, opened 2026-02-25
Configure who should be able to send invitation link

ArtyomSavchenko:
Added the following functionality:

- Configure what user roles should be able to send invitation link
- Add ability to select user role for users joining by invite link

Before:
Screenshot 2026-02-20 at 17 01 51
After:
Screenshot 2026-02-25 at 18 02 46
Screenshot 2026-02-25 at 18 02 52


BykhovDenis:
I'm not sure about default role, could be too dangerous

BykhovDenis (approved):

ArtyomSavchenko:
> I'm not sure about default role, could be too dangerous

Hi @BykhovDenis 
Thanks, I've updated the PR to allow only `Guest` and `User` as default role.

PR #10615, opened 2026-03-10
fix(login): guard logIn call when token is undefined on password signup

### LINKED ISSUE #10518: Bug: Password signup shows "Unknown error: Unexpected token" when email service is configured

varaprasadreddy9676:
## Summary

Fixes #10518

When `MAIL_URL` is configured (production deployments with email confirmation), `signUp` intentionally returns `token: undefined`. `SignupForm.svelte` was calling `logIn(result)` unconditionally, which triggered `setCookie()` without a valid token. The account service responded with 401 and an invalid JSON body, causing the client to crash with:

> Unknown error: Unexpected token 'N', "Not Found" is not valid JSON

The account **was** created successfully, but the user was stuck on the signup page and had to navigate to Sign In manually.

## Root cause

```svelte
// Before (buggy)
if (result != null) {
  await logIn(result)        // called even when result.token is undefined
  goTo('confirmationSend')
}
```

## Fix

Added a `result.token != null` guard before `logIn()`, matching the existing pattern already used correctly in the OTP flow (`doLoginNavigate` in `utils.ts`):

```svelte
// After
if (result != null) {
  if (result.token != null) {
    await logIn(result)
  }
  goTo('confirmationSend')
}
```

## Test plan

- [ ] Sign up with password when `MAIL_URL` is **not** configured — user is logged in and redirected normally
- [ ] Sign up with password when `MAIL_URL` **is** configured — user is redirected to "confirmation sent" page without error
- [ ] OTP signup flow is unaffected

### DIFF
--- a/plugins/login-resources/src/components/SignupForm.svelte
+++ b/plugins/login-resources/src/components/SignupForm.svelte
@@ -95,7 +95,9 @@
         status = loginStatus
 
         if (result != null) {
-          await logIn(result)
+          if (result.token != null) {
+            await logIn(result)
+          }
           goTo('confirmationSend')
         }
       }
--- a/plugins/notification-resources/src/components/inbox/SettingsPopup.svelte
+++ b/plugins/notification-resources/src/components/inbox/SettingsPopup.svelte
@@ -36,26 +36,31 @@
     if (key.code === 'ArrowUp') {
       key.stopPropagation()
       key.preventDefault()
-      list.select(selection - 1)
+      selectIndex(selection - 1)
       return true
     }
     if (key.code === 'ArrowDown') {
       key.stopPropagation()
       key.preventDefault()
-      list.select(selection + 1)
+      selectIndex(selection + 1)
       return true
     }
-    if (key.code === 'Enter') {
+    if (key.code === 'Home') {
+      key.stopPropagation()
+      key.preventDefault()
+      selectIndex(0)
+      return true
+    }
+    if (key.code === 'End') {
+      key.stopPropagation()
+      key.preventDefault()
+      selectIndex(items.length - 1)
+      return true
+    }
+    if (key.code === 'Enter' || key.code === 'Space') {
       key.preventDefault()
       key.stopPropagation()
-      items[selection]?.onToggle()
-      items = items.map((item, index) => {
-        if (index === selection) {
-          return { ...item, on: !item.on }
-        }
-
-        return item
-      })
+      toggleSelectedItem()
       return true
     }
     return false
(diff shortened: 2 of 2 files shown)


varaprasadreddy9676:
Closing this as the maintainers have indicated they rolled back a prior fix for #10518 and plan to test more thoroughly before merging. Happy to reopen or assist with testing when ready.

PR #10624, opened 2026-03-11
feat(account): revocable API tokens managed from account settings

dnplkndll:
## Summary

Personal API tokens with a built-in REST API, managed from account settings. Tokens are JWTs with a configurable expiry (7–365 days) that can be revoked, with expiry and revocation enforced centrally so the policy is reusable across services.

## What changed since the last review

Thanks for the review and the patience — conflicts are resolved and every comment is addressed. Two things also changed on their own merits:

**Token scopes are gone.** They were only consulted in the transactor's REST handler, so they advertised a boundary the platform did not keep. The WebSocket transport decodes the token and never looks at scopes, so a "Read Only" token could open a socket and issue any transaction; on the REST path itself `delete:*` only matched a top-level `TxRemoveDoc` and was bypassed by nesting the removal in a `TxApplyIf`; and nothing stopped a scoped token from calling `createApiToken` to mint an unscoped one. Narrowing a token's rights belongs in the pipeline, next to `GuestPermissionsMiddleware`, where it covers every transport — that is a larger design than this feature, and worth its own PR. Until then a token carries its account's rights and the UI says so, instead of showing a restriction that does not hold.

**`docs/openapi.yaml` is dropped.** It described a `/_tokens` minting service and a `tools/mint-token` CLI that do not exist in this repo, used reverse-proxy prefixes rather than the transactor's `/api/v1` routes, and stated that revocation was a future enhancement — which is what this PR implements. Nothing referenced it.

## Review comments

- **Revocation check was circular** (@aonnikov) — `verifyToken()` now sits alongside `decodeToken()` and verifies signature, expiry and, for revokable tokens, revocation through a pluggable checker (`setApiTokenRevocationChecker`). The account is authoritative: `wrap()` rejects revoked or expired tokens, so the transactor's checker just reuses `getLoginInfoByToken`. The bespoke `checkApiTokenRevoked` RPC is gone.
- **Transactor endpoint cannot be built client-side** (@aonnikov) — the REST base URL is derived from the login endpoint the account service returns, the same source `ServerManagerGeneral` uses.
- **Use the i18n dropdown** (@aonnikov) — done; the remaining hardcoded English labels in the token table are gone too.
- **Restrict to User and above, not guests** (@ArtyomSavchenko) — enforced server-side in `createApiToken`, and now covered by a test that fails if the check is removed.
- **Untranslated strings and locale parity** (@ArtyomSavchenko) — every locale is now at parity with `en.json`; Korean and Polish were the two still missing these keys. The API access strings orphaned when this PR replaced the old "Generate API token" button are removed from all 14 locales.

## Hardening

- API tokens can no longer create, list or revoke API tokens — otherwise a leaked token renews itself and revokes the tokens meant to stop it.
- Revocation fails closed. An unreachable account now rejects the token rather than admitting a revoked one to whoever can keep the account busy; verdicts stay trusted for the cache TTL so brief outages do not cut off healthy tokens. The cache is bounded and re-checks negative verdicts.
- Revoking your own token no longer requires a role in its workspace, so leaving a workspace cannot strand a credential you can never revoke.
- The per-account limit counts only usable tokens — revoked and expired ones are kept for the audit trail, and counting them would eventually lock out anyone who rotates.
- Rejected REST tokens are logged with a reason; expired, revoked and unverifiable are one opaque 401 from outside.
- The unused `listWorkspaceApiTokens` / `revokeWorkspaceApiToken` pair is removed — no caller, no test.

## Placement and UI

The settings page moved from workspace settings to account settings: tokens belong to the account, the page already listed them across every workspace, and creating one only needs `User` — the category required `Owner`, so the very users the role check was written for could not reach it. Failed loads and revokes are now surfaced instead of logged to the console, the copy dialog no longer dead-ends outside a secure context, and the docs disclosure and copy targets carry the expected aria attributes.

## Verification

- `rush validate` (full repo), `rush check`, `check-versions`, `fast-format`, `svelte-check` — all clean
- `rushx test` in `server/account` — 499 passed; the 4 `postgres-real` failures need a live database and fail identically on a clean checkout
- New tests cover the role restriction, the API-token guard on all three management methods, ownership on revoke, limit accounting and fail-closed revocation. Mutation-checked: stubbing out the guard fails 3, removing the role check fails 1.

## Known limitation

Revocation is checked when a WebSocket session is established, not continuously, so an already-open session outlives revocation until it reconnects. The REST path re-checks within the cache TTL. Closing that gap means re-validating live sessions in the session manager, which I've kept out of this PR.

https://claude.ai/code/session_01ANdoXbdn5k2hZy734EwKe7


dnplkndll:
## Follow-up: REST API should apply sensible defaults for `TxCreateDoc`

While testing the token management flow end-to-end (mint token → create issues via REST), I discovered that `POST /api/v1/tx/{workspace}` passes `TxCreateDoc` transactions through with zero default-filling. Issues created this way are missing fields that the UI auto-populates, causing them to not appear in project views (only in all-issues).

### Missing fields for `tracker:class:Issue`

When the UI creates an issue, it uses `addCollection()` which wraps the `TxCreateDoc` in a `TxCollectionCUD`, setting:
- `attachedTo: tracker:ids:NoParent`
- `attachedToClass: tracker:class:Issue`
- `collection: "subIssues"`

Plus these attributes that get no server-side defaults:
- `kind` → must be `tracker:taskTypes:Issue` (not `tracker:ids:ClassingProjectType`)
- `childInfo: []`, `parents: []`, `subIssues: 0`, `comments: 0`, `reports: 0`
- `assignee: null`, `component: null`, `dueDate: null`
- `estimation: 0`, `remainingTime: 0`, `reportedTime: 0`

### Impact

Issues created via bare `TxCreateDoc` (the natural REST usage) are invisible in project views because the Svelte live query filters on `attachedTo`/`collection`. They show up in the all-issues aggregated view only.

### Possible approaches

1. **Document it** — update REST API docs to show the required `TxCollectionCUD` wrapper pattern for issue creation
2. **Server-side defaults middleware** — a new middleware in the tx pipeline that merges missing fields from the model schema for known classes (significant arch change)
3. **Higher-level REST endpoints** — add `POST /api/v1/create-issue/{workspace}` that accepts a simplified payload and constructs the full `TxCollectionCUD` server-side, like the importer does

Option 3 seems most API-friendly without changing the platform's "caller provides everything" philosophy. Happy to contribute this if there's interest.

ArtyomSavchenko:
Hi @dnplkndll 
Thank you for your contribution.
Could you please check formatting for server/account/src/operations.ts?
The corresponding CI step is failed: https://github.com/hcengineering/platform/actions/runs/23164707679/job/67348370400?pr=10624
You can use `rush format` to format all project files or `rushx format` for a specific package.

dnplkndll:
@ArtyomSavchenko Thanks for the review! Formatting has been fixed — the issue was a prettier version mismatch (local 3.8.1 vs project's 3.6.2). New code now matches the existing codebase style with no formatting noise on existing lines.

### Changes in this update

**3 commits:**

1. **`feat: API token management`** — core feature (create/list/revoke tokens, UI, DB, i18n)
2. **`docs: token scope roadmap`** — roadmap comments for future scope-based access control
3. **`feat: enforce API token revocation`** — addresses a gap where revoked tokens remained usable until JWT expiry

### What's new since last push

- **Token revocation enforcement** — the transactor now checks a per-token revocation cache (60s TTL) before granting API access. Revoked tokens are rejected within ~60 seconds instead of remaining valid until expiry.
- **Owner-level workspace token visibility** — workspace owners can list and revoke any member's API tokens via `listWorkspaceApiTokens` / `revokeWorkspaceApiToken` (OWNER role required).
- **Ledo instance URL removed** — `openapi.yaml` now uses a generic `https://{host}` server variable.

### Suggestions for future consideration

1. **Token scopes** — the roadmap commit documents a proposed `extra.scopes` JWT field for fine-grained access control (read-only tokens, class-restricted access). This would enable safe MCP/AI agent integrations without MCP-layer changes.

2. **Higher-level REST endpoints** — as noted in my earlier comment, `POST /api/v1/tx` requires the caller to construct full `TxCollectionCUD` wrappers. A `POST /api/v1/create-issue` style endpoint would make the API much more accessible for integrations.

3. **Revocation cache scaling** — the current per-token cache works well for moderate usage. For high-traffic deployments, a Redis-backed revocation set or short-lived tokens (< 1 hour) with refresh would be more robust.

plugins/setting-assets/lang/pt.json
```diff
@@ -210,6 +210,49 @@
...
+    "ApiUsageDescription": "Use your API token with the built-in REST API to query and modify workspace data. Pass the token as a Bearer token in the Authorization header.",
+    "ApiUsageTitle": "Using the REST API",
+    "ApiWorkspaceId": "Your workspace ID (UUID) is included in the token. Pass it as :workspaceId in the URL.",
+    "CountSpaces": "{count, plural, =0 {No spaces} =1 {# space} other {# spaces}}",
+    "CreateApiToken": "Create token",
+    "Created": "Created",
+    "Description": "Description",
+    "Expires": "Expires",
+    "General": "General",
+    "NewSpaceType": "New space type",
```
ArtyomSavchenko, line 266: Could you check that the same translation keys have been added across all locales?
It looks like there are some unused translations (e.g. NewSpaceType), so the locale tests are failing now: https://github.com/hcengineering/platform/actions/runs/23206455358/job/67521658706?pr=10624

server/account/src/__tests__/apiTokenScopes.test.ts
```diff
@@ -0,0 +1,249 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import {
+  AccountRole,
+  type MeasureContext,
```
ArtyomSavchenko: CI formatting is failed in this file:
https://github.com/hcengineering/platform/actions/runs/23206455358/job/67521658723?pr=10624

server/account/src/operations.ts
```diff
@@ -2353,6 +2354,256 @@ async function deleteMailbox (
...
+  const days = Math.floor(expiryDays)
+  if (days < 1 || days > 365) {
+    throw new PlatformError(new Status(Severity.ERROR, platform.status.BadRequest, {}))
+  }
+
+  const { account } = decodeTokenVerbose(ctx, token)
+
+  // Verify the user has access to this workspace
+  const role = await db.getWorkspaceRole(account, workspaceUuid)
+  if (role == null) {
```
ArtyomSavchenko, line 2735: I think it would be better to restrict the roles of those who can use API tokens.
I think they should be available only to users with the "User" role or higher, rather than to guests.
For example, similar check can be used:
`verifyAllowedRole(callerRole, AccountRole.User, extra)`

server/account/src/__tests__/apiTokenScopes.test.ts
```diff
@@ -0,0 +1,250 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import { AccountRole, type MeasureContext, type PersonUuid, type WorkspaceUuid } from '@hcengineering/core'
+import platform, { PlatformError, Severity, Status } from '@hcengineering/platform'
+import { decodeTokenVerbose, generateToken } from '@hcengineering/server-token'
```
ArtyomSavchenko: Could you please check warnings in this file?
CI formatting step is failed due to this:
https://github.com/hcengineering/platform/actions/runs/23317689234/job/67851578625?pr=10624

ArtyomSavchenko:
Hi @dnplkndll 
Could you please resolve conflicts?

dnplkndll:
sure what is your opinion on the use of MCP versus api? My motivation was that I would prefer to have tool access in a controlled predicable / scriptable way with permissions on things. then you can test on a staging build out script you can repeat. 
there are a few things I ran into with the api that the UI handles that should probably be refined, like defaults, too. to simplify not putting records in a bad state.

plugins/setting-assets/lang/ja.json
```diff
@@ -228,6 +228,45 @@
...
+    "ApiTokenExpiry90Days": "90 days",
+    "ApiTokenExpiry180Days": "180 days",
+    "ApiTokenExpiry365Days": "365 days",
+    "ApiTokenLoadError": "Failed to load API tokens",
+    "ApiTokenCreateError": "Failed to create token. Please try again.",
+    "ApiTokenPermissions": "Permissions",
+    "ApiTokenScopePreset": "Permissions",
+    "ApiTokenScopeReadOnly": "Read Only",
+    "ApiTokenScopeReadWrite": "Read & Write",
+    "ApiTokenScopeFullAccess": "Full Access"
```
ArtyomSavchenko: Could you please check translations in this file and others?
Looks like many sentences were not translated and remained in English.

plugins/setting-resources/src/components/ApiDocsSection.svelte
```diff
@@ -0,0 +1,240 @@
...
+  import { copyTextToClipboard } from '@hcengineering/presentation'
+  import setting from '@hcengineering/setting'
+
+  let showApiDocs = false
+
+  function getBaseUrl(): string {
+    return window.location.origin
+  }
+
+  $: baseApiUrl = getBaseUrl() + '/_transactor/api/v1'
```
aonnikov: That's not correct, transactor endpoint address cannot be built in this way, it is returned by account service when user is authenticated.

plugins/setting-resources/src/components/ApiTokenCreatePopup.svelte
```diff
@@ -0,0 +1,219 @@
...
+    { _id: 'read-only', label: 'Read Only', scopes: ['read:*'] },
+    { _id: 'read-write', label: 'Read & Write', scopes: ['read:*', 'write:*'] },
+    { _id: 'full-access', label: 'Full Access', scopes: ['read:*', 'write:*', 'delete:*'] }
+  ]
+  let scopePresetItems: ListItem[] = scopePresets.map((p) => ({ _id: p._id, label: p.label }))
+  let selectedScopePreset: ListItem = scopePresetItems[0]
+
+  async function resolveScopeLabels(): Promise<void> {
+    const lang = $themeStore.language
+    const labels = await Promise.all([
```
aonnikov: There is no need to manually translate strings, we have dropdown version that uses i18n strings

pods/server/src/rpc.ts
```diff
@@ -129,14 +129,71 @@ async function sendJson (
...
+  const now = Date.now()
+  const cached = revocationCache.get(apiTokenId)
+
+  // Permanently cached once revoked
+  if (cached?.revoked === true) return true
+
+  // Re-check if stale or missing
+  if (cached == null || now - cached.checkedAt > REVOCATION_CACHE_TTL_MS) {
+    try {
+      const revoked = await accountClient.checkApiTokenRevoked(apiTokenId)
```
aonnikov: This does not make sense.

- User comes with revokable token that contains token Id (that is how we identify revokable token).
- We use this token (that is potentially revoked or expired at the moment)  to check whether the  token is revoked

If the token is revoked, the account cannot reply anything other than 401 or 403, because the token is revoked/expired.

The account method, that returns boolean is redundant. We can use `selectWorkspace` or `getWorkspaceInfo` for this.

Additionally, I would implement token expiration / revocation logic and lower level. Maybe at the same level as `decodeToken`, for example we can add a new `verifyToken` method next to the `decodeToken`, that:

1. Decodes token
2. Checks whether the token is expired
3. If the token is revokable, checks whether token is expired
4. Returns decoded token

This will allow reuse token expiration / revocation logic in services to prevent accessing blobs, or other data with expired tokens.

But in order to do this, we will have to additionally set up some metadata in the server-token plugin (either address of the account endpoint, or method to verify whether the token is expired).

aonnikov:
Thanks for the contribution 
Overall the idea is great, but need to fix architectural and UI issues.

We also have a button to generate token on workspace settings page, we will need to remove it in favor of the new approach.

dnplkndll:
Thanks for the reviews — pushed an update addressing the feedback. Conflicts with `develop` are resolved (merged current `develop`; the API token migration is renumbered to **V27** after develop's V26 `pending_configuration`).

**@aonnikov**
- *Revocation logic / redundant boolean method:* reworked. Token expiry + revocation now live in a new `verifyToken()` next to `decodeToken()` in `server-token`, with a pluggable checker (`setApiTokenRevocationChecker`) so the policy is reusable across services. The account is now authoritative — `wrap()` rejects revoked/expired API tokens, so `selectWorkspace`/`getWorkspaceInfo`/etc. return 401 naturally. The dedicated `checkApiTokenRevoked` RPC is **removed**; the transactor's checker reuses the existing `getLoginInfoByToken`.
- *Transactor endpoint in `ApiDocsSection`:* fixed — the REST base URL is now derived from the account-provided transactor endpoint (ws→http), matching the `ServerManagerGeneral` pattern, instead of `window.location`.
- *Manual translations in the create popup:* it uses `DropdownLabelsIntl` with i18n strings (no hand-rolled translations).

**@ArtyomSavchenko**
- *Role restriction:* token creation/revocation now require `AccountRole.User` or higher via `verifyAllowedRole(...)` (guests excluded).
- *Locale tests / untranslated strings:* en/ru parity restored and the API token strings are now translated across all locales (cs/de/es/fr/it/ja/pt/pt-br/ru/tr/zh).
- *Formatting:* re-ran the project formatter; `operations.ts` and the scope test are clean.

Unit tests added for `verifyToken` (signature/expiry/revocation) and scope enforcement. Happy to iterate further.

ArtyomSavchenko:
Hi @dnplkndll 
Sorry for the delay in the review.
Could you please resolve the merge conflicts?
If you don't have time for this, let me know, and I can finish it.

ArtyomSavchenko (approved):

aonnikov (approved):

PR #10626, opened 2026-03-12
fix(text-editor): fix heading hierarchy and prose spacing

ComputerCrack:
Problem

The editor's typography had several issues:

    h1–>h3 only!
    h4–>h6 had no styles, falling back to browser defaults which were visually
    indistinguishable from body text
    150% line-height on headings
    semantically correct but visually too loose for large font sizes;
    tight letter-spacing on display sizes is standard practice
    Double paragraph margins
    margin-block-start + margin-block-end on < p > combined with margin-top on >+ produced inconsistent spacing between paragraphs and other block elements
    Horizontal rule
    1px, no vertical breathing room, visually weak
    List bullets misaligned
    ::before pseudo-element had a fixed line-height: 1.65 while positioned
    absolute, causing bullets to sit higher than the text baseline
    List indent too narrow
    0.75rem made nested content feel cramped

Changes

    h1–>h6
    fully defined with proportional font-size, line-height, letter-spacing and margins
    following a clear visual scale
    Paragraph spacing moved entirely to >+ (owl selector)
    single source of truth for block spacing, no more margin collapsing issues
    HR: 2px, 1.5rem vertical margin
    List bullet ::before: line-height: inherit + top: 0
    bullets now vertically centered with text at all font sizes
    List indent increased to 1.5rem
    prose.scss aligned to the same system

### DIFF
--- a/packages/theme/styles/_text-editor.scss
+++ b/packages/theme/styles/_text-editor.scss
@@ -5,7 +5,7 @@
 
   &.text-editor-view {
     font: inherit;
-    line-height: 150%;
+    line-height: 1.65;
   }
 }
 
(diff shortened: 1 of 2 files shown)


ArtyomSavchenko:
Hi @ComputerCrack 
Thank you for the contribution!
Could you please let me know if the **Draft** label is still actual, or if the PR is already ready for review?

ComputerCrack:
Hey @ArtyomSavchenko,

It's an actual draft. I thought I was happy with the results but in the following days I found that there are many more issues with the text editors styling properties.
The current plan is to continously watch and change things on our instance. When I get to a point where I'm happy with the result I will update the PR accordingly.

It's not a priority atm tho because where working on a gitea integration. 
I'll keep you updated.

PR #10666, opened 2026-03-21
fix(view): "Not Specified" group shows assigned issues (#10606)

### LINKED ISSUE #10606: Not Specified” assignee group incorrectly includes issues with assigned users
When grouping or sorting issues by Person / Assignee and then by Status, the “Not Specified” assignee section incorrectly contains issues that do have assignees. This is especially noticeable for issues in Backlog and Ready for QA, where the assignee is clearly set but the issue still appears under “Not Specified.”

dnplkndll:
## Summary

- Fix "Not Specified" assignee group incorrectly including issues that have assignees when using secondary grouping (e.g., group by Assignee → then by Status)
- Root cause: `getGroupByKey()` in `ListCategories.svelte` passed `undefined` as the query value for the "Not Specified" category — `undefined` is stripped during JSON serialization, so the server received no filter and returned all documents
- Fix: use `null` instead of `undefined` so the filter survives serialization and correctly matches only documents where the field is unset

## Test plan

- [ ] Open Tracker → All Issues view
- [ ] Group by Assignee, then add a secondary grouping by Status
- [ ] Verify "Not Specified" section only contains issues with no assignee
- [ ] Verify assigned issues appear only under their correct assignee group
- [ ] Test with other groupable Person/Ref fields (e.g., component, milestone) to confirm no regression

Fixes #10606

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/foundations/core/packages/core/src/__tests__/query.test.ts
+++ b/foundations/core/packages/core/src/__tests__/query.test.ts
@@ -1,0 +1,65 @@
+//
+// Copyright © 2026 Hardcore Engineering Inc.
+//
+// Licensed under the Eclipse Public License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License. You may
+// obtain a copy of the License at https://www.eclipse.org/legal/epl-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


ArtyomSavchenko (approved):

ArtyomSavchenko:
Hi @dnplkndll 
Thank you for the contribution.
I have fixed formatting in separate PR: https://github.com/hcengineering/platform/pull/10668.

PR #10672, opened 2026-03-23
feat(ai-bot): make bot account email configurable via AI_BOT_EMAIL env var

### LINKED ISSUE #239: Allow aibot email to be changed
Right now, the used email is hardcoded in the `bundle.js`:

```js
    var aiBotId = "ai-bot";
    var aiBotAccountEmail2 = "huly.ai.bot@hc.engineering";
    var aiBotEmailSocialKey3 = (0, import_core58.buildSocialIdString)({
      type: import_core58.SocialIdType.EMAIL,
      value: aiBotAccountEmail2
    });
```

I would expect this email to be configurable via environment variables, similar to the first and last name.
Changing the line "aiBotAccountEmail2" in the `bundle.js` inside the docker container manually solves this, but this is not a good solution.

The hardcoded Email leads to the system sending emails to the hardcoded address "huly.ai.bot@hc.engineering", especially when one pings the bot or writes a PM to the bot whilst offline.

Sadly I did not find a repository for the AIBot, or I would have provided a PR for this.

dnplkndll:
## Summary

Makes the AI bot account email configurable via `AI_BOT_EMAIL` env var using the platform Metadata pattern (per @aonnikov's review feedback), instead of mutable module state.

Resolves hcengineering/huly-selfhost#239

## Motivation

The bot email `huly.ai.bot@hc.engineering` is an `@hc.engineering` address embedded in every self-hosted deployment's account system. Self-hosters have no control over this domain, and it can cause bounce-back noise if an external SMTP notification service attempts to email all workspace members.

## Design

Uses platform `Metadata` — the same pattern as `FrontVersion`, `EndpointURL`, and other startup config values:

```ts
// Plugin definition (plugins/ai-bot/src/index.ts)
const aiBot = plugin(aiBotId, {
  metadata: {
    EndpointURL: '' as Metadata<string>,
    BotEmail: '' as Metadata<string>        // ← new
  }
})

// Startup (transactor + ai-bot service)
setMetadata(aiBot.metadata.BotEmail, process.env.AI_BOT_EMAIL)

// Consumers read via getter with default fallback
export function getAiBotAccountEmail (): string {
  return getMetadata(aiBot.metadata.BotEmail) ?? aiBotAccountEmail
}
```

- No mutable module state (`set`/`get` function pairs)
- Original `aiBotAccountEmail` and `aiBotEmailSocialKey` constants kept for client-side / default use
- Empty-string guard on env var reads

## Env vars

| Service | Var | Default | Required |
|---------|-----|---------|----------|
| ai-bot | `AI_BOT_EMAIL` | `huly.ai.bot@hc.engineering` | No |
| transactor | `AI_BOT_EMAIL` | `huly.ai.bot@hc.engineering` | No |

Both must use the same value if set.

## Changes (11 files)

| File | Change |
|------|--------|
| `plugins/ai-bot/src/index.ts` | Add `BotEmail` metadata key, `getAiBotAccountEmail()`, `getAiBotEmailSocialKey()` |
| `pods/server/src/__start.ts` | Set metadata from `AI_BOT_EMAIL` env var |
| `services/ai-bot/pod-ai-bot/src/start.ts` | Set metadata from `AI_BOT_EMAIL` env var |
| `services/ai-bot/pod-ai-bot/src/utils/account.ts` | Use getter functions |
| `services/ai-bot/pod-ai-bot/src/utils/platform.ts` | Use getter functions |
| `foundations/server/packages/middleware/src/identity.ts` | Import from `@hcengineering/ai-bot` (was duplicate hardcoded string) |
| `foundations/server/packages/middleware/src/pluginConfig.ts` | Use getter function |
| `server-plugins/ai-bot-resources/src/index.ts` | Use getter function |
| `**/package.json` + `pnpm-lock.yaml` | Add `@hcengineering/ai-bot` dependency |

**Not changed:** `plugins/ai-bot-resources/src/utils.ts` (client-side, correctly uses constant — metadata is not available client-side).

## Known limitation

Client-side code (`aiBotSocialIdentityStore` in Love/video meetings) queries by the static default social key. If a custom email is configured server-side, the client query won't match. This is acceptable for the current server-side scope — full client support would require pushing the configured email through workspace config.

## Testing

Deployed to self-hosted K8s cluster with `AI_BOT_EMAIL=aibot@ledoweb.com`:

- ai-bot: 0 restarts, registers new social identity in CockroachDB
- Original identity preserved (backwards compat)
- DM with Huly AI works — bot receives and responds
- Default behavior (no env var) unchanged

### DM with bot (custom email configured)

DM with Huly AI

### Bot responding in Chunter

Bot response

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pods/server/src/__start.ts
+++ b/pods/server/src/__start.ts
@@ -3,6 +3,7 @@
 //
 
 // Add this to the VERY top of the first file loaded in your app
+import { setAiBotAccountEmail } from '@hcengineering/ai-bot'
 import { Analytics } from '@hcengineering/analytics'
 import { configureAnalytics, createOpenTelemetryMetricsContext, SplitLogger } from '@hcengineering/analytics-service'
 import contactPlugin from '@hcengineering/contact'
@@ -85,6 +86,9 @@ setMetadata(serverNotification.metadata.MailUrl, config.mailUrl ?? '')
 setMetadata(serverNotification.metadata.MailAuthToken, config.mailAuthToken)
 setMetadata(serverNotification.metadata.WebPushUrl, config.webPushUrl)
 setMetadata(serverAiBot.metadata.EndpointURL, process.env.AI_BOT_URL)
+if (process.env.AI_BOT_EMAIL !== undefined) {
+  setAiBotAccountEmail(process.env.AI_BOT_EMAIL)
+}
 setMetadata(serverCalendar.metadata.EndpointURL, process.env.CALENDAR_URL)
 setMetadata(serverCard.metadata.CommunicationEnabled, process.env.COMMUNICATION_API_ENABLED === 'true')
 
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -5020,6 +5020,9 @@ importers:
 
   ../../foundations/server/packages/middleware:
     dependencies:
+      '@hcengineering/ai-bot':
+        specifier: workspace:^0.7.0
+        version: link:../../../../plugins/ai-bot
       '@hcengineering/analytics':
         specifier: workspace:^0.7.17
         version: link:../../../core/packages/analytics
@@ -31130,6 +31133,9 @@ importers:
       '@hcengineering/account-client':
         specifier: workspace:^0.7.21
         version: link:../../foundations/core/packages/account-client
+      '@hcengineering/ai-bot':
+        specifier: workspace:^0.7.0
+        version: link:../../plugins/ai-bot
       '@hcengineering/analytics':
         specifier: workspace:^0.7.17
         version: link:../../foundations/core/packages/analytics
@@ -61557,7 +61563,7 @@ snapshots:
   node-loader@2.0.0(webpack@5.102.1):
     dependencies:
       loader-utils: 2.0.4
-      webpack: 5.102.1
+      webpack: 5.102.1(esbuild@0.25.12)(webpack-cli@5.1.4)
 
   node-localstorage@2.2.1:
     dependencies:
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -5020,6 +5020,9 @@ importers:
 
   ../../foundations/server/packages/middleware:
     dependencies:
+      '@hcengineering/ai-bot':
+        specifier: workspace:^0.7.0
+        version: link:../../../../plugins/ai-bot
       '@hcengineering/analytics':
         specifier: workspace:^0.7.19
         version: link:../../../core/packages/analytics
@@ -31157,6 +31160,9 @@ importers:
       '@hcengineering/account-client':
         specifier: workspace:^0.7.25
         version: link:../../foundations/core/packages/account-client
+      '@hcengineering/ai-bot':
+        specifier: workspace:^0.7.0
+        version: link:../../plugins/ai-bot
       '@hcengineering/analytics':
         specifier: workspace:^0.7.19
         version: link:../../foundations/core/packages/analytics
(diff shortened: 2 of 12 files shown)


dnplkndll:
**Alternative approach:** Since no actual email is ever sent to or from this address (it's purely an account identity key), a simpler fix would be to just change the hardcoded default from `huly.ai.bot@hc.engineering` to something domain-neutral like `ai-bot@huly` — no env var needed.

The reason I went with the configurable approach is backwards compatibility: existing deployments already have the bot registered with `huly.ai.bot@hc.engineering` in their `social_id` table. Changing the constant would require a migration to update the social identity, otherwise the bot would create a duplicate account on next startup.

If a migration is acceptable, the simpler rename would be a ~5-line change. Happy to pivot to that approach if preferred.

pods/server/src/__start.ts
```diff
@@ -85,6 +86,9 @@ setMetadata(serverNotification.metadata.MailUrl, config.mailUrl ?? '')
 setMetadata(serverNotification.metadata.MailAuthToken, config.mailAuthToken)
 setMetadata(serverNotification.metadata.WebPushUrl, config.webPushUrl)
 setMetadata(serverAiBot.metadata.EndpointURL, process.env.AI_BOT_URL)
+if (process.env.AI_BOT_EMAIL !== undefined) {
```
aonnikov, line 89: It is idiomaic to use metadata for such kind of configuration, i.e. something like:
```
setMetadata(serverAiBot.metadata.Email, process.env.AI_BOT_EMAIL)
```
and in the ai-bot plugin:

```
export function getAiBotEmailSocialKey (): string {
  return buildSocialIdString({
    type: SocialIdType.EMAIL,
    value:getMetadata(...) ?? aiBotAccountEmail
  })
}
```

aonnikov: Honestly, I don't think it has to be a email, ideally that should be a separate SocialIdType, but this is something that requires larger refactoring.

ArtyomSavchenko:
@dnplkndll Looks like the following commit should be in a separate PR: [feat(mobile): add Flutter iOS app with OAuth fix and Huly UX redesign](https://github.com/hcengineering/platform/pull/10672/changes/b029841ce7ee7e2c43bca8ee3d5e4da9fae7ad1b) .

screenshots/08-dm-with-bot.png
ArtyomSavchenko, line 1: Could you explain why these screenshots are needed in the code base?

pods/server/src/__start.ts
```diff
@@ -85,6 +86,9 @@ setMetadata(serverNotification.metadata.MailUrl, config.mailUrl ?? '')
 setMetadata(serverNotification.metadata.MailAuthToken, config.mailAuthToken)
 setMetadata(serverNotification.metadata.WebPushUrl, config.webPushUrl)
 setMetadata(serverAiBot.metadata.EndpointURL, process.env.AI_BOT_URL)
+if (process.env.AI_BOT_EMAIL !== undefined && process.env.AI_BOT_EMAIL !== '') {
+  setMetadata(aiBot.metadata.BotEmail, process.env.AI_BOT_EMAIL)
+}
```
ArtyomSavchenko, line 91: I think it is better to set this metadata in [dev/prod/src/platform.ts](https://github.com/hcengineering/platform/blob/38df5a3cdb2529248eb1cb15be0f18ffc3d7253a/dev/prod/src/platform.ts#L518) and [desktop/src/ui/platform.ts](https://github.com/hcengineering/platform/blob/38df5a3cdb2529248eb1cb15be0f18ffc3d7253a/desktop/src/ui/platform.ts#L399), since `@hcengineering/ai-bot` is not server plugin.

aonnikov: No, the code is correct. This is needed for the transactor. platform.ts is for UI

aonnikov (approved):

PR #10715, opened 2026-04-02
[codex] Fix dev deployment defaults and login persistence

WaaSakura:
## What changed

This PR cleans up a set of dev deployment issues that surfaced while bringing up the platform end to end, and also makes login state more resilient across refreshes and workspace navigation.

## Why

The dev stack mixed browser-facing hostnames with container-to-container traffic, which caused services inside Docker to call `huly.local` instead of other compose services. That breaks workspace bootstrap, account lookups, and other cross-service flows once the stack is exposed through remapped ports or a non-local hostname.

At the same time, the login flow only kept the token in volatile metadata. Refreshing or returning to a workspace route could drop the token and strand the user in loading flows even though the account session was still valid.

## Highlights

- switch internal compose traffic to Docker service names where appropriate
- make browser-facing host and port settings overridable, while preserving the old local defaults
- fix the Redpanda healthcheck so the service reports healthy with the actual broker configuration
- update the Hulypulse dev setup to use a valid published image and avoid binding Redis to a host port by default
- make the Hulypulse Dockerfile work without relying on buildx-only platform args
- persist the login token in local storage and restore it when reopening login / onboard / workbench flows
- downgrade the noisy "no document found, failed to apply model transaction" client message from warning to info
- add retry logic when external image pulls are flaky
- avoid failing workspace image builds when the optional init scripts directory is absent

## Impact

- dev deployments are more robust in Docker and less tied to `huly.local`
- front-end refresh / workspace reconnect flows are less likely to get stuck due to missing token metadata
- the default dev console is less noisy and better reflects actual service health

## Validation

- `docker compose -f dev/docker-compose.yaml config`
- rebuilt the front image and verified the app loads into a workspace successfully
- verified the browser console reaches `0 errors / 0 warnings` with the staged changes
- verified Redpanda becomes `healthy` with the updated healthcheck

### DIFF
--- a/plugins/workbench-resources/src/utils.ts
+++ b/plugins/workbench-resources/src/utils.ts
@@ -212,6 +212,7 @@ export async function logIn (loginInfo: { account: string, token?: string }): Pr
   await getAccountClient(accountsUrl, loginInfo.token).setCookie()
 
   setMetadata(presentation.metadata.Token, loginInfo.token)
+  setMetadataLocalStorage(login.metadata.LoginToken, loginInfo.token ?? null)
   setMetadataLocalStorage(login.metadata.LastAccount, loginInfo.account)
   setMetadataLocalStorage(login.metadata.LoginAccount, loginInfo.account)
 }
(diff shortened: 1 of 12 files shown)


plugins/workbench-resources/src/utils.ts
```diff
@@ -212,6 +212,7 @@ export async function logIn (loginInfo: { account: string, token?: string }): Pr
   await getAccountClient(accountsUrl, loginInfo.token).setCookie()
 
   setMetadata(presentation.metadata.Token, loginInfo.token)
+  setMetadataLocalStorage(login.metadata.LoginToken, loginInfo.token ?? null)
```
aonnikov, line 215: We cannot save token in the local storage. It is easy to steal it from there

PR #10718, opened 2026-04-03
Remove mobile not supported placeholder

ArtyomSavchenko:


BykhovDenis:
Why? We still not support it)

ArtyomSavchenko:
> Why? We still not support it)

Some basic things can be done with the current mobile view and I didn't see any critical problems on Android.
In my opinion, the banner solves no problems and does not improve the impression of the product.
If users report bugs in the mobile view, I or other community members could fix them.

PR #10723, opened 2026-04-04
feat: Support find in desktop client

ArtyomSavchenko:
Screenshot 2026-04-04 at 21 05 42


BykhovDenis (approved):
Not sure that we really need all these changes

ArtyomSavchenko:
> Not sure that we really need all these changes

Initially, there were fewer changes.
However, when the dialog was displayed in the standard mode, it was part of the main page document, and search results were found within it as well.

PR #10778, opened 2026-04-16
Fix UI freeze on controlled doc update view

ArtyomSavchenko:
Before(UI freeze after show more click):
Screenshot 2026-04-16 at 21 49 13
After:
Screenshot 2026-04-16 at 21 49 53

### DIFF
--- a/plugins/text-editor-resources/src/components/MarkupDiffViewer.svelte
+++ b/plugins/text-editor-resources/src/components/MarkupDiffViewer.svelte
@@ -15,8 +15,11 @@
 //
 -->
 <script lang="ts">
+  import { Analytics } from '@hcengineering/analytics'
   import { Class, Doc, Ref } from '@hcengineering/core'
+  import { getEmbeddedLabel } from '@hcengineering/platform'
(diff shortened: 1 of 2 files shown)


plugins/text-editor-resources/src/components/MarkupDiffViewer.svelte
```diff
@@ -77,33 +88,66 @@
...
 </script>
 
 <div class="ref-container">
-  <div class="textInput">
+  <div class="textInput" hidden={editorMountFailed}>
     <div class="select-text" style="width: 100%;" bind:this={element} />
   </div>
+  {#if editorMountFailed}
+    <div class="error-fallback">
+      <Label label={getEmbeddedLabel('Unable to show this diff.')} />
```
aonnikov: Why embedded label? Just show an error or use translated string

ArtyomSavchenko: Thanks, fixed.

aonnikov (approved):

PR #10786, opened 2026-04-20
Fix tracker project actions

ArtyomSavchenko:
Before:
Screenshot 2026-04-20 at 08 18 41
After:
Screenshot 2026-04-20 at 08 18 08

### DIFF
--- a/plugins/view-resources/src/visibilityTester.ts
+++ b/plugins/view-resources/src/visibilityTester.ts
@@ -27,6 +27,16 @@ import { getClient } from '@hcengineering/presentation'
 import { get } from 'svelte/store'
 import { spaceSpace } from './utils'
 
+function convertToSingleDoc (doc?: Doc | Doc[]): Doc | undefined {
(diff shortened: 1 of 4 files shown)


plugins/view-resources/src/visibilityTester.ts
```diff
@@ -27,6 +27,16 @@ import { getClient } from '@hcengineering/presentation'
 import { get } from 'svelte/store'
 import { spaceSpace } from './utils'
 
+function convertToSingleDoc (doc?: Doc | Doc[]): Doc | undefined {
```
aonnikov: Maybe this should be a part of the code that calls visibiliy testers. I see couple of options:
- always use array and check array length
- when array contains only one element, pss the element instead of array

ArtyomSavchenko: Thanks, fixed.

aonnikov (approved):

PR #10793, opened 2026-04-22
Fix link action

### LINKED ISSUE #10790: Adding link on document
Whenever I try to add link on document, the tool thing comes on top sometime on top of the Save Button which make adding links dificult
Image

ArtyomSavchenko:
Before:
Screenshot 2026-04-22 at 12 41 27
After:
Screenshot 2026-04-22 at 12 41 42


plugins/text-editor-resources/src/components/extension/toolbar/toolbar.ts
```diff
@@ -267,7 +267,8 @@ export function ToolbarControlPlugin (editor: Editor, options: ToolbarOptions):
         placement: viewOptions?.placement,
         hideOnClick: 'toggle',
         appendTo: viewOptions?.floating !== true ? (options.popupContainer ?? document.body) : document.body,
-        zIndex: 10001,
+        // Stay below application popups (base z-index 10000 in packages/ui Popup.svelte) so Link, Note, etc. are usable.
+        zIndex: 9999,
```
aonnikov: This may break something, ideally we should avoid using fixed zindex and issue them via some function that is reused for popups and editor

PR #10848, opened 2026-05-18
fix(fulltext): bump model version to 0.7.423 to match deployed workspaces

MichaelUray:
## Summary

The compiled `MODEL_VERSION` in fulltext-pod was `0.7.422` while workspaces are
already on `0.7.423`. Tx events were rejected with a `wrong version` warning at
the indexer, so every new document landed in ES with a missing entry — the
symptom was that a Tracker title-search would return zero hits even for issues
that exist in the workspace.

This patch bumps `common/scripts/version.txt` from `0.7.422` to `0.7.423`. The
same file is the build-time source-of-truth for `MODEL_VERSION` via
`common/scripts/show_version.js`, so rebuilding any pod from this branch will
embed `0.7.423` in its `bundle/model.json`.

## Root-cause investigation

After the \`0.7.422 → 0.7.423\` model migration was applied to workspaces,
deploys that hadn't bumped \`version.txt\` continued to ship a fulltext-pod
compiled for \`0.7.422\`. The pod connects to the Tx queue fine, but every
incoming event for a workspace on a higher patch is rejected with:

\`\`\`
supportedVersion: 0.7.422  (compiled into fulltext-pod)
version:          0.7.423  (workspace tx payload)
workspace:        <uuid>
\`\`\`

No retries, no indexing. New workspaces never get their documents indexed; old
workspaces miss every post-migration document. Search silently returns zero hits
for any term that only appears in post-migration documents.

## Acceptance criterion

After deploying a fulltext-pod built from this branch, the workspaces that were
upgraded to \`0.7.423\` start indexing their backlog. Direct ES verification:

\`\`\`
curl -s -X POST localhost:9200/huly_storage_index_v2/_search \\
  -H 'Content-Type: application/json' \\
  -d '{\"query\":{\"multi_match\":{\"query\":\"<any newly indexed term>\",
       \"fields\":[\"searchTitle\",\"fulltextSummary\",\"identifier\"]}}}'
\`\`\`

returns hits within a few minutes of the pod restart (no manual reindex needed
— the pod re-consumes the buffered queue).

## Future-prevention (out of scope for this PR)

The deploy pipeline does not currently fail when a workspace's persisted
\`core:class:Version\` patch exceeds the patch baked into any deployed pod's
\`bundle/model.json\`. A version-skew detector that refuses to publish in that
state would catch this class of bug at CI time. Tracking that separately.

## Test plan

- [x] Built fulltext-pod from this branch on a staging environment.
- [x] Verified the pod logs no \`wrong version\` warnings against workspaces on \`0.7.423\`.
- [x] Verified ES doc count grows after the pod restart (no manual reindex).
- [x] Title-search in Tracker UI returns hits for newly indexed documents.


MichaelUray:
Closing — this PR's diff was unintentionally scoped against `hcengineering/platform:develop`, which made it carry ~250 unrelated commits from a parallel Gantt-feature branch on my fork.

I'll reopen this as a focused PR cherry-picked onto a fresh branch off `develop` once the related Gantt feature stack (8 sequential PRs) lands. Apologies for the noise.

MichaelUray:
The fulltext version bump (`0.7.422` → `0.7.423`) is now folded into PR1 #10851 alongside the schema migration that requires it — see PR1's commit history.

PR #10849, opened 2026-05-18
feat(tracker): inline filter chips + advanced search + view-option toggles

MichaelUray:
## Summary

Redesign of the Tracker IssuesView toolbar and search pipeline. Pulls the
filter UI out from under the SpaceHeader into an always-visible inline strip,
extends `$search` with prefix-operators + a per-viewlet `searchScope`, adds
an "Add filter" button that stays usable while filters are active, and
unifies the toolbar across List, Kanban and Gantt modes.

This PR is **independent of [#10848](https://github.com/hcengineering/platform/pull/10848)** (fulltext model-version bump),
though that fix is a prerequisite for verifying search end-to-end on a live
workspace.

## What changed

**Filter UI**
- `InlineFilterChips` — chips render directly under the SpaceHeader (or
  inline in Gantt's toolbar row), with a `+N` overflow popover when the
  cluster would otherwise push the trailing toolbar controls off-screen.
- `FilterButton` is always "+ Add filter" once filters exist (clear-all
  moves to a trailing ghost button on the chip strip).
- `FilterBar` keeps its legacy chip-render for the 17 non-Tracker consumers
  (lead / contact / drive / recruit / training / controlled-documents /
  workbench / chunter / task plus Tracker's own MilestoneBrowser /
  ComponentBrowser / IssueTemplatesView). Tracker IssuesView passes
  `hideChips=true` so the chips render via `InlineFilterChips` instead.
- `IModeSelector` gains `disabled?` + `disabledReason?`. When a Status
  filter is active, the All/Active/Backlog ModeSelector greys out with a
  tooltip explaining why; on the rising edge it also resets `mode` to
  `'all'` so the visible mode stops lying about what is filtered.

**Search**
- `SearchInputAdvanced` (new component in packages/ui) replaces the legacy
  SearchInput for Tracker. Prefix shorthands:
  - `title:foo`    → `searchTitle:foo`
  - `id:foo`       → `identifier:foo`
  - `comments:foo` → `comments.message:foo`
- ES adapter routes through `query_string` (with per-field boosts) only
  when a **known** field-name precedes a colon — stray colons in normal
  text (`POC: design review`, `meeting 12:30`, URLs) keep the legacy
  `simple_query_string` path. Bare terms wrapped in a scoped clause are
  Lucene-escaped so the inner colon cannot re-trigger field parsing.
- Customize-View toggles (`showQuickModeSelector`, `searchScope`,
  `searchHighlight`) are shared between the List/Kanban and Gantt viewlets
  via a single `SEARCH_VIEW_OPTIONS` constant.
- `HighlightedText` (new) wraps the matched substring in `<mark>` inside
  Issue identifiers + Gantt sidebar titles. Honours
  `viewOptions.searchHighlight` via the `searchHighlightEnabledStore`.

**Empty-state**
- `SearchEmptyState` card shows when the user has typed a search and zero
  documents matched. ViewletContentView stays mounted underneath so its
  LiveQuery keeps running — earlier iterations unmounted the viewlet on
  zero results and the card self-locked because nothing was around to
  produce a new count. The card visually replaces the viewlet via
  CSS `display:none` on a `display: contents` wrapper.
- All three viewlets (List, Kanban, Gantt) re-arm their query-ready flag
  on every query mutation so a stale count from a previous query never
  leaks into the empty-state condition.

**Gantt-mode toolbar polish (commit `f7a1113577`)**
- Rightmost Fullscreen button + JumpToStart / JumpToEnd icons.
- `dynamicMinPpd` floor based on `BAR_COVERAGE_MIN` so users can't zoom
  out into a void where the bars collapse to a tiny island.
- These three changes happen to live in a single bundled commit. They're
  orthogonal to filter/search; happy to split on request.

## Why it's structured as 12 commits

The branch is organised so each step in the redesign is a self-contained
commit (refactor → feature → feature → ...). Two "fix" commits sit at the
end: one rolls up the chip-render follow-ups that came out of live
verification (sizing, flicker, overflow caps, +N styling), the other
collects the code-review feedback addressed during pre-PR review.

## Test plan

- [x] svelte-check clean across `packages/ui`, `plugins/view-resources`,
      `plugins/tracker-resources`.
- [x] Jest: 19 ui tests (`SearchInputAdvanced.encoder` + `HighlightedText`),
      10 view-resources filter tests (`InlineFilterChips.helpers` +
      `query-builder`), 663 tracker-resources tests.
- [x] Manual UI verification on staging:
  - List view: add a filter via "+ Add filter", verify chip + clear-all
    render below the header, query updates.
  - Kanban view: same.
  - Gantt view: chips render inline in toolbar row, +N overflow popover
    works with ≥4 filters on a 1366×768 viewport.
  - Status-filter active → ModeSelector greys with tooltip, falls back
    to `'all'`.
  - Search `title:foo`, `id:HULY-`, `comments:bar` route as expected;
    `C++ developer` and `meeting 12:30` no longer trip the parser.
  - SearchInputAdvanced clear-button (×) + Escape keybinding work.
  - SearchEmptyState shows on zero hits, dismisses when a new query
    returns results (viewlet stays mounted underneath).
  - Customize-View `searchScope` toggle re-encodes the current search
    immediately (no need to re-type).

## i18n coverage

Translations for all 13 supported locales (en, de, es, fr, it, pt, pt-br,
ru, cs, tr, ja, ko, zh) added for: AddFilter, HiddenFilters,
ModeSelectorDisabledByFilter, ShowQuickModeSelector, SearchScope* (4
labels), SearchHighlight, SearchEmpty* (4 labels), GanttCornerNoOpInSwimlane.

## Backwards compatibility

- `FilterBar` consumers outside Tracker keep the legacy chip render
  (verified by grep — 17 callsites untouched). Tracker IssuesView opts
  out via the new `hideChips` prop.
- `IModeSelector.disabled?` / `disabledReason?` are optional — legacy
  callers (e.g. workbench-resources) compile unchanged.
- `query_string` routing is additive: input without a known field prefix
  still goes through `simple_query_string`. ES adapter has identical
  semantics for the pre-redesign call sites.


MichaelUray:
Closing — this PR's diff was unintentionally scoped against `hcengineering/platform:develop`, which carried ~280 unrelated commits from a parallel Gantt-feature branch on my fork. The filter+search redesign actually *depends* on that Gantt feature, so it can only land upstream after the Gantt stack does.

Plan: reopen this after the 8-PR Gantt stack (currently being prepared) is merged. Apologies for the noise.

MichaelUray:
This work depends on the Gantt feature being upstream (it touches `GanttSidebarColumn.svelte`, `GanttToolbarBar.svelte` slots and Gantt stores). It will be re-opened as a follow-up PR once the Gantt stack ( #10851 → #10859) is merged. Stack overview: #10851.

MichaelUray:
Successor PR opened with the rebased + extended Plan-2 content: [#10872](https://github.com/hcengineering/platform/pull/10872). Same substance plus five rounds of search-encoder hardening, a leading-hyphen escape fix, and a typed-tokenizer refactor. Closing this thread here.

PR #10884, opened 2026-05-25
fix(client): downgrade slow query diagnostics

ignatremizov:
## Summary

- Change slow client `findAll` measurements from `console.error` to `console.warn`.
- Keep the existing slow-query thresholds, log message, payload, and timing details unchanged.
- Clarify that successful-but-slow client queries are performance diagnostics rather than query errors.

## Testing

- `git diff --check upstream/develop...HEAD`
- `git diff --check`
- `node common/scripts/install-run-rush.js build --to @hcengineering/client-resources`

## Notes

- This only changes the browser console severity for slow-query diagnostics. It does not change query execution, thresholds, or diagnostic payloads.


ignatremizov:
It appeared like a lot of errors first time, when it's actually just diagnostic noise. Helpful to add indexes from the diagnostics though, so useful to keep the log line.

PR #10895, opened 2026-05-30
Mention-grants: Collaborator-based cross-space access for @-mentions

MichaelUray:
# Mention-grants: Collaborator-based cross-space access for @-mentions

When a user @-mentions someone on a doc whose class is opted into mentions-grant-access, that mentioned account should gain read access to the doc — even if they are not a member of the doc's containing space. This is the cross-space access path that today's Collaborator infrastructure already supports for explicit `addCollaborator` calls; this PR wires it up to the @-mention authoring flow with an opt-in disclosure dialog.

**Companion docs:** [hcengineering/huly-docs#72](https://github.com/hcengineering/huly-docs/pull/72) (draft; text complete, screenshots will be added in a follow-up commit on the same branch).

> **Scope:** `tracker.class.Issue` is the only class that opts in today via `mentionsGrantAccess: true`. The architecture is generic (any `ClassCollaborators` provider can opt in), but rolling it out to other classes is an explicit follow-up decision per class.

## What this builds

**V1 — Cross-space Space-visibility for Guests.** Guests who hold a Collaborator edge to a doc inside a space they are not a member of need that space to surface in their navigator so the doc is reachable without deep-link gymnastics. Three layers cooperate:

1. **`foundations/server/packages/middleware/src/spaceSecurity.ts`** — skip the middleware-level space pre-filter for `Guest` / `ReadOnlyGuest` reads of Space-derived classes (`spaceCollabBypass`). The Postgres adapter then does the filtering at SQL level.
2. **`foundations/server/packages/postgres/src/storage.ts`** — `addSecurity` appends an `OR EXISTS (collaborator WHERE space = domain._id AND collaborator = caller)` branch for Guest reads on `DOMAIN_SPACE`. A Space hosting any Collaborator record naming the caller becomes readable. Self-Collaborator visibility for non-Admin non-System callers is added in the same change.
3. **`plugins/workbench-resources/src/components/Navigator.svelte`** — client-side: union of `member-spaces` + `collab-spaces`. The `collab-spaces` set comes from `findAll(Collaborator, { collaborator: self })` with `projection: { space: 1 }`, then a batched `findAll(Space, { _id: { $in: ... } })`. Tracker's per-project tree marks collab-only projects with a dim icon + tooltip ("Shared with you") and hides non-Issues sub-views (Components / Milestones / Templates) because they would render empty for a collab-only Guest.

**V3 — Send-time disclosure for @-mentions.** When the author types `@person`, the picker passes the user-intent into a Markup attribute on the reference node (`grantsAccess`, values: `'true' | 'false' | undefined`). On send, if the message contains any reference with `grantsAccess !== 'false'` and the mentioned employee is not already a member of the doc's space, a disclosure dialog asks the author to confirm per grantee. (The server trigger additionally dedups against existing Collaborator records, so the dialog may surface for a user who already has the grant; the trigger then emits no new Tx.) Cancelling the dialog preserves the unsent comment.

**V3c (server filter)** — `mentionsGrantAccess` trigger ignores any reference whose `grantsAccess` is explicitly `'false'`. **V3d (edit path)** — re-grants on a message edit are strictly add-only: a removed mention does not retract anyone's read access (a previously-granted user remains a Collaborator until an explicit `removeDoc` on the Collaborator).

## Architecture sketch

```
Author types @person          Reference node carries grantsAccess attr
        │                              │
        ▼                              ▼
Comment composer (V3a)        Server trigger applyMentionGrants (V3c/d)
        │                              │
        ▼                              ▼
 Disclosure dialog (V3b)       TxCreateDoc Collaborator (add-only)
        │                              │
        ▼                              ▼
   Send / Cancel              Guest: postgres OR-branch + Navigator (V1)
                                      │
                                      ▼
                              Doc + space visible cross-space
```

## What is opt-in

- Per class via `ClassCollaborators.mentionsGrantAccess: true` (default `false`).
- Today only `tracker.class.Issue` opts in.

## Tests

| Suite | Result |
|-------|--------|
| `foundations/core/packages/core` | 982 / 982 |
| `foundations/core/packages/text-core` | 19 / 19 (new `reference.test.ts`) |
| `plugins/chunter-resources` | 5 / 5 (new `mentionGrants.test.ts`) |
| `server-plugins/chunter-resources` | 6 / 6 (new `mentionGrantsTrigger.test.ts`) |
| `foundations/server/packages/middleware` | 30 / 30 |

`rush install` / `rush build` / `rush validate`: all exit 0. `svelte-check` across `tracker-resources` / `chunter-resources` / `workbench-resources`: 0 errors.

## Live deployment evidence

Deployed and verified on a self-hosted Huly v0.7.423 instance with the CockroachDB backend. Specifically exercised:

- Guest user with Collaborator records on Issues across two projects (one project where they are also a member, one where they are collab-only) — nav surfaces both, with the collab-only one dimmed and sub-nodes hidden; mention picker correctly lists their collab-targets.
- Postgres OR-branch verified at SQL level (`EXISTS` query against `collaborator` table returns the expected projects).
- V3b dialog opens on send when the message contains an unprivileged-mention; Cancel returns to the composer with the message intact.

## Sign-off

Single author, signed-off-by on every commit, DCO-compliant. 28 commits, intentionally not squashed for traceability. Squash-merge fine if preferred.

### DIFF
--- a/foundations/core/packages/core/src/collaborators.ts
+++ b/foundations/core/packages/core/src/collaborators.ts
@@ -13,7 +13,16 @@
 // limitations under the License.
 //
 
-import core, { Class, ClassCollaborators, Doc, Hierarchy, ModelDb, Ref } from '.'
+import core, {
+  AttachedDoc,
+  Class,
+  ClassCollaborators,
+  Doc,
+  DocumentQuery,
+  Hierarchy,
+  ModelDb,
+  Ref
+} from '.'
 
 export function getClassCollaborators<T extends Doc> (
   model: ModelDb,
@@ -34,4 +43,50 @@ export function getClassCollaborators<T extends Doc> (
       return res
     }
   }
+}
+
+/**
+ * Walk a Doc's attachedTo chain to find the nearest ancestor (including the
+ * Doc itself) whose ClassCollaborators has BOTH provideSecurity===true AND
+ * mentionsGrantAccess===true. Returns that ancestor Doc as the grant target,
+ * or null if no such class is reached within the depth cap.
+ *
+ * Used by both the chunter mention-trigger (server) and the warning popup
+ * (client) so the disclosure UX matches the actual server-side grant. The
+ * helper is isomorphic via the findAll dependency injection — server passes
+ * `(cls, q) => control.findAll(control.ctx, cls, q)`, client passes
+ * `(cls, q) => getClient().findAll(cls, q)`.
+ *
+ * The ClassCollaborators lookup is exact-class (not inherited via ancestors)
+ * — adequate for tracker.class.Issue and avoids surprising matches on
+ * abstract base classes like AttachedDoc. If future opt-in classes need
+ * inherited semantics, switch to `getClassCollaborators(model, hierarchy, _class)`
+ * here (requires plumbing ModelDb + Hierarchy through the dependency
+ * injection — kept out for now to keep the helper isomorphic without
+ * the ModelDb tax on the client).
+ *
+ * Depth cap (8) defends against pathological attachedTo cycles.
+ */
+export async function resolveMentionGrantTarget (
+  start: Doc,
+  findAll: <T extends Doc>(cls: Ref<Class<T>>, q: DocumentQuery<T>) => Promise<T[]>
+): Promise<Doc | null> {
+  let cur: Doc | undefined = start
+  for (let i = 0; i < 8 && cur != null; i++) {
+    const cc = (await findAll(core.class.ClassCollaborators, {
+      attachedTo: cur._class
+    } as DocumentQuery<ClassCollaborators<Doc>>))[0]
+    if (cc?.provideSecurity === true && cc.mentionsGrantAccess === true) {
+      return cur
+    }
+    const attached = cur as AttachedDoc
+    if (attached.attachedTo == null || attached.attachedToClass == null) {
+      return null
+    }
+    const parent = (await findAll(attached.attachedToClass, {
+      _id: attached.attachedTo
+    } as DocumentQuery<Doc>))[0]
+    cur = parent
+  }
+  return null
 }
--- a/foundations/core/packages/core/src/collaborators.ts
+++ b/foundations/core/packages/core/src/collaborators.ts
@@ -13,7 +13,17 @@
 // limitations under the License.
 //
 
-import core, { AttachedDoc, Class, ClassCollaborators, Doc, DocumentQuery, Hierarchy, ModelDb, Ref } from '.'
+import core, {
+  AttachedDoc,
+  Class,
+  ClassCollaborators,
+  Doc,
+  DocumentQuery,
+  FindOptions,
+  Hierarchy,
+  ModelDb,
+  Ref
+} from '.'
 
 export function getClassCollaborators<T extends Doc> (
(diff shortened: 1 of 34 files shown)


foundations/core/packages/core/src/collaborators.ts
```diff
@@ -35,3 +35,51 @@ export function getClassCollaborators<T extends Doc> (
...
+export async function resolveMentionGrantTarget (
+  start: Doc,
+  findAll: <T extends Doc>(cls: Ref<Class<T>>, q: DocumentQuery<T>) => Promise<T[]>
+): Promise<Doc | null> {
+  let cur: Doc | undefined = start
+  for (let i = 0; i < 8 && cur != null; i++) {
+    const ccQuery: DocumentQuery<ClassCollaborators<Doc>> = {
+      attachedTo: cur._class
+    }
+    const cc = (await findAll(core.class.ClassCollaborators, ccQuery))[0]
```
ArtyomSavchenko: I think it is better to add `limit: 1` if only one item is required.

server-plugins/chunter-resources/src/index.ts
```diff
@@ -208,25 +211,62 @@ async function OnChatMessageCreated (ctx: MeasureContext, tx: TxCUD<Doc>, contro
...
+      await control.findAll<Collaborator>(control.ctx, core.class.Collaborator, {
+        attachedTo: grantTarget._id
+      })
+    ).map((c) => c.collaborator)
+
+    for (const collab of collaboratorsFromMessage) {
+      if (grantCollabs.includes(collab)) {
+        continue
+      }
+      res.push(
```
ArtyomSavchenko, line 259: I think when an issue has no collaborators yet, two code paths both create Collaborator records without seeing each other's uncommitted transactions, so the same person can be added twice.

The "seed collaborators" block runs because currentCollaborators is empty, pushing getAddCollaboratTxes for the issue's auto-collaborators (createdBy / assignee) and updating the local currentCollaborators list.
The new grant branch then dedups against grantCollabs — a fresh DB query that is still empty, because the transactions from step 1 aren't committed yet.
Because the grant branch ignores the currentCollaborators list that step 1 just populated, any person present in both sets gets a duplicate record.

foundations/core/packages/core/src/collaborators.ts
MichaelUray: Applied. Both `[0]` lookups in `resolveMentionGrantTarget` now use `limit: 1`, with `FindOptions` forwarded through all three callers. I also bounded the equivalent `targetClassCollab` lookup in the trigger.

server-plugins/chunter-resources/src/index.ts
MichaelUray: The grant branch only considered committed Collaborator records. I traced the seed path and found that it currently queues records on `tx.objectId` — the ChatMessage — rather than `targetDoc._id`, so the exact issue-level duplicate described here does not occur in the current path. This looks like a long-standing pre-existing mismatch in that seed path.

The missing in-flight dedup still needs to be handled: the grant path now includes matching queued Collaborator create transactions on `grantTarget._id` in its dedup basis. The set is also updated in-loop, which fixes the current self-mention duplicate. The seed-target mismatch remains outside this PR.

PR #10969, opened 2026-07-09
Fix filtered view visibility

ArtyomSavchenko:


aonnikov (approved):

plugins/view-resources/src/components/filter/FilterBar.svelte
```diff
@@ -37,6 +37,9 @@
   const hierarchy = client.getHierarchy()
   const dispatch = createEventDispatcher()
 
+  const account = getCurrentAccount()
+  const canSaveFilteredView = account.role !== AccountRole.ReadOnlyGuest && account.role !== AccountRole.DocGuest
```
aonnikov, line 41: Can it actually be a DocGuest? Do we use FilterBar in documents?

aonnikov: Maybe we need to refactor this approach later. I don't like that we do this in individual components. If we add a new role, we have to review so many places. Maybe a global store / state would work ...

ArtyomSavchenko: I will refactor it in a separate PR

PR #10976, opened 2026-07-14
feat: use hot prowser instance to improve stability

aonnikov:


services/print/pod-print/src/browser.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  async createPage (): Promise<PrintBrowserPage> {
+    const browser = await this.getBrowser()
+
+    try {
+      return await this.createPageInBrowser(browser)
+    } catch (err) {
+      if (this.browser === browser) {
+        await this.resetBrowser(browser)
```
ArtyomSavchenko, line 100: Is it safe in case when PRINT_CONCURRENCY > 1?

services/print/pod-print/src/browser.ts
```diff
@@ -0,0 +1,153 @@
+//
+// Copyright © 2026 Hardcore Engineering Inc.
+//
+
+import puppeteer, { type Browser, type BrowserContext, type Page } from 'puppeteer'
+
+import config from './config'
+
+const BROWSER_PAGE_INIT_TIMEOUT_MS = 2000
```
ArtyomSavchenko, line 9: Should we make this timeout configurable?

PR #10992, opened 2026-07-24
feat(tracker): Gantt chart for the issue tracker

MichaelUray:
The search/filter rework this built on has merged (#10998), so this is now a standalone Gantt PR on current `develop`.

### What it adds

- **Read-only viewlet:** time-scale canvas with day/week/month/quarter zoom, sidebar columns, today marker and jump-to-start/end/date navigation.
- **Editable bars:** drag to move, resize handles for start/due, drag an unscheduled row to schedule it, with optional confirm prompts and permission/conflict handling on commit.
- **Dependencies:** FS/SS/FF/SF relations with lag, a dependency editor and arrows, cascade shifting with cycle detection, and critical-path plus slack.
- **Visual polish & coloring:** bar labels, quick-info popover, milestone and deadline markers, non-working-day shading, shortcut help; bar coloring by status/priority/assignee/component/milestone with overdue and blocked overlays and sub-issue progress fill.
- **Tiers:** undo/redo, saved views, bulk select and bulk drag, auto-scheduling with manual pinning; row virtualization for large projects; a read-only mobile layout, tree view with hierarchy breadcrumbs and a predecessor column.
- **Export & fullscreen:** PNG and PDF export plus a fullscreen toggle, confined to the Gantt toolbar.
- **Dependency-shift notification:** after a dependency cascade commits, a server-side trigger notifies affected issue collaborators, falling back to the assignee when no Collaborator records exist. It derives the sender, issue metadata, recipient spaces, and same-project issue set server-side rather than trusting client-supplied notification data.

### The engine package

Introduces `@hcengineering/gantt`, a runtime-dependency-free package for the domain-neutral part of the Gantt library — time-scale and working-day calculations, zoom and viewport primitives, drag/pointer state, responsive toolbar helpers, and the neutral `GanttItem`, `GanttDependency`, and `WorkingCalendar` contracts. Tracker-specific scheduling, layout, critical-path and dependency-routing logic remains in the Tracker adapter; the Svelte UI, model schema, Huly actions, notifications, persistence and sanity coverage remain in their corresponding Tracker packages.

### Data model and migrations

Adds optional scheduling fields to Tracker documents, including issue deadlines and scheduling mode, project working-day configuration, and component/milestone colors. It also registers the dependency-shift request and notification documents and includes idempotent upgrades for Gantt view options and legacy relation-activity records. Existing issues and projects retain their previous behaviour because the new scheduling fields are optional.

### New runtime dependencies

In `plugins/tracker-resources`, both MIT: `@tanstack/svelte-virtual` (row virtualization) and `jspdf` (PDF export). Export re-renders the chart from the data model into an SVG rather than capturing the DOM; the PDF is paginated into bounded pages so large charts stay within the browser canvas and jsPDF size limits.

### Locales

`en`, `de` and `ru` carry the full set of new strings, and a unit test enforces ICU placeholder parity across all locale files. Other locales fall back to English at runtime.

### Testing

Full CI is green on a fork mirror of this exact head (`02933ee0f0`) — `build`, `svelte-check`, unit `test`, `formatting`, `docker-build`, `uitest`, `uitest-pg`, `uitest-workspaces`, `uitest-qms`. Fork-CI run: https://github.com/MichaelUray/huly-platform/actions/runs/31542585071

One small cross-cutting change: GanttView is a third writer to the shared result-count store from the search rework, so it uses the owner-token gate (claim/release like List and Kanban); the gate now resets to the pending sentinel on claim, so switching viewlets never shows a stale count.


ArtyomSavchenko:
Hi @MichaelUray 
Thank you for your contribution.
I have the following suggestions:
**1. Split out the search/filter rework into its own PR**
The one thing that stands out as not Gantt infrastructure is the search/filter overhaul:

packages/ui: SearchInputAdvanced (field-prefixed field:value search), HighlightedText
foundations/server/packages/elastic: the new query_string branch
plugins/view-resources: inline filter chips + overflow popover, query-builder, empty state

**2. A separate package for Gantt engine**
I think that Gantt engine(gantt/lib) is a good candidate to live as its own reusable package, with the tracker integration staying in tracker-resources. I think we could then reuse it elsewhere — for cards, for example.

MichaelUray:
Understood. I'll split the shared search/filter rework into a separate PR.

For the Gantt code, I'll extract a tracker-agnostic engine package with generic item, dependency, scheduling, layout and viewport interfaces. Tracker-specific adapters, Huly actions, notifications and the Svelte integration will remain in `tracker-resources`.

I'll move this PR back to Draft while restructuring it and update it once both boundaries build and test independently.

packages/gantt/src/confirm-gate.ts
```diff
@@ -0,0 +1,44 @@
...
+ * The pointer handlers consult {@link isConfirming} to short-circuit.
+ *
+ * The flag is module-scope (not store-based) on purpose — there is only
+ * ever one drag in flight, the consumer is GanttView, and the flag is
+ * always set/cleared inside the same async commit path. A Svelte store
+ * would force a reactive cycle through `$:` blocks for a guard that
+ * needs to be read synchronously inside event handlers.
+ */
+
+let confirming = false
```
ArtyomSavchenko: Looks like the drag-confirmation gate is a module-level singleton shared across every mounted GanttView, not scoped per view. If two Gantt views are ever mounted simultaneously (e.g. two tabs), confirming a drag in one silently blocks drag input in the other.

plugins/tracker/src/index.ts
```diff
@@ -55,6 +57,36 @@ export * from './analytics'
...
+   *   bit 0 = Mon, bit 1 = Tue, …, bit 5 = Sat, bit 6 = Sun.
+   *
+   *   Mon–Fri   = 0b0011111 = 31
+   *   Mon–Sat   = 0b0111111 = 63
+   *   All days  = 0b1111111 = 127
+   */
+  weekdayMask: number
+
+  /** Holidays as UTC-midnight timestamps. Order is irrelevant; duplicates allowed. */
+  holidays: Timestamp[]
```
ArtyomSavchenko: Looks like the new per-project WorkingDaysConfig.holidays duplicates the existing workspace-level HR holiday concept (hr.class.PublicHoliday) with no integration, so holidays must be re-entered per project and can drift from the real HR calendar, skewing cascade/critical-path dates.

plugins/tracker-resources/src/components/gantt/lib/sidebar-sort.ts
```diff
@@ -0,0 +1,133 @@
...
+  if (col === 'modifiedOn') return issue.modifiedOn ?? 0
+  if (col === 'createdOn') return issue.createdOn ?? 0
+  return 0
+}
+
+function readDate (issue: Issue, col: SidebarColumnKey): number | null {
+  if (col === 'startDate') return issue.startDate ?? null
+  if (col === 'dueDate') return issue.dueDate ?? null
+  if (col === 'deadline') {
+    // No native `deadline` field on Issue — preserve the
```
ArtyomSavchenko: Is this comment actual?
Looks like deadline field was added in this PR

packages/gantt/src/confirm-gate.ts
MichaelUray: Correct — the gate was a module-level flag shared by every mounted GanttView. Replaced it with a `createConfirmGate()` factory returning a per-instance `{ setConfirming, isConfirming }` closure; GanttView creates one instance at mount, so two simultaneously-mounted Gantt views no longer share drag-confirmation state. Added a two-instance independence test.

plugins/tracker/src/index.ts
MichaelUray: Agreed — the per-project holidays array duplicated `hr.class.PublicHoliday` and could drift. I removed it. `WorkingDaysConfig` now carries only the weekday mask and an optional `holidayDepartment`. The tracker adapter resolves holidays for the selected department plus all ancestors up to `hr.ids.Head`, matching HR's existing inheritance semantics. Missing or invalid references fall back to company-wide holidays; unrelated departments are never included.

A small `WorkingDaysEditor` in the existing project create/edit dialog lets users with the existing project permissions enable the mode, choose weekdays, and select an HR department. It persists through the existing project create/update transaction paths and uses `$unset` when disabled.

Calendar loading is explicit: project-scoped drag, resize, cascade, keyboard-shift, and auto-scheduling mutations remain disabled until configuration and holiday data have loaded. A generation ticket is checked before every calendar-dependent date commit, preventing work started in one project from being persisted after a project switch. The all-projects view remains display-only for calendar-dependent edits.

This depends on the lightweight `@hcengineering/hr` model package, but not on `hr-resources`; actual HR model presence is checked through the hierarchy at runtime. Without the HR model, holidays resolve to an empty set. `@hcengineering/gantt` remains domain-neutral.

plugins/tracker-resources/src/components/gantt/lib/sidebar-sort.ts
MichaelUray: Correct, `deadline` was added to the Issue model in this PR, so the comment was stale. Removed the cast and read `issue.deadline` directly.

plugins/tracker-resources/src/components/gantt/lib/scheduler.ts
```diff
@@ -0,0 +1,421 @@
...
+        // touch. The snap term still wins when the
+        // constraint itself requires a larger jump (e.g. lag changed).
+        //
+        // Other kinds don't need this: SS already anchors on cur.start,
+        // so its required anchor naturally moves with curStartDelta. FF
+        // and SF anchor on the cur side that may not match the dragged
+        // side, so gap preservation isn't a clean concept there — pure
+        // snap is the intended semantics.
+        const snap = requiredAnchor
+        const newAnchor = r.kind === 'finish-to-start' ? Math.max(snap, targetAnchor + curStartDelta) : snap
```
ArtyomSavchenko: Looks like FS gap preservation adds curStartDelta directly as milliseconds. With a Mon–Fri calendar, moving a predecessor Friday → Monday yields a 3-calendar-day delta, so its successor is pushed three calendar days (e.g. Monday → Thursday) instead of one working day. Use working-day arithmetic for the preserved delta, or omit this raw-ms floor when a working calendar is active.

MichaelUray: Fixed in 634532f556. The FS gap-preservation floor advanced the successor by the predecessor's movement in raw calendar milliseconds. With a working calendar, that over-shifted across non-working days: a Friday-to-Monday move is three calendar days but one working day, and the raw floor could land a shifted successor on Saturday.

When a working calendar is active, the floor now advances the successor by the predecessor's due-side movement measured in working days, using the new workingDayDelta helper in @hcengineering/gantt. The existing snap term was already calendar-aware; only the gap-preservation floor used raw milliseconds. In the affected forward-shift cases, the floor now follows the configured workdays and holidays.

The gate is day-granular, so a right resize with an unchanged start day and a left resize with an unchanged due date remain pure-snap operations without a spurious floor push. Without a working calendar, the existing raw-millisecond behavior is unchanged. Added engine tests for workingDayDelta plus cascade coverage for weekend-spanning movement, a floor-dominant multi-hop chain, holidays, negative movement, time-of-day inputs, and both resize directions.

PR #10996, opened 2026-07-26
fix(github): recover write access when user OAuth token expires

### LINKED ISSUE #10994: Bug: expired user OAuth token permanently breaks GitHub writes — getOctokit returns a dead-token Octokit instead of undefined, and revokeUserAuth is a structural no-op
### Description of the issue

When a user's GitHub OAuth token becomes invalid, the GitHub integration
enters a permanent write-failure state that it can never recover from
automatically. Every Huly → GitHub write (issue creation, PR update,
comment sync) fails with `401 Bad credentials` indefinitely.

Three defects in `services/github/pod-github` combine to cause this. Each
one individually disables a recovery mechanism that the code already has.

All three are still present on `develop` (verified at commit c36dd74f).

---

#### Defect 1 — `getOctokit` returns a dead-token Octokit, defeating the installation-token fallback

`services/github/pod-github/src/worker.ts:642-661`

```ts
if (record !== undefined) {
  ctx.info('get octokit', { ... })
  if (!(await this.platform.checkRefreshToken(ctx, record))) {
    record.octokit = undefined          // <-- only clears the cache
  }
  if (record.octokit !== undefined) {
    return record.octokit
  }

  record.octokit = ctx.withSync('create-octokit', {}, () =>
    new Octokit({ auth: record.token, ... })   // <-- rebuilt with the SAME dead token
  )
  return record.octokit
}
```

When `checkRefreshToken` returns `false` (refresh failed, token revoked),
line 645 clears only the cached instance. Control then falls straight
through to line 651, which constructs a new `Octokit` from the very
`record.token` that was just determined to be unusable.

The consequence is that `getOctokit` never returns `undefined` as long as
a user record exists. Every call site is written to expect exactly that:

`services/github/pod-github/src/sync/issues.ts:753-756`

```ts
const okit = ensureGraphQLOctokit(
  (await this.provider.getOctokit(ctx, existingIssue.modifiedBy)) ?? container.container.octokit,
  container
)
```

The `?? container.container.octokit` fallback to the app installation
Octokit (built at `platform.ts:802` via `getInstallationOctokit`) is
correct and would make the write succeed — but it is unreachable, because
the left operand is never nullish. The same pattern appears in
`comments.ts:156,365,441`, `issueBase.ts:343`, `reviewComments.ts:159,442,520`,
`reviewThreads.ts:383,461`, `reviews.ts:154,439`, `issues.ts:754,803`.

`ensureGraphQLOctokit` (`sync/utils.ts:46-51`) does not catch this either —
it checks `typeof okit.graphql === 'function'`, i.e. the *shape* of the
object. A dead-token Octokit is structurally perfect; only its
credentials are invalid.

**Expected:** `getOctokit` returns `undefined` when the token cannot be
validated or refreshed, so callers fall back to the installation token.

---

#### Defect 2 — `revokeUserAuth` is a structural no-op

`services/github/pod-github/src/platform.ts:1411-1421`

```ts
public async revokeUserAuth (ctx: MeasureContext, record: GithubUserRecord): Promise<void> {
  for (const [ws, acc] of Object.entries(record.accounts)) {
    await this.updateAccountAuthRecord(ctx, { workspace: ws as WorkspaceUuid, accountId: acc },
      { login: record._id }, undefined, true)
  }
}
```

`record.accounts` is always `{}`, because the only function that
materialises a `GithubUserRecord` from storage overwrites it:

`services/github/pod-github/src/users.ts:38-45`

```ts
private secretToUserRecord (secret: IntegrationSecret, login: string): GithubUserRecord | undefined {
  return {
    ...(JSON.parse(secret.secret) ?? {}),
    account: secret.socialId,
    _id: login,
    accounts: {}        // <-- clobbers the persisted value (spread comes first)
  }
}
```

The field is genuinely persisted — it is populated during the OAuth
callback at `platform.ts:488` and `platform.ts:495`, and serialised by
`updateUser` via `JSON.stringify(dta)` (`users.ts:134`, `users.ts:142`).
The read path then discards it. Both `getAccount` (`users.ts:28`) and
`getAccountByRef` (`users.ts:84`) go through `secretToUserRecord`, so no
caller ever observes a populated `accounts`.

Because the loop body never executes, `revokeUserAuth` cleans up nothing,
and the `AuthenticatedWithGithubRequired` re-auth notification is never
delivered. The user is not told their integration is broken. All three
call sites are affected: `platform.ts:721`, `platform.ts:761`,
`platform.ts:1198`, and `worker.ts:597`.

**Expected:** `accounts` survives a persist/load round-trip (drop the
`accounts: {}` line, or default it with `accounts: parsed.accounts ?? {}`),
so revocation and the re-auth notification actually fire.

---

#### Defect 3 — `checkRefreshToken`'s `force` parameter is never read

`services/github/pod-github/src/platform.ts:740`

```ts
async checkRefreshToken (ctx: MeasureContext, auth: GithubUserRecord, force: boolean = false): Promise<boolean> {
  if (auth.refreshToken != null && auth.expiresIn != null && auth.expiresIn < Date.now() / 1000) {
    ...
  }
  return true      // line 786
}
```

`grep -n force services/github/pod-github/src/platform.ts` returns exactly
one hit: the signature on line 740. The parameter is not referenced in the
body.

This makes the deliberate forced-refresh call a no-op:

`services/github/pod-github/src/worker.ts:591`

```ts
await this.platform.checkRefreshToken(ctx, record, true)
```

Related, and arguably the deeper root cause: line 786 returns `true`
unconditionally whenever the line 741 guard does not match. If
`refreshToken` or `expiresIn` is `null`, the token's validity is never
checked at all — the function reports success without contacting GitHub.
Combined with Defect 1, a record in that state yields a confidently
returned, permanently broken Octokit.

**Expected:** `force === true` bypasses the expiry check and performs a
refresh (or at least a validation call) regardless of `expiresIn`.

---

### Your environment

* Version of Huly: self-hosted `hardcoreeng/github:v0.7.426` (Kubernetes).
  Source verified unchanged on `develop` @ `c36dd74f` and on tag `v0.7.432`.
* Browser: not applicable — server-side integration service
* OS: Linux x86_64, Node.js 22 (per the service container)

### Steps to reproduce

1. Connect a GitHub account to a workspace via user OAuth and link a repository.
2. Cause the user OAuth token to become invalid — e.g. let it expire, or
   revoke the authorization from GitHub's *Applications* settings page.
3. In Huly, create or edit an issue in a project synced to that repository,
   attributed to that user (`modifiedBy`).
4. Observe the `github` service log.

### Actual behaviour

Every write attempt fails with `401 Bad credentials` and never recovers.
The `createIssue` GraphQL mutation is sent with the dead user token instead
of falling back to the installation token:

```json
{"level":"error","message":"failed to sync doc","objectClass":"tracker:class:Issue",
 "error":{"name":"HttpError","status":401,
          "response":{"data":{"message":"Bad credentials","status":"401"}},
          "request":{"method":"POST","url":"https://api.github.com/graphql",
                     "headers":{"authorization":"token [REDACTED]"}}}}
```

The service log also confirms the code path — `get octokit` (worker.ts:643,
the "record found" branch) is reached, so the dead-token Octokit is
constructed and returned rather than the `get octokit: return bot` fallback:

```json
{"level":"info","message":"get octokit"}
```

No `revokeUserAuth` effect and no `AuthenticatedWithGithubRequired`
notification is ever produced, so from the user's perspective the
integration silently stops writing. Re-authorizing manually is the only
way out, and nothing in the product indicates that this is required.

### Expected behaviour

1. A user token that cannot be validated or refreshed causes `getOctokit`
   to return `undefined`, so writes fall back to the app installation
   token and succeed.
2. `revokeUserAuth` clears the stale authentication record and notifies
   the user that re-authentication is required.
3. `checkRefreshToken(ctx, auth, true)` actually forces a refresh.

### Possible solutions

1. **worker.ts:644** — return early instead of only clearing the cache:

   ```ts
   if (!(await this.platform.checkRefreshToken(ctx, record))) {
     record.octokit = undefined
     return undefined
   }
   ```

2. **users.ts:43** — stop discarding the persisted value:

   ```ts
   const parsed = JSON.parse(secret.secret) ?? {}
   return { ...parsed, account: secret.socialId, _id: login, accounts: parsed.accounts ?? {} }
   ```

3. **platform.ts:741** — honour `force`, and treat a missing `expiresIn`
   as "unknown" rather than "valid":

   ```ts
   const expired = auth.expiresIn != null && auth.expiresIn < Date.now() / 1000
   if (auth.refreshToken != null && (force || expired)) { ... }
   ```

I am happy to open a PR against `develop` for any or all of these if the
approach looks right.

dnplkndll:
Fixes #10994

## What changed

When a user's GitHub OAuth token became invalid, the integration entered a permanent write-failure state (`401 Bad credentials` on every Huly → GitHub write). Three small defects in `services/github/pod-github` each disabled a recovery mechanism the code already has:

1. **`worker.ts` — `getOctokit` returned a dead-token Octokit.** After `checkRefreshToken` returned `false`, the code cleared the cached instance but then rebuilt a new `Octokit` from the same invalid `record.token`. It now returns `undefined` in that case, so the existing `?? container.container.octokit` installation-token fallback at every call site becomes reachable and writes succeed.

2. **`users.ts` — `secretToUserRecord` clobbered the persisted `accounts` map with `{}`** (the spread came first). That made `revokeUserAuth` iterate an empty object — a structural no-op — so cleanup never ran and the `AuthenticatedWithGithubRequired` re-auth notification never reached the user. The parsed value is now preserved (`parsed.accounts ?? {}`), matching what the OAuth callback persists.

3. **`platform.ts` — `checkRefreshToken` never read its `force` parameter**, so the deliberate forced refresh in `syncUsers` (`worker.ts`) did nothing. `force === true` now triggers a refresh regardless of expiry when a refresh token is present.

## Verification

- `rush validate --to pod-github` — clean
- `rush format --to pod-github` — no changes
- `rushx test` in pod-github — 4 suites, 77 tests passed

https://claude.ai/code/session_01ANdoXbdn5k2hZy734EwKe7

### DIFF
--- a/services/github/pod-github/src/platform.ts
+++ b/services/github/pod-github/src/platform.ts
@@ -738,7 +738,7 @@ export class PlatformWorker {
   }
 
   async checkRefreshToken (ctx: MeasureContext, auth: GithubUserRecord, force: boolean = false): Promise<boolean> {
-    if (auth.refreshToken != null && auth.expiresIn != null && auth.expiresIn < Date.now() / 1000) {
+    if (auth.refreshToken != null && (force || (auth.expiresIn != null && auth.expiresIn < Date.now() / 1000))) {
       const uri =
         'https://github.com/login/oauth/access_token?' +
         makeQuery({
--- a/services/github/pod-github/src/users.ts
+++ b/services/github/pod-github/src/users.ts
@@ -36,11 +36,12 @@ export class UserManager {
   }
 
   private secretToUserRecord (secret: IntegrationSecret, login: string): GithubUserRecord | undefined {
+    const parsed = JSON.parse(secret.secret) ?? {}
     return {
-      ...(JSON.parse(secret.secret) ?? {}), // TODO: Add security
+      ...parsed, // TODO: Add security
       account: secret.socialId,
       _id: login,
-      accounts: {}
+      accounts: parsed.accounts ?? {}
     }
   }
 
--- a/services/github/pod-github/src/worker.ts
+++ b/services/github/pod-github/src/worker.ts
@@ -642,7 +642,11 @@ export class GithubWorker implements IntegrationManager {
     if (record !== undefined) {
       ctx.info('get octokit', { account, recordId: record._id, workspace: this.workspace.uuid })
       if (!(await this.platform.checkRefreshToken(ctx, record))) {
+        // Token is invalid and could not be refreshed: report no octokit so
+        // callers fall back to the installation token instead of using a
+        // client with dead credentials.
         record.octokit = undefined
+        return undefined
       }
       if (record.octokit !== undefined) {
         return record.octokit


ArtyomSavchenko:
Hi @dnplkndll
Thanks for your contibution.
However, a similar fix had already been merged: https://github.com/hcengineering/platform/pull/10995/changes.

PR #10872, opened 2026-05-21
feat(tracker): Gantt — #9 inline filter chips + advanced search + view-option toggles

MichaelUray:
## Summary

Successor to the closed [#10849](https://github.com/hcengineering/platform/pull/10849). Same Plan-2 substance (inline filter chips + advanced search bar + view-option toggles for the Tracker viewlets) plus several rounds of encoder hardening based on review feedback after #10849 was closed. Includes the follow-up bugfix round from live testing on our staging deployment.

The 14 commits in this PR break into three groups:

1. **Plan-2 core (7 commits)** — the same content as #10849: All/Active/Backlog inline filter chips, advanced search with field-prefix support (`title:`, `id:`, `comments:`), Customize-View popover with `Search in...` scope + highlight-match + quick-filter toggles, empty-state card with reset actions, match-highlight in list and Gantt sidebar.
2. **Search-encoder hardening (5 commits)** — found while extending Plan-2 to a Gantt viewlet. Each fixes a real Elasticsearch `query_string` parse failure that surfaced as silent zero-hit results:
    - `2461879b4b` — fixes three independent issues: (a) IssuesView search state leaking across route switches, (b) prefix-targeted values with Lucene-reserved chars (`title:C++`, `comments:foo/bar`) crashing the parser, (c) Saved Gantt View reload being a no-op when re-selecting the active view. **Includes a 12-line saved-view-reload fix in `GanttView.svelte`** that was discovered while testing the Plan-2 search-state leak; bundled here for atomicity rather than carried in a follow-up PR.
    - `19024b8a65` — colon-in-value escape (`title:POC:123`, `title:12:30`) so the inner colon doesn't re-enter field-targeted parsing.
    - `ff1bf76105` — broaden orphan-token escape from `:` only to the full reserved set, plus quoted-phrase-aware tokenizer.
    - `56326b81ef` — capture attached parens/brackets as part of the bare value (`title:foo(bar)`, `title:list[0]`).
    - `28c9712e01` — refactor: replace the 2-pass regex encoder with a single typed-tokenizer + per-type renderer. Each token kind has exactly one rendering rule; future edge cases need only either a new token kind or a tweak to the relevant renderer, not a cross-cutting regex re-think. Behavior preserved bit-for-bit (all 30 existing encoder tests pass without modification).
3. **Final correctness + cleanup (2 commits)** —
[...]

PR #10998, opened 2026-07-27
feat(tracker): field-prefixed search, inline filter chips and zero-hit empty state

MichaelUray:
## Summary

Split out of #10992 following review feedback. This supersedes the closed #10872 with a clean branch based on current `develop` and contains only the shared search/filter work; no Gantt code.

Adds `SearchInputAdvanced`, match highlighting, inline filter chips with an overflow popover, a reusable zero-hit empty state in `view-resources`, and search-scope/highlight view options wired into List and Kanban.

Known field prefixes are routed through Elasticsearch `query_string`; queries without a recognized prefix retain the existing `simple_query_string` path. Workspace and space access filtering remain unchanged.

The exact head passed all required fork-CI jobs, including `uitest`, `uitest-pg`, `uitest-workspaces`, `uitest-qms`, `build`, `svelte-check`, `test`, `formatting`, and `docker-build`.

The Gantt feature remains in #10992 and builds on this PR.


packages/ui/src/components/SearchInputAdvanced.encoder.ts
```diff
@@ -0,0 +1,378 @@
...
+const PREFIX_ALIAS: Record<string, string> = {
+  title: 'searchTitle',
+  id: 'identifier',
+  comments: 'comments.message'
+}
+
+/**
+ * Allowed ES-native fields users may type directly.
+ *
+ * KEEP IN SYNC with elastic/src/adapter.ts KNOWN_FIELD_RE (and vice versa).
```
ArtyomSavchenko: What do you think about refactoring it to ensure both sides use synced fields?
For example, add a small reusable package with common configs for server side and UI.

models/tracker/src/viewlets.ts
```diff
@@ -23,6 +23,37 @@ import tags from '@hcengineering/tags'
 import { type ViewOptionModel, type BuildModelKey, type ViewOptionsModel } from '@hcengineering/view'
 import tracker from './plugin'
 
+// Shared Customize-View knobs reused by `issuesOptions()` (List + Kanban).
+// Kept as a single named constant so the search-scope / quick-filter /
+// highlight toggles stay consistent across every viewlet that opts in.
+const SEARCH_VIEW_OPTIONS: ViewOptionModel[] = [
+  {
+    key: 'showQuickModeSelector',
```
ArtyomSavchenko, line 31: Do I understand correctly that this isn't used in this PR, but will be used in Gantt charts?

packages/ui/src/__test__/SearchInputAdvanced.test.ts
```diff
@@ -0,0 +1,251 @@
+import { encodeSearch } from './SearchInputAdvanced.encoder'
+import { propSyncValue } from './SearchInputAdvanced.sync'
+
+describe('encodeSearch', () => {
```
ArtyomSavchenko, line 4: I think it would be better to place tests in a separate `__tests__` folder

plugins/tracker-assets/lang/zh.json
```diff
@@ -293,7 +293,18 @@
...
+    "ShowQuickModeSelector": "Quick filter: All / Active / Backlog",
+    "SearchScopeLabel": "Search in…",
+    "SearchScopeTitle": "Title only",
+    "SearchScopeTitleDescription": "Title + Description",
+    "SearchScopeAll": "Title + Description + Comments",
+    "SearchHighlight": "Highlight search matches",
+    "SearchEmptyTitle": "No issues found for \"{query}\"",
+    "SearchEmptyActiveFilters": "Active filters: {filters}",
+    "SearchEmptyClearFilters": "Search without filters",
+    "SearchEmptyAllProjects": "Search in all projects"
```
ArtyomSavchenko: I suggest to fix translations for this and other locales

packages/ui/src/components/SearchInputAdvanced.encoder.ts
```diff
@@ -0,0 +1,378 @@
...
+      // see the constant's JSDoc for why `-`, `*`, `?` are excluded).
+      //
+      // Special case: leading '-' would be parsed as Lucene NOT-operator
+      // even though mid-token '-' is tolerant. Escape the leading minus
+      // explicitly so the orphan token stays a literal term.
+      if (tok.raw.startsWith('-')) {
+        return '\\-' + tok.raw.slice(1).replace(/[+!(){}[\]^"~\\/:&|<>=]/g, '\\$&')
+      }
+      if (!PREFIX_VALUE_RESERVED_RE.test(tok.raw)) return tok.raw
+      return tok.raw.replace(/[+!(){}[\]^"~\\/:&|<>=]/g, '\\$&')
```
ArtyomSavchenko: [minor] I suggest to move this regex to const

plugins/tracker-resources/src/components/issues/KanbanView.svelte
```diff
@@ -143,6 +144,17 @@
...
 
+  // Feed the shared result-count store so IssuesView can show
+  // its SearchEmptyState card when the user's search yields zero hits.
+  // The count is written from the same fast-query callback that sets the
+  // data source (see docsQuery below), not reactively off `tasks` — the old
+  // coupling read `tasks` (fast + lagging slow query) while gating on a flag
+  // set by the fast query alone, so a stale slow-query result could skew the
+  // count right after a search changed.
+  onDestroy(() => {
+    resultIssueCountStore.set(-1)
```
ArtyomSavchenko: `resultIssueCountStore` is a single global mutated by multiple viewlets.
List and Kanban both write it; IssuesView resets it to -1 on query/filter change. The -1 sentinel + onDestroy reset largely covers teardown races, but during a viewlet swap (List→Kanban) both can be briefly mounted and write competing counts.

packages/ui/src/components/SearchInputAdvanced.encoder.ts
MichaelUray: Done — exported `fullTextSearchFields` from `@hcengineering/core`. Both `@hcengineering/ui` and `@hcengineering/elastic` already depend on core and it already hosts fulltext concepts, so no new package or dependency changes were needed. The encoder and the elastic adapter now derive their field lists from that single constant; per-field boosts stay in the adapter. A shared `escapeRegExp` (also in core) keeps the regex processing identical on both sides.

models/tracker/src/viewlets.ts
MichaelUray: It is already used in this PR: `SEARCH_VIEW_OPTIONS` is included by `issuesOptions()`, which is consumed by both the List and Kanban viewlets. #10992 reuses it for Gantt as a third consumer.

packages/ui/src/__test__/SearchInputAdvanced.test.ts
MichaelUray: Moved all five new co-located test files into the existing package folders: `packages/ui/src/__test__` and `plugins/view-resources/src/__tests__`.

plugins/tracker-assets/lang/zh.json
MichaelUray: Translated the new keys in all supported locales, reusing each locale's existing terminology. The placeholder-parity test verifies `{query}`, `{filters}`, and `{count}` across all locale files.

packages/ui/src/components/SearchInputAdvanced.encoder.ts
MichaelUray: Extracted — the character class was duplicated three times; it now derives from a single constant, with separate test and global-replace RegExp instances to avoid the shared-`lastIndex` pitfall.

plugins/tracker-resources/src/components/issues/KanbanView.svelte
MichaelUray: Implemented an owner-token gate for the shared result count. Each primary result viewlet claims and releases ownership; writes and teardown resets from superseded owners are ignored. Reporting from the generic List component is opt-in and defaults to false; ListView opts in, while embedded lists remain outside the protocol. Unit tests cover owner supersession, stale writes, stale releases, query resets, and the embedded-list regression.

ArtyomSavchenko:
Hi @MichaelUray 
I've merged this PR, thank you for the contribution.

PR #11003, opened 2026-08-04
fix(account): allow user tokens to look up accounts by social key

IamCoder18:
## Context

Looking up an account by social key is something any logged in user should be able to do. The function that handles it was rejecting every regular user token with a Forbidden error, so lookups failed. My CLI at https://github.com/IamCoder18/huly-cli requires this functionality.

## Implementation

I removed the service-token gate. The function still throws a BadRequest when the social key is empty, looks up the person by key, and returns the account UUID when the caller asks for one and the person has an account. The reads themselves are intentional and safe to expose to any authenticated user.

## Tests

Added a small describe block covering the user-token success path, empty-input rejection, the missing-key case, and both flavors of requireAccount (account present and no account).


IamCoder18:
Hey @dnplkndll and @ArtyomSavchenko. If you have time, could you take a look at this? Thank you!

IamCoder18:
Disclaimer: the tests I added were AI generated, the **functional code was not**.
I (a human) verified all changes.

server/account/src/serviceOperations.ts
```diff
@@ -997,10 +997,6 @@ export async function findPersonBySocialKey (
     throw new PlatformError(new Status(Severity.ERROR, platform.status.BadRequest, {}))
   }
 
-  const { extra } = decodeTokenVerbose(ctx, token)
-
-  verifyAllowedServices(['tool', 'workspace', 'aibot', ...integrationServices], extra)
```
ArtyomSavchenko, line 1002: We can’t accept this change because it removes authorization from a global account lookup. Any regular user token could then probe arbitrary social keys and learn whether an email or external identity exists, as well as obtain the associated account/person UUID. This creates a cross-tenant user-enumeration and privacy risk.

ArtyomSavchenko (changes requested):

server/account/src/serviceOperations.ts
IamCoder18: You're right. I've modified it only allow lookups if both accounts share a workspace.

PR #11005, opened 2026-08-12
feat(drive): download folder as archive

AitiX:
## Description

Adds a **Download as archive** action for folders in the Drive module. Previously only
individual files could be downloaded; folders with many files (e.g. build artifacts)
had to be fetched one file at a time.

The action walks the folder tree through the platform client, fetches every file blob
and packs everything into a single `<folder name>.zip` on the client, preserving the
relative folder structure inside the archive.

### Implementation notes

- **New action** `drive.action.DownloadFolder` / `drive.actionImpl.DownloadFolder`,
  registered for `drive.class.Folder` next to the existing file `Download` action
  (same category, `tools` group, context + browser modes).
- **Traversal** (`plugins/drive-resources/src/archive.ts`): recursive
  `client.findAll` over `drive.class.File` (with `$lookup` of the current
  `FileVersion`) and `drive.class.Folder`, sorted by title for deterministic
  archives. A visited-set guards against cycles, and duplicate titles within one
  folder get a numeric suffix.
- **Blob access** reuses `getFileUrl` from `@hcengineering/presentation` (the same
  URL the file `Download` action uses) and plain `fetch`, matching how
  `getJsonOrEmpty` fetches blobs elsewhere in the platform.
- **Zip library**: [`fflate`](https://www.npmjs.com/package/fflate) (`^0.8.2`), added
  to `@hcengineering/drive-resources`. No workspace package currently depends on a
  zip library directly (`jszip` appears in the lockfile only as a transitive
  dependency of docx tooling), so the smallest streaming option was chosen.
  Entries are added via `ZipPassThrough` (store, no re-compression): Drive blobs are
  typically already compressed or incompressible (archives, media, build artifacts),
  so this keeps CPU usage minimal while still producing a standard zip.
- Files are fetched sequentially; the archive is assembled in memory and saved via a
  temporary object URL. Empty folders produce no download. Errors are reported through
  `setPlatformStatus(unknownError(...))`, consistent with other drive-resources utils.
- **Localization**: new `DownloadAsArchive` string (en: "Download as archive",
  ru: "Скачать архивом"). Other locales fall back to English, matching existing
  partially-translated keys in `drive-assets`.

### Changed files

- `models/drive/src/index.ts` — register the folder action
- `models/drive/src/plugin.ts` — `action.DownloadFolder`, `actionImpl.DownloadFolder` ids
- `plugins/drive-resources/src/archive.ts` — new: traversal + zip + save
- `plugins/drive-resources/src/index.ts` — `DownloadFolder` action implementation
- `plugins/drive-resources/src/plugin.ts` — `DownloadAsArchive` string id
- `plugins/drive-resources/package.json` — add `fflate`
- `plugins/drive-assets/lang/en.json`, `plugins/drive-assets/lang/ru.json` — new string
- `common/config/rush/pnpm-lock.yaml` — lockfile update for `fflate`

## How to test

1. Open a workspace with the Drive application enabled.
2. Create a drive, add a folder with a few files and at least one nested subfolder
   with files (mixed types; ideally also two files with the same name in one folder).
3. Right-click the folder (or use the `...` menu) → **Download as archive**.
4. A `<folder name>.zip` download should start; after extraction the content must
   mirror the Drive folder structure, including nested subfolders.
5. Verify an empty folder triggers no download, and a folder whose name contains
   `/` or `\` still produces a valid archive name.
6. Regression: file **Download**, folder **Rename**/**Delete**/**Move** actions
   still work.

## Checklist

- [x] Affected packages compile (`rush build --to @hcengineering/model-drive --to @hcengineering/drive-resources`)
- [x] Localized strings added (en, ru)
- [ ] Screenshots / screencast (to attach when opening the PR)

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -20361,6 +20361,9 @@ importers:
       fast-equals:
         specifier: ^5.2.2
         version: 5.3.2
+      fflate:
+        specifier: ^0.8.2
+        version: 0.8.3
       svelte:
         specifier: ^4.2.20
         version: 4.2.20
@@ -47614,6 +47617,9 @@ packages:
   fecha@4.2.3:
     resolution: {integrity: sha512-OP2IUU6HeYKJi3i0z4A19kHMQoLVs4Hc+DPqqxI2h/DPZHTm/vjsfC6P0b4jCMy14XizLBqvndQ+UilD7707Jw==}
 
+  fflate@0.8.3:
+    resolution: {integrity: sha512-tbZNuJrLwGUp3zshBtdy4W+ORxZuIh8a5ilyIEQDC5rY1f3U20JMry0Ll3WBzU58EZKsEuJFXhb5gwv8CsPvgA==}
+
   file-api@0.10.4:
     resolution: {integrity: sha512-RVBXJGmsnQxokdpy264pmsdBjbUuxE6QT2xxhOrO2pzwTetbTNoWVFgkONFWmopm5mellsXrQIQhMY9fjufi9g==}
 
@@ -59263,6 +59269,8 @@ snapshots:
       pend: 1.2.0
 
   fecha@4.2.3: {}
+
+  fflate@0.8.3: {}
 
   file-api@0.10.4:
     dependencies:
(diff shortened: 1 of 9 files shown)


AitiX:
Closing: moving this work to a private repository. Happy to reopen from a fresh fork if there is interest in the feature.

PR #11013, opened 2026-08-13
test(sanity): make ChannelPage.sendMention's retry loop reachable

clayrisser:
## Problem

`ChannelPage.sendMention` wraps its work in a three-attempt retry loop (`tests/sanity/tests/model/channel-page.ts:108-120`):

```ts
for (let i = 0; i < 3; i++) {
  try {
    await this.inputMessage().fill(`@${message}`)
    await this.selectMention(message, categoryName)
    break
  } catch (error: any) {
    if (i === 2) {
      throw error
    }
```

**Attempts 2 and 3 are unreachable.** `selectMention` (`common-page.ts:206-208`) is:

```ts
async selectMention (mentionName: string, categoryName?: string): Promise<void> {
  await this.mentionPopupListItem(mentionName, categoryName).first().click()
}
```

— a bare `click()` with no timeout. The sanity project sets `timeout: 60000` and `expect.timeout: 15000` but **no `actionTimeout`** (`tests/sanity/tests/playwright.config.ts:10-43`), so Playwright's default applies to that click and it waits indefinitely. When the mention popup has no matching entry, the click blocks until the 60-second *test* timeout fires — and a test timeout aborts the test rather than raising into the `catch`.

So the retry loop never runs a second attempt. It reads like resilience and provides none.

## Fix

Click with an explicit timeout, so a missing popup entry surfaces as a catchable error well inside the test budget:

```diff
       try {
         await this.inputMessage().fill(`@${message}`)
-        await this.selectMention(message, categoryName)
+        // An explicit timeout is required: the suite sets no `actionTimeout`, so the
+        // default click would wait out the whole test timeout and the retries below
+        // would never run.
+        await this.mentionPopupListItem(message, categoryName).first().click({ timeout: 5000 })
         break
```

`mentionPopupListItem` is the same locator `selectMention` uses, inherited from `CommonPage` (`common-page.ts:44`), so this calls the same thing with a bound. `selectMention` itself is left alone — its other callers are not in a retry loop and are not affected by this reasoning.

**The retry needs no other change.** `fill()` on the contenteditable selects the existing content and replaces it, so the second attempt's `fill` re-triggers the suggestion plugin (exit, then start with a fresh query) even though the editor already holds `@name`. I checked whether an explicit "clear the input between attempts" step was needed and it is not — the traces with and without it are identical, so it would have been dead code.

## Scope and residual risk

- Test-only. No product code is touched.
- 5 s is comfortably longer than a healthy popup takes to appear and comfortably shorter than the 60 s test budget, leaving room for all three attempts plus the rest of the test.
- If the popup genuinely never appears, the test still fails — it just fails after three real attempts and with a clear `TimeoutError` on the click, rather than as an opaque whole-test timeout.

## Verification

Verified against `develop` @ `1be6047c8`: `playwright.config.ts` still sets `timeout: 60000` and `expect: { timeout: 15000 }` with no `actionTimeout` in `use`, `selectMention` is still an unbounded `click()`, and `mentionPopupListItem` is still on `CommonPage`. `git apply` is clean.

I have not run the sanity suite against a live stack as part of preparing this PR — the argument above is from the config and the two call sites. Your CI running the suite is the check that matters here.

## Related

This is one half of an intermittent chat-backlinks failure. The other half is the mention popup itself never retrying its fulltext query, which I have opened separately as a product-code fix. Together they explain the flake: indexing lags, the popup stays empty, and the retry that should have absorbed it cannot run. The two are independent and can land in either order — this one is worth having regardless, because an unreachable retry loop is misleading on its own.

### DIFF
--- a/tests/sanity/tests/model/channel-page.ts
+++ b/tests/sanity/tests/model/channel-page.ts
@@ -109,7 +109,10 @@ export class ChannelPage extends CommonPage {
     for (let i = 0; i < 3; i++) {
       try {
         await this.inputMessage().fill(`@${message}`)
-        await this.selectMention(message, categoryName)
+        // An explicit timeout is required: the suite sets no `actionTimeout`, so the
+        // default click would wait out the whole test timeout and the retries below
+        // would never run.
+        await this.mentionPopupListItem(message, categoryName).first().click({ timeout: 5000 })
         break
       } catch (error: any) {
         if (i === 2) {


clayrisser:

PR #11030, opened 2026-09-02
fix(love): clear stale ParticipantInfo on LiveKit participant_left / room_finished

nigad:
Companion to #11029 (client side). This one fixes the state for everyone else, server side.

### Description of the issue

`ParticipantInfo` (who is in which Love room) is written only by the person's own client, on join and on leave, and dropped by the server only when the person goes fully offline (`OnUserStatus`). When the client goes away mid-call — page reload, crash, laptop lid, network loss — LiveKit drops the participant, but the record keeps pointing at the room. The person stays listed as a member of the room for everyone, indefinitely, and nobody can remove them: the client offers **Kick** only to an office owner inside their own office.

The Love service already receives LiveKit webhooks (`/webhook`, used for `egress_ended`) and logs `room_started` / `room_finished` without acting on them.

### Steps to reproduce

1. Two users join a video room.
2. One of them reloads the page (or loses network) mid-call.
3. Observe the room from the other user's office view.

### Expected behaviour

Once LiveKit has dropped the participant, the room no longer lists them.

### Actual behaviour

They remain listed in the room until they go fully offline. On a self-hosted instance we saw a user pinned to a room this way for the rest of the day; LiveKit had logged `participant closing … CLIENT_REQUEST_LEAVE` at the moment of the reload.

### Fix

- `services/love/src/rooms.ts`: `parseRoomName` (workspace = first `_` segment, room id = last; room names may contain underscores) and `resetRoomParticipants(client, roomId, person?)`, which moves each matching `ParticipantInfo` to the person's office or to `love.ids.Reception` — the same update the client's `kick()` performs, so the existing `OnParticipantInfo` triggers keep `RoomInfo` / meeting minutes consistent.
- `WorkspaceClient`: `leaveRoom(person, roomId)` and `clearRoom(roomId)` over the existing system-token `TxOperations`.
- `/webhook`: `participant_left` → `leaveRoom(identity, roomId)` (the LiveKit identity is the Huly `Person._id` the client posts to `/getToken`); `room_finished` → `clearRoom(roomId)`. Filtering by room makes a late `participant_left` harmless for someone who already moved to another room. Errors are logged and the webhook still returns 200 so LiveKit does not retry forever.

LiveKit only emits `participant_left` after the participant is fully gone (past its reconnect window), so no extra grace period is needed. Agent / bot identities have no `ParticipantInfo` and are a no-op. No new configuration: deployments that already receive egress webhooks get this for free.

### Testing

Unit tests for `parseRoomName` and `resetRoomParticipants` (`services/love/src/__tests__/rooms.test.ts`): office target, reception fallback, person already elsewhere, whole-room reset. The reset write itself was verified on a self-hosted v0.7.432 instance through the transactor API (it is what freed the stuck user).

Developed with AI assistance; reviewed and tested by the author.

### DIFF
--- a/services/love/src/__tests__/rooms.test.ts
+++ b/services/love/src/__tests__/rooms.test.ts
@@ -1,0 +1,93 @@
+//
+// Copyright © 2026 Hardcore Engineering Inc.
+//
+// Licensed under the Eclipse Public License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License. You may
+// obtain a copy of the License at https://www.eclipse.org/legal/epl-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import { type TxOperations } from '@hcengineering/core'
+import love from '@hcengineering/love'
+import { parseRoomName, resetRoomParticipants } from '../rooms'
+
+describe('parseRoomName', () => {
+  it('takes the workspace from the first segment and the room id from the last', () => {
+    expect(parseRoomName('ws-1_All hands_room-1')).toEqual({ workspace: 'ws-1', roomId: 'room-1' })
+  })
+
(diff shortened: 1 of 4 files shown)


nigad:
Closing; will be resubmitted from the correct account.

